PR #28, opened 2020-04-29
reduced readme

chrisdane:
Hi

As a user, I find it very confusing to have 2 different places of documentation, the `README.rst` and `readthedocs.io`. My suggested readme would make it easier for me, and hopefully for many future users as well.

Kind regards,
Chris


pgierz:
While reducing the readme to a minimum would be nice, currently, it's displayed as one section of the readthedocs page. We should think about what would be actually required for the landing page on GitHub, but I think just linking to readthedocs is insufficient. Users might want to at least have some info to go over directly from GitHub. I guess writing a correct documentation and splitting it up is an art for itself....

dbarbi:
I agree with Paul, having information at the exact spot where someone might look is difficult, that's why I am strictly against a minimal README. I included a line that points toward the readthedocs-page however.

PR #46, opened 2020-06-11
Bugfixes foci for paul prep minor

seb-wahl:
compiles and runs FOCI if, and only if the branch 'fix-datemath-for-other-than-seconds' from esm_parser (see https://github.com/esm-tools/esm_parser/pull/4) and the branch 'fix-globbing-and-need_year_after' branch of esm_runscripts (see https://github.com/esm-tools/esm_runscripts/pull/4) are used.
FOCI-OIFS will be tested (very confident that it will work without further fixes) once I have access to the OIFS src code that sits in the more restricted ec-earth tree on gitlab.dkrz.de.


pgierz:
Thanks Sebastian! I'm still fixing up some PISM stuff, but tomorrow I'll hopefully get to merging everything in esm-runscripts into one common branch. Once that's done, I might ask you to do some tests again with new branches of parser and runscripts.

I've also started on Continuous Integration testing, which should (hopefully) make the merge parties more "party" and less "work" in the future :-)

PR #54, opened 2020-06-18
fixed FOCI compile after setup2models splitup

seb-wahl:
as the splitup was based on an old setup2models.yaml (at least w.r.t
FOCI settings)


pgierz:
Note: Failed checks relate to awicm-3.0 and awicm-3.1, which are both triggered since oifs.yaml changed. I don't have access to that repository, so the check fails during the git-clone step since it tries to use my user account. This should be re-thought if we fully adopt the CI (I think we should adopt it). Merging to release and develop as a hot fix.

PR #79, opened 2020-07-02
hotfix for moving the echam compilation standalone

mandresm:


pgierz:
Hey Miguel, you can go ahead and merge, the failing checks are from:

* awiesm with PISM (which works in a branch, but we will need to finish modifying esm-master, which Dirk will do)
* awicm-recom (I don't have access to one of the required repos, and the CI might be configured incorrectly)
* awicm with esm-interfaces (Nadine doesn't want this automatically tested, I still need to remove the CI jobs)

mandresm:
Closed because of a more generalized solution was found:  esm-tools/esm_master#9

PR #80, opened 2020-07-02
Hotfix/echam comp

mandresm:


mandresm:
Closed because of a more generalized solution was found:  esm-tools/esm_master#9

PR #84, opened 2020-07-08
Packaged Configs and Namelists

pgierz:
This feature allows for namelists and configuration files to be loaded directly from the pip-installed esm_tools package. Effectively, this removes the dependence of `FUNCTION_PATH` and `NAMELIST_PATH` in the esm tools rc file.

Before, you had to access a specific configuration like this:

```python
config = yaml_file_to_dict(FUNCTION_PATH+"/echam/echam.yaml")
... further steps that are then done with the parser ...
```

Instead, you can now do this:
```python
config = esm_tools.read_config_file("echam/echam.yaml")
... further steps that are done with the parser ...
```

This has one big advantage: You could now, in principle, keep multiple environment installs of esm_tools on a machine, and have each one work independently of the other.

The same works for namelists as well (use `esm_tools.read_namelist_file`), and it works both in editable mode (`pip install -e`) as well as regular install (`pip install`)

It also is an important step that would allow us to publish the esm_tools on pypi, so in future users could just do:

```
$ pip install esm_tools
```
without all the git stuff.

Since the config files are meant to provide sane defaults for everything, the users should not really want or need to modify them, since this can all be done over the runscript.


pgierz:
One thing I am not sure about is how this will work together with the `-U` option... Since the configs and namelists are always read from the package, updating them doesn't work anymore. Maybe something we need to discuss further...

seb-wahl:
In principle a good idea, but having **sane** defaults for every model component and all it's use cases (think of ECHAM6 being used in uncoupled and coupled setups (MPIOM, NEMO, FESOM) and different versions) is a though task I think

pgierz:
@seb-wahl you can still modify the config files if you do an editable install, so they will still be changed just like before.

pgierz:
Closing, manually merged into prep release

PR #111, opened 2020-08-31
OIFS updates

seb-wahl:
Merge should be smooth for other models because modifications only affect OIFS, NEMO and XIOS, except for comments (`esm_plugins.yaml`) or three new variables (`oasis3mct.yaml`) that are only used if oasis3mct version 4.0 is used.

### DIFF
--- a/configs/components/oasis3mct/oasis3mct.yaml
+++ b/configs/components/oasis3mct/oasis3mct.yaml
@@ -2,6 +2,9 @@ model: oasis3mct
 #model_dir: ${esm_master_dir}/oasis
 cf_name_dir: ${model_dir}
 time_step: 900 # changed by coupled setup anyway
+norestart: F
+use_lucia: False
+mct_version: 2.8 # TODO: set mct_version = version (can't do at the moment as version is a string)
 
 config_files:
         cf: cf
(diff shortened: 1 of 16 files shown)


configs/components/oasis3mct/oasis3mct.yaml
```diff
@@ -2,6 +2,9 @@ model: oasis3mct
 #model_dir: ${esm_master_dir}/oasis
 cf_name_dir: ${model_dir}
 time_step: 900 # changed by coupled setup anyway
+norestart: F
```
pgierz, line 5: Should this be a string? I'm not sure if F just becomes False with the parser

configs/esm_software/esm_runscripts/esm_plugins.yaml
```diff
@@ -26,7 +26,15 @@ core:
 
 # To add your custom plugin, see oifs.yaml, section compute_recipe, and compare to the list above.
 # the custom plugin 'preprocess' listed in the compute_recipe in oifs.yaml needs to be installed beforehand with
+#
 # git clone https://github.com/esm-tools-plugins/preprocess
 # cd preprocess
-# if required activate python3 e.g. module load anaconda3/bleeding_edge on mistral
+# if required activate python3 e.g. 
+# module load anaconda3/bleeding_edge on
 # pip install --user
```
pgierz, line 34: You could add that to the list of commands oifs executes when it is installed.

pgierz (commented):
I can't comment on all the yaml files for correctness, but it looks good. If it works we can put it in.

configs/esm_software/esm_runscripts/esm_plugins.yaml
seb-wahl: Good point, but I'll do that on a later stage given the fact that we are very close to the release

pgierz: OK, I'll make an issue so we don't forget later on

PR #128, opened 2020-09-09
Feature/compile model at start

### LINKED ISSUE #99: Feature request: Model compilation and source code copying becomes part of job submission
To make sure that we can always reproduce the source code from which a specific simulation stems, it would be perfect if the esm-tools, or runscripts, would perform the following two additional tasks _before the first model year only_:

1. before copying the model executables to the simulation base dir, trigger a compilation - and quit / do not submit compute job if compilations fails
2. also copy the whole source code branch (less than 500 MB) to the simulation folder (copying is already done for the functions, so general infrastructure that can be employed for copying the model source code should already exist in the esm tools/functions)

pgierz:
Hi @christian-stepanek, this does what we had talked about for compiling the model before the first run. See if you understand the code, and feel free to comment on any documentation that is missing. See also: https://github.com/esm-tools/esm_runscripts/pull/35

### DIFF
--- a/configs/esm_software/esm_runscripts/esm_plugins.yaml
+++ b/configs/esm_software/esm_runscripts/esm_plugins.yaml
@@ -1,6 +1,7 @@
 core:
         esm_runscripts:
                 compute:
+                        - "compile_model"
                         - "_create_setup_folders"
                         - "_create_component_folders"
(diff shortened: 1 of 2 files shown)


dbarbi (changes requested):
That should not be the default

PR #132, opened 2020-09-14
LC_ALL variable defined in all machines to avoid locale issues

mandresm:
Users are confronting the problem where the locale is missing (they have not have export `LC_ALL=en_US.utf-8`) and this leads to models not compiling.

In this bugfix I added the the export of `LC_ALL=en_US.utf-8` in every machine file, to avoid this problem so that users don't have to worry about the locale.


pgierz (approved):
I find it odd that compilation requires the local to be set to US English...but since it's only happening in a sub shell, it shouldn't mess up the user's other choices in their main environment I guess.

mandresm:
The error happened in blogin where `perl` was complaining about the locale and after that, cmake will crash returning the following:

```
Module for Intel-MPI (version 2019.5) loaded.
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LC_CTYPE = "UTF-8",
        LANG = "en_US.utf-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
-- setting default build type: Release
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
-- Check for working C compiler: /sw/comm/impi/compilers_and_libraries_2019.5.281/linux/mpi/intel64
/bin/mpiicc
-- Check for working C compiler: /sw/comm/impi/compilers_and_libraries_2019.5.281/linux/mpi/intel64
/bin/mpiicc -- broken
CMake Error at /sw/tools/cmake/3.16.2/skl/share/cmake-3.16/Modules/CMakeTestCCompiler.cmake:60 (mes
sage):
  The C compiler

    "/sw/comm/impi/compilers_and_libraries_2019.5.281/linux/mpi/intel64/bin/mpiicc"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /home/hbkmandr/model_codes/fesom-recom-1.4/fesom-1.4/build/CMakeFiles/CMakeTmp
[ ... ]
```



> it shouldn't mess up the user's other choices in their main environment I guess.

I checked that, and infact is not overriding the user's choices.

Issue #130, opened 2020-09-09
ECHAM restart problems


ghost:
Hi, 
I started a 3 year ECHAM standalone time slice simulation on mistral that is located at

`/work/bb1152/Module_A/A3_ArcClimEx/a271093_riebold/model_codes/Experiments/test_run_LAMO_HICE_sara_3years/`

with the corresponding runscript at

`/work/bb1152/Module_A/A3_ArcClimEx/a271093_riebold/model_codes/Experiments/test_run_LAMO_HICE_sara_3years/scripts/echam_johannes.yaml`

Currently, I am submitting a new job for each year.
As intended restarts files are written at the end of each year, but it's currently not using them when starting the subsequent year. Therefore, the problem apparrently is that for each year the same initial run is done as for the first year.

I'm not completely sure, but I think a potential source of this problem might be that in the echam namelist of the runs for the 2nd and 3rd year 
e.g. 
 `/mnt/lustre02/work/bb1152/Module_A/A3_ArcClimEx/a271093_riebold/model_codes/Experiments/test_run_LAMO_HICE_sara_3years/run_18510101-18511231/work/echam.namelist`

 the `dt_start` parameter is defined and increased accordingly. However, as explained in some ECHAM docs

> Make sure that dt_start is either not present in the RUNCTL namelist or has still the (old)start date of the beginning of the experiment

this parameter shouldn't be defined when starting from restart. `lresume` is set to True as expected for the 2nd and 3rd year.
Any ideas how to fix this problem?

Cheers Johannes

mandresm:
Could you please copy here the output of `esm_versions check`?

pgierz:
...can you also paste a link to the echam docs? That seems like something we definitely should be taking care on.

ghost:
> 
> 
> Could you please copy here the output of `esm_versions check`?


output of `esm_versions check`:

> You are using the following esm_tools versions:
> -----------------------------------------------
> esm_archiving : unknown version!
> esm_autotests : unknown version!
> esm_calendar: 4.0.1
> esm_database: 4.0.0
> esm_environment: 4.2.0
> esm_master: 4.2.0
> esm_parser: 4.2.1
> esm_profile: 4.0.0
> esm_rcfile: 4.0.0
> esm_runscripts: 4.2.1
> esm_tools: 4.2.0 (development install, on branch: Johannes, describe=v4.2.0-2-gc9f93fe-dirty)
> esm_plugin_manager: 4.2.0
[...]

PR #133, opened 2020-09-14
fixes the namelist_changes inside echam.yaml

mandresm:
This is a fix for issue #130.

Fixes the `namelist_changes` (renamed into `add_namelist_changes`) inside… the choose_<option> that were not working in echam.yaml.

I'll prepare another hotfix for the `esm_parser` where a file with two `namelist_changes` returns and error or at least asks if one should overwrite the another.

### DIFF
--- a/configs/components/echam/echam.yaml
+++ b/configs/components/echam/echam.yaml
@@ -356,7 +356,7 @@ choose_lresume:
                                   # - ndate "<--set_dim--" ${steps_in_year_before}
         True:
                 # pseudo_start_date: $(( ${start_date} - ${time_step} ))
-                namelist_changes:
+                add_namelist_changes:
                         namelist.echam:
                                 runctl:
                                         dt_start: "remove_from_namelist"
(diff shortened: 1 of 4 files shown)


configs/components/echam/echam.yaml
```diff
@@ -356,7 +356,7 @@ choose_lresume:
                                   # - ndate "<--set_dim--" ${steps_in_year_before}
         True:
                 # pseudo_start_date: $(( ${start_date} - ${time_step} ))
-                namelist_changes:
+                add_namelist_changes:
                         namelist.echam:
                                 runctl:
                                         dt_start: "remove_from_namelist"
```
pgierz, line 362: I'd still like to actually understand what the correct value here should be...

pgierz (approved):
Looks good. I still would like to understand what in Echam the required value for dt_start really should be, and having the parser complain when it sees multiple namelist changes would be a good idea.

PR #140, opened 2020-09-15
sample component added

mandresm:
The sample component is a simple C script that generates files with different names inside the same folder the binary is run. It can be installed using the standard `esm_master install-sample-1.0` and run with `esm_runscripts <runscript_name> -e <experiment_id>`. Its main purpose is to debug file dictionaries (testing of esm-tools/esm_runscripts#37) with minimal resources, but could be also used in the future for workshops.


pgierz (approved):
Looks good. I think we can use this as a template during workshops, and maybe include it directly in the handbook. It might be that we add more detailed comments, but that will become clear when we use it for demonstration the first time.

PR #149, opened 2020-10-05
specific environment for coupled setup fesom-recom fixed inside fesom-recom.yaml

mandresm:


pgierz (commented):
Is the double add correct?

mandresm:
funny enough yes, without the double add it destroys the `AWI_FESOM_YAML` in the previous `add_export_vars`.

pgierz (approved):

PR #155, opened 2020-10-09
Adds ability for DDT debugger on Slurm systems

pgierz:
Mistral has the ability to run interactive debugging on large scale Fortran or C programs. See more info here:
https://www.dkrz.de/up/services/code-tuning/debugging

By default, this is turned off. However, if you need it, you can add:

```yaml
general:
    debugger: ddt
```

This results in a slightly modified execute command:
```
time ddt --connect srun -l --kill-on-bad-exit=1 --cpu_bind=cores --multi-prog hostfile_srun &
```

### DIFF
--- a/configs/other_software/batch_system/slurm.yaml
+++ b/configs/other_software/batch_system/slurm.yaml
@@ -24,10 +24,10 @@ choose_accounting:
 
 choose_mail_type:
         "*":
-                mail1:  "--mail_type=${mail_type}"
+                mail1:  "--mail-type=${mail_type}"
 choose_mail_user:
         "*":
-                mail2:  "--mail_user=${mail_user}"
+                mail2:  "--mail-user=${mail_user}"
 
 notification_flag: "${mail1} ${mail2}"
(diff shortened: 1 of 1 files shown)


pgierz:
There is currently no way to change the module commands from the batch system yaml file. I tried:

```diff
choose_debugger:
    None:
          debugger_flags_prelauncher: ""
                     ddt:
                           debugger_flags_prelauncher: "ddt --connect"
+                          add_computer.module_actions:
+                                - "load ddt"
```

but that did not work. @dbarbi, any ideas?

nwieters:
The module on Mistral is `arm-forge`:
`module load arm-forge`

I ran `ddt` on Mistral several months ago, and I found in my notes that you need to compile with `-g -O0 ` and you need X-forwarding since `ddt` will open a GUI and you have to 'accept the incoming reverse connection request'. And I am not sure if you need to have the `ddt` GUI already running before submitting the job script.

mandresm (approved):

PR #166, opened 2020-10-20
Hotfix for Missing CDO in MISTRAL.yaml

pgierz:
Why is there no CDO in the mistral.yaml file?

It is needed to generate the o2a_flux and a2o_flux so oasis can restart.


pgierz:
Closes #168

PR #172, opened 2020-10-21
Feature/PISM Support

### LINKED ISSUE #8: Missing Models
So it doesn't get forgotten...

we still need PISM and VILMA support so I can keep working on the coupler.

pgierz:
PISM Support in the ESM Tools. Still needs a plugin, see: https://github.com/esm-tools-plugins/esm_pism

Documentation to follow.

Closes #8


mandresm (approved):
Hi Paul! It looks great. I have only one question, is the ksh file `pism.function` needed?

pgierz:
Ah, no...obviously not.

pgierz:
Here also is the documentation: https://esm-pism.readthedocs.io/en/latest/index.html
@mandresm: Should I add a link to that to the PISM section of the handbook?

mandresm:
> Should I add a link to that to the PISM section of the handbook?

Sure!

pgierz:
but, where? extra column in the supported models table? 

Maybe something we can discuss in the next meeting; the documentation we have so far is really nice, but we can think about a better structure I think. 

If there are no protests, I'll will merge this to develop on wednesday; then it will end up in the next release sometime before Christmas :-)

mandresm:
I agree we should discuss this on Monday, ECHAM users asked for a section in the documentation where we talk specifically about the ECHAM defaults, so maybe we are going to end up having a chapter for model specific stuff and then it could go in this chapter. The link could also be included in the model table, but I think there, it has very little visibility.

Anyway, agreed that we should discuss about the documentation structure. I have ended up putting most of the things I worked on inside of the YAML syntax chapter, and is sort of a mess now, specially with the amount of info this chapter has ended up with.

Issue #181, opened 2020-10-23
Cookbook chapter


mandresm:
**Request/contribution description**
Create the Cookbook chapter

**Suggested chapter/section in the documents**
Cookbook

**Is this specific to any model/couple setup or to any ESM-Tools functionality?**
General

PR #184, opened 2020-10-23
Docs/cookbook

mandresm:
Addresses issue #181

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -90,7 +90,10 @@ esm_tools_modules = [
  "esm_version_checker",
 ]
 esm_tools_modules.remove("esm_tools")
-# Uncomment the following if you don't want to compile the docstrings
+# Creating a docs/.docstrings.yml file allows you to control the compilation of
+# the docstrings. Fill this file with ``docstrings: 0`` if you want to avoid
+# docstring's compilation, or ``docstrings: 1`` if you want to compile the
+# docstrings. If the file does not exists it always compiles the docstrings.
 if os.path.isfile(".docstrings.yml"):
     with open(".docstrings.yml") as docstrings_yaml:
          docstrings_dict = yaml.load(docstrings_yaml, Loader=yaml.FullLoader)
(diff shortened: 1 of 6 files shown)


docs/conf.py
```diff
@@ -90,6 +90,12 @@
  "esm_version_checker",
 ]
 esm_tools_modules.remove("esm_tools")
+# Uncomment the following if you don't want to compile the docstrings
```
nwieters: Should these lines are commented out by default?

mandresm: This is a deprecated comment. Now it is easier to avoid compiling the docstrings. You just need to create a `.docstrings.yml` with the content `docstrings: 0`. I'll change this comment.

nwieters: Now I got it. Thanks.

nwieters (approved):

docs/recipes/template.rst
```diff
@@ -0,0 +1,36 @@
+Recipe Title
```
pgierz, line 1: We should make sure that this doesn't end up in the actual website. We can still include it in the repository for people to use as a template, but I think the cookbook would be cleaner if we just have user-friendly things there.

docs/recipes/template.rst
```diff
@@ -0,0 +1,36 @@
...
+============
+
+.. use = for sections, ~ for subsections and - for subsubsections
+
+**Feature available since version:** <version_num>
+
+General description here.
+
+1. Step 1
+   make sure you indent correctly inside steps
```
pgierz, line 11: Sometimes, I hate RST... I have so often had to recompile docs because I am missing on single space. :-(

pgierz (approved):

docs/recipes/template.rst
mandresm: Yes, sorry, I assumed that the first person adding a recipe will substitute the template call with the first recipe call, so that the template won't make it to the final documentation.

PR #206, opened 2020-11-05
fix(configs/couplings): specifies replacement for FESOM_COUPLED in se…

pgierz:
…d replacements of cmakelists

Due to how sed works, if the FESOM 2.0 CMakeLists contained both
FESOM_COUPLED and OIFS_COUPLED, both were turned on as Jan Streffing had
a comment that OIFS_COUPLED also required FESOM_COUPLED. See disucssion
here: https://github.com/FESOM/fesom2/issues/30


mandresm (approved):

pgierz:
Miguel, will you take care of merging this to both branches, or should I do it?

PR #220, opened 2020-11-24
Feature/awicm3

mandresm:


configs/components/xios/xios.yaml
```diff
@@ -55,10 +61,16 @@ destination: xios
 git-repository: https://git.geomar.de/foci/src/xios.git
 install_bins: bin/xios.x
 use_oasis: ''
-# 
+
+compiletime_environment_changes:
+  choose_computer.name:
```
seb-wahl, line 66: Why do you need `ncl` during compile time? I know it doesn't hurt but I'm surprised to see it here.

mandresm: I actually don't know but it might have to do with some of the libraries used for AWICM3, compiled by Jan. XIOS does not compile without it. @JanStreffing might be able to add more information, tbh I have not look in detail into it, but I agree, that should be removed in some point.

JanStreffing: I don't know about xios/ncl, but I can tell why it's in awicm3. I think is actually not needed at compile time. I was loading it in the ksh tools for some use during post-processing.

seb-wahl: Ok, so I will remove it with my next commit (I have some small changes to xios.yaml to be pushed anyhow)

JanStreffing (commented):
- github.com/FESOM/fesom2.git
Should be the default repository. The gitlab one is starting to fall behind and github has much better development features.

configs/components/oifs/oifs.yaml
```diff
@@ -101,10 +107,23 @@ compile_infos:
             git-repository: https://gitlab.dkrz.de/ec-earth/oifs-43r3.git
             install_bins: make/esm/oifs/bin/oifs
           43r3-awi:
-            branch: master
+            requires:
+            - oasis3mct-4.0-oifsdeck
+            branch: master # or oifs-deck???
```
JanStreffing, line 112: We should use https://gitlab.dkrz.de/ec-earth/oifs-43r3/-/tags/awicm-3-deck
Master will keep moving on.

configs/components/oifs/oifs.yaml
```diff
@@ -387,69 +498,87 @@ choose_resolution:
                 oasis_grid_name: "048"
                 res_number: 95
                 res_number_tl: "95_3"
+                truncation: "TQ"
```
JanStreffing, line 501: Is OpenIFS capable of running TQ grids? more a question for @joakimkjellsson

configs/setups/awicm3/awicm3.yaml
```diff
@@ -1,196 +1,657 @@
...
+        # 287 fulfills this requirement and 288 divides well with 36 and 48
+        # (287 = 41 * 7)
+        # so that the configuration fills up the compute nodes
+        # 575 would also be ok (575 = 23 * 25)
+        nproca: 287
+        nprocb: 1
+
+        # By default we turn on ECWAM wave model (simulation FR10) but it should be
+        # false for DRH6
+        wam: true
```
JanStreffing, line 143: Since you are running something like the DRH6 exp for now, and that one had no wam, we probably need this set for false for now. It can be changed with a small patch afterwards

configs/setups/awicm3/awicm3.yaml
```diff
@@ -1,196 +1,657 @@
...
+
+        runtime_environment_changes:
+                add_export_vars:
+                        # Turn on FESOM coupling
+                        - 'FESOM_USE_CPLNG="active"'
+                        - 'ECE_CPL_NEMO_LIM="false"'
+                        - 'ECE_CPL_FESOM_FESIM="true"'
+                        - 'ECE_AWI_CPL_FESOM="true"'
+
+        res_number_tl: ${resolution}${levels} # @JOAKIM: what is this variable for? @JAN: Is the directory correct?
```
JanStreffing, line 213: yes

configs/components/oifs/oifs.yaml
joakimkjellsson: Should work! 
They are listed at
https://confluence.ecmwf.int/display/OIFS/4.3+OpenIFS%3A+Horizontal+Resolution+and+Configurations
but I don't know anyone using them. 
Not sure why we have it listed in ESM-Tools. I suggest to remove it.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -1,196 +1,657 @@
...
+        choose_general.resolution:
+                TL159_CORE2:
+                        nproca: 279 # @JAN, are these numbers correct for this resolution?
```
JanStreffing, line 216: Yes, as long as we are not using tasksets and MPI/OMP. Once we use MPI/OMP, we need full nodes for oifs. But for now this is correct!

configs/components/oifs/oifs.yaml
JanStreffing: Agreed

configs/setups/awicm3/awicm3.yaml
joakimkjellsson: Sounds good to me. I want to use it, and I think it works, but I always struggled with the output. 
Let's make "false" the default. Some testing is needed to justify "true".

mandresm: @joakimkjellsson , this line is defined in `awicm3.yaml`, in the oifs section, meaning changing this line would not change the default you set in `oifs.yaml`. Do you want me to switch it to false also in the `oifs.yaml`?

configs/setups/awicm3/awicm3.yaml
```diff
@@ -1,196 +1,657 @@
...
+                        lsmssig: false
+                TL:
+                        nlat: "$(( ${res_number}+1 ))"
+                        lecompgrid: false
+                        lgradsp: true
+                        lassi: true
+                        lfdbop: true
+                        lsmssig: true
+
+        #@JAN/JOAKIM: is ok to ignore this files? anything else that needs to be removed?
```
JanStreffing, line 238: I'm not sure what this does. Cleanup after the run is done to save space? Maybe this is obsolete, since we link these from pool dir now?

configs/setups/awicm3/awicm3.yaml
joakimkjellsson: OpenIFS will not accept prime numbers as nproca (don't ask me why...). 
So any multiple of 2 works, or 279 = 31 * 9 works too. 
I often use 287 = 41 * 7, + 1 for rnfmap = 288 which fits well on 36, 48 or 96-core nodes.

configs/setups/awicm3/awicm3.yaml
joakimkjellsson: Do it! ;-)

configs/setups/awicm3/awicm3.yaml
```diff
@@ -1,196 +1,657 @@
...
+                rmp_f2a_G: ${fesom.nproc}/rmp_feom_to_${oifs.oasis_grid_name_a}_GAUSWGT.nc
+                rmp_a2r_B: ${fesom.nproc}/rmp_${oifs.oasis_grid_name_r}_to_RnfA_BICUBIC.nc
+                rmp_r2f_G: ${fesom.nproc}/rmp_RnfO_to_feom_GAUSWGT.nc
+
+
+# Set up CPU layout
+choose_cores_per_node:
+        24:
+                choose_resolution:
+                        T159_CORE2:
```
JanStreffing, line 576: Do we use these? If so they should be called TL or TCO + truncation number, rather than T + truncation number,

configs/setups/awicm3/awicm3.yaml
joakimkjellsson: No idea what these things are. Something about modifying the LSM? 
Yes, we link everything we need from the pool dirs, e.g. 159_4, ifsdata, etc.

configs/setups/awicm3/awicm3.yaml
joakimkjellsson: Yes you are right. We should actually be careful to always use Tco, TL, or Tq
Changing it will probably break some code, but we need to be strict with this to avoid confusing, even if we are mostly using Tco.

namelists/oifs/43r3/awicm3/202011/fort.4
```diff
@@ -0,0 +1,511 @@
...
+  SSPNAME        = 'historical',
+  CMIP6DATADIR   = '/work/ollie/jstreffi/input/oifs-43r3/TL159L60//../cmip6-data/',
+  LA4xCO2        = FALSE,
+  L1PCTCO2       = FALSE,
+  LCMIP5         = FALSE,
+  CMIP5DATADIR   = '/work/ollie/jstreffi/input/oifs-43r3/TL159L60//../cmip5-data/',
+  NCMIPFIXYR     = 1990,
+  NRCP           = 0,
+  LHVOLCA        = FALSE,
+  CRTABLEDIR     ='/work/ollie/jstreffi/runtime/awicm-3.1/DRH6/work//rtables/',
```
JanStreffing, line 59: perhaps the default can point to a pool dir already. I know it's changed later, but its a bit strange to have it point to an old exp in this file.

configs/setups/awicm3/awicm3.yaml
JanStreffing: Yeah, perhaps a comment in this line with a nod to the prime problems would be appropriate. Once you do MPI/OMP your nproca*OMP_NUM_THREADS has to fill a node. Then the rules are different again.

configs/setups/awicm3/awicm3.yaml
JanStreffing: To be clear, the comment by @mandresm  is about the l239 and below. It a bit strange how the github review comment shows only the lines above. l235 and l236 are dealing with TL vs TCO differences.

mandresm: This comment refers to lines 241-243:
```
        remove_input_sources:
                - rtables
                - vtables
```
It's an old comment I should have removed cause it is a correct action: as you know, ESM-Tools first reads `oifs.yaml` and then `awicm3.yaml`. In `oifs.yaml` the rtables and vtables are copied into the experiment directory (specified in `input_sources` list). However, for AWICM3 we don't want any of this copying, at least at the time I was working on that part, as the rtables origin paths are specified in the namelist. That is why I chose to remove the sources so that copying for AWICM3 does not happen. I think in reality we want to do the copying into the experiment folder, and then do the linking for each run folder but this is something I did not have time to address yet (I have a note about it somewhere from a talk with @JanStreffing ).

configs/setups/awicm3/awicm3.yaml
mandresm: You are right @JanStreffing , this is effectively doing nothing as `resolution` contains TL or TCO so the case is never met. I will change the cases.

mandresm: Actually I will comment this whole `choose_` block. We are still a step behind, where we are trying to get a simulation running in Ollie with TL159_CORE2 truncation. We can uncomment it and generalize it later for offering support for other computers and truncations.

configs/setups/awicm3/awicm3.yaml
JanStreffing: The last sentence somehow contradicts the one before? I agree with the one before and think vtables and rtables don't need to be copied them at all. I don't think they ever change. I wouldn't even know how to or what format they are in. Its not netcdf or grib or txt.

mandresm: Ok, I was not sure if they needed to be copied or not (I didn't remember). I knew that it was currently working as it was, so I wanted to look at my notes to confirm, but thanks for the clarification. They will never be copied :)

mandresm:
>     * github.com/FESOM/fesom2.git
>       Should be the default repository. The gitlab one is starting to fall behind and github has much better development features.

`juwels_chunk_io` is not a tag in GitHub.

mandresm:
All suggestions added.

PR #222, opened 2020-11-27
new chapter for ESM variables

mandresm:
The new chapter can be read here:
- https://esm-tools.readthedocs.io/en/dos-esm_variables/esm_variables.html

There are still some variables in my TODO list that I still need to add, but I'm running out of time, and I think I covered the most essential ones. Please, let me know if you are missing something relevant.

### DIFF
--- a/docs/esm_variables.rst
+++ b/docs/esm_variables.rst
@@ -1,0 +1,120 @@
+ESM-Tools Variables
+===================
+
+The `esm_parser` is used to read the multiple types of `YAML` files contained in `ESM-Tools`
+(i.e. model and coupling configuration files, machine configurations, runscripts, etc.). Each of
+these `YAML` files can contain two type of `YAML` elements/variables:
+
+  * **Tool-specific elements**: `YAML-scalars`, `lists` or `dictionaries` that include instructions and
+    information used by `ESM-Tools`. These elements are predefined inside the `esm_parser` or other
+    packages inside `ESM-Tools` and are used to control the `ESM-Tools` functionality.
+
+  * **Setup/model elements**: `YAML-scalars`, `lists` of `dictionaries` that contain
+    information defined in the model/setup config files (i.e. ``awicm.yaml``, ``fesom.yaml``, etc.).
+    This information is model/setup-specific and causes no effect unless it is combined with the
+    **tools-specific elements**. For example, in ``fesom.yaml`` for `FESOM-1.0` the variable
+    ``asforcing`` exists, however this means nothing to `ESM-Tools` by its own. In this case, this
+    variable is used in ``namelist_changes`` (a tool-specific element) to state the type of forcing
+    to be used and this is what actually makes a difference to the simulation. The advantage of
+    having this variable already defined and called in ``namelist_changes``, in the ``fesom.yaml``
+    is that the front-end user can simply change the forcing type by changing the value of
+    ``asforcing`` (no need for the front-end user to use ``namelist_changes``).
+
+The following subsection lists and describes the **Tool-specific elements** used to operate `ESM-Tools`.
+
+.. Note::
+   Most of the **Tool-specific elements** can be defined in any file (i.e. `configuration file`,
+   `runscript`, ...) and, if present in two files used by ESM-Tools at a time, the value is chosen
+   depending on the ESM-Tools file priority/read order (:ref:`yaml_hierarchy:YAML File Hierarchy`).
+   Ideally, you would like to declare as many elements as possible inside the `configuration files`,
+   to be used by default, and change them in the `runscripts` when necessary. However, it is ultimately
+   up to the user where to setup the Tool-specific elements.
+
+Tool-Specific Elements/Variables
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following keys should/can be provided inside configuration files for models
+(``<PATH>/esm_tools/configs/components/<name>/<name>.yaml``), coupled setups
+(``<PATH>/esm_tools/configs/setups/<name>/<name>.yaml``) and runscripts. You can find
+runscript templates in ``esm_tools/runscripts/templates/``).
+
+Installation variables
+----------------------
+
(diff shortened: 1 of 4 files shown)


docs/esm_variables.rst
```diff
@@ -0,0 +1,120 @@
...
+these `YAML` files can contain two type of `YAML` elements/variables:
+
+  * **Tool-specific elements**: `YAML-scalars`, `lists` or `dictionaries` that include instructions and
+    information used by `ESM-Tools`. These elements are predefined inside the `esm_parser` or other
+    packages inside `ESM-Tools` and are used to control the `ESM-Tools` functionality.
+
+  * **Setup/model elements**: `YAML-scalars`, `lists` of `dictionaries` that contain
+    information defined in the model/setup config files (i.e. ``awicm.yaml``, ``fesom.yaml``, etc.).
+    This information is model/setup-specific and causes no effect unless it is combined with the
+    **tools-specific elements**. For example, in ``fesom.yaml`` for `FESOM-1.0` the variable
```
nwieters, line 15: tools-specific elements

you named it above
tool-specific elements

docs/esm_variables.rst
```diff
@@ -0,0 +1,120 @@
...
+   :header: Key, Description
+   :widths: 15, 85
+
+   model,               "Name of the model/setup as listed in the config files (``esm_tools/configs/components`` for models and ``esm_tools/configs/setups`` for setups)."
+   setup_name,          Name of the coupled setup.
+   version,             "Version of the model/setup (one of the available options in the ``available_versions`` list)."
+   available_versions,  List of supported versions of the component or coupled setup.
+   git-repository,      Address of the model's git repository.
+   branch,              Branch from where to clone.
+   destination,         "Name of the folder where the model is download and compiled, in a coupled setup."
```
nwieters, line 54: downloaded

docs/esm_variables.rst
```diff
@@ -0,0 +1,120 @@
...
+   comp_command,        Command used to compile the component.
+   executable,          Name of the component executable file.
```
nwieters, line 56: I could imagine that it is unclear, wether a user can set/choose the name of the executable or just need to put here the name of executable, that the component gets through compilation. I think the latter case is true.

docs/esm_variables.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Runtime variables
+-----------------
+.. csv-table::
+   :header: Key, Description
+   :widths: 15, 85
+
+   account,             User account to be used to run the experiment.
```
nwieters, line 64: User account of the HPC system to be used to run the experiment.

docs/esm_variables.rst
```diff
@@ -0,0 +1,120 @@
...
+   model_dir,           "Absolute path of the model directory (where it was installed by `esm_master`)."
+   setup_dir,           "Absolute path of the setup directory (where it was installed by `esm_master`)."
+   compute_time,        "Estimated computing time for a run, used for submitting a job with `slurm`."
```
nwieters, line 67: Is this variable only needed for 'slurm' as job scheduler? I am not sure if any supported HPC at the moment uses another job scheduler. 
I would write 'job scheduler' instead of slurm.

docs/esm_variables.rst
```diff
@@ -0,0 +1,120 @@
...
+   time_step,           Time step of the component in seconds.
+   lresume,             Boolean to indicate whether the run is an initial run or a restart.
+   pool_dir,            Path to the pool directory.
```
nwieters, line 70: Path to the pool directory to read in ... data.

I think there are mesh data, forcing data, anything else?

docs/esm_variables.rst
```diff
@@ -0,0 +1,120 @@
...
+   namelists,           "List of namelist files required for the model."
+   namelist_changes,    "Functionality to handle changes in the namelists from the yaml files (see :ref:`yaml:Changing Namelists`)."
+   nproc,               Number of processors to use for the model.
+   nproca/nprocb,       "Number of processors for different MPI tasks/ranks. Incompatible with ``nproc``."
+   base_dir,            Path to the directory that will contain the experiment folder (where the experiment will be run and data will be stored).
+   post_processing,     Boolean to indicate whether to run postprocessing or not.
+   ":ref:`yaml:File Dictionaries`",     "`YAML` dictionaries used to handle input, output, forcing, logging, binary and restart files (see :ref:`yaml:File Dictionaries`)."
+   expid,               "ID of the experiment, also defined when calling ``esm_runscripts`` with the ``-e`` flag."
```
nwieters, line 78: ID of the experiment. This variable can also be defined when calling ``esm_runscripts`` with the ``-e`` flag.

docs/esm_variables.rst
```diff
@@ -0,0 +1,120 @@
...
+   ini_restart_exp_id,  "ID of the restarted experiment in case the current experiment has a different ``expid``."
+   ini_restart_dir,     "Path of the restarted experiment in case the current experiment runs in a different directory."
```
nwieters, line 80: I guess for the last two variables 'lresume' needs to be 'true'. This could be stated here also.

docs/esm_variables.rst
```diff
@@ -0,0 +1,120 @@
...
+   execution_command,   "Command for executing the component, including ``${executable}`` and the necessary flags."
+
+Calendar variables
+------------------
+.. csv-table::
+   :header: Key, Description
+   :widths: 15, 85
+
+   initial_date,        "Date of the beginning of the **simulation** in the format YYYY-MM-DD. If the simulation is a restart, ``initial_date`` marks the beginning of the restart."
```
nwieters, line 89: Is somewhere described what the difference between 'simulation' and 'run' is?

mandresm: It is kind of implicitly defined in the experiment folder structure section, do you think that is enough?

nwieters: Yes, I think we don't need to explain it here again.

PR #223, opened 2020-11-30
documentation for environment_, namelist_ and coupling_changes added

mandresm:
Changes can be read here:
- https://esm-tools.readthedocs.io/en/docs-changes_functionality/yaml.html#environment-and-namelist-changes-changes

### DIFF
--- a/docs/yaml.rst
+++ b/docs/yaml.rst
@@ -608,7 +608,7 @@ It is also possible to specify namelist changes to a particular section of a nam
                                     clonp: -0.008
                                     yr_perp: "remove_from_namelist"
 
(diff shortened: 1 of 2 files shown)


docs/yaml.rst
```diff
@@ -457,10 +457,139 @@ syear     Year from a given date.
...
+      We also see here how ``runtime_environment_changes`` is used to add nested
+      information about the output schedules for `FESOM` into an
+      ``AWI_FESOM_YAML`` variable that will be exported to the runtime environment.
+
 Changing Namelists
-~~~~~~~~~~~~~~~~~~
+------------------
 
-It is also possible to specify namelist changes to a particular section:
+It is also possible to specify namelist changes to a particular section of a namelist:
```
nwieters, line 592: Below the example you state that the namelist variable will be set to `.false.` but in the yaml file you say `false`. It occures probably somewhere else, but is it needed to state it in the yaml file without points? 
Are such boolean variables interpreted by python? I remember having problems with it, that these variables end up being upper case and I had to lowercase them all for a xml configuration file.

mandresm: Thanks for pointing out this :) , I added a paragraph after the example.

PR #250, opened 2020-12-14
Preventative bug message for 5.0.5: instructions for installing a plugin

pgierz:
@dbarbi, can you please check if this is correct? I wasn't sure about the syntax exactly. It might be nice to allow MOTD to be configured to use a "test branch" before merging into the main one...

### DIFF
--- a/esm_tools/motd/motd.yaml
+++ b/esm_tools/motd/motd.yaml
@@ -2,48 +2,57 @@ release5_esm_tools:
         package: "esm_tools"
         versions: "<4.0.0"
         message: |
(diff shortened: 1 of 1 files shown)


dbarbi (approved):
Ok. I don't like it at all though as it comes up with every call to esm_master and esm_runscripts. Can we include that plugin into the core so that we can get rid of the message again?

dbarbi:
Merged this by hand

pgierz:
The code for it is here: https://github.com/pgierz/gfw_creator

PR #253, opened 2020-12-15
fix: rad streams in echam and hyper off on DKRZ

pgierz:
Rad streams where not properly moved to next compute job, led to model crash after first year.

Also, hyper threading is off now on MISTRAL by default, as it induces longer wall times.

@mandresm, can you please put this in develop too?

### DIFF
--- a/configs/machines/mistral.yaml
+++ b/configs/machines/mistral.yaml
@@ -5,6 +5,7 @@ name: mistral
 additional_flags: "--mem=0"
 account: None
 
+hyperthreading_flag: "--ntasks-per-core=1"
 choose_general.use_hyperthreading:
         "1":
                 hyperthreading_flag: ""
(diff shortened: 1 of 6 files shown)


configs/machines/mistral.yaml
```diff
@@ -6,14 +6,14 @@ additional_flags: "--mem=0"
...
-                hyperthreading_flag: "--ntasks-per-core=1"
-        "*":
+        "1":
                 hyperthreading_flag: ""
+        1:
+                hyperthreading_flag: ""
+        True:
+                hyperthreading_flag: ""
+        "*":
+                hyperthreading_flag: "--ntasks-per-core=1"
```
mandresm: Have you tried this lines on `mistral`? My guess is that they will break as `1` and `True` are considered as the same once evaluated by the parser and tagged as duplicates. `"1"` and `True` should cause no issue.

pgierz: Just tried it now again, it did not seem to complain....but it also didn't turn the flag on, I'll send you another commit in a bit.

mandresm: The flag does not need to be on for it to complain. So I would leave it as it is (including the `1`).

pgierz: OK, I fixed it now...

Also, the "tidy done" that I had was wrong earlier, the "maybe resubmit" kills off any steps that might happen afterward.

PR #266, opened 2021-01-11
Hotfix/fix lang issues

denizural:
After Dirk's comments on warning the users about the Linux settings before installing the esm-tools from scratch, I first added some checks to the `install.sh`. We can add more checks similar to that format later on.

I also changed all occurences of the LC_ALL and LANG variable settings in the rst document files and YAML files. Although it does not matter, the correct format is capital UTF8 (not utf8). I changed them just to be on the safe side.


runscripts/echam/echam-blogin-initial-yearly.yaml
```diff
@@ -8,6 +8,7 @@ general:
         nyear: 1                        # Number of years per run
         #model_dir: /home/shkifmsw/esm/models
         model_dir: /home/shkifmsw/esm/models/echam-${echam.version}
+        use_venv: False
```
mandresm: I agree with all the changes. However, I will remove the following line in the `echam-blogin-initial-yearly.yaml` so that the dialog box about the virtual environment shows up and teaches the user on how to use or not the virtual environment.

mandresm (changes requested):

runscripts/echam/echam-blogin-initial-yearly.yaml
denizural: That venv stuff is accidental. I don't know how it came there but when I commited the changes it went there.

PR #267, opened 2021-01-11
comment out namelist.diag

denizural:
Hi @mandresm, I think FESOM-2 does not have namelist.diag anymore, right? I am getting some errors with AWICM when it uses FESOM 2 (eg. AWICM-2). Could you please have a look at that.

FESOM 2 GitHub:
https://github.com/FESOM/fesom2/tree/master/config


mandresm (changes requested):
I think that is again an issue with the versioning of FESOM. First, I think it would be good to understand if there is any of the FESOM-2 version that we support that needs `namelist.diag`. If there is not we can go ahead and remove it, if there is we will need to added into the corresponding `choose_version` case (`add_namelist: [namelist.diag]`).

mandresm:
I checked that the `namelist.diag` does not exist in any FESOM-2 version and therefore can be safely removed without the need of a `choose_version` option.

PR #301, opened 2021-03-19
Feature/awicm3 refac esm environment

mandresm:


seb-wahl:
I'm fine with all the modifications except for the fact that you silently replaced "our" paths on juwels (e.g. `IO_LIB_ROOT` in `oifs...preprocess.sh`, compiler settings in`juwels.yaml`) on juwels with Jan's settings which will break things for us. I don't have a good solution on how to overcome this as we have to set SOME defaults.

mandresm:
Hi @seb-wahl, I thought about this and that we need to discuss this when we are ready to merge this branch into `geomar_awicm3` (I think we are not there yet), but when we are, we can organize another merging party to discuss all the conflicts.

PR #309, opened 2021-04-08
Awicm3 master

JanStreffing:
This allows us to run the lastest master branches that contain physics fixes for awicm3 from Tido Semmler.


JanStreffing:
This is the first time I implement new branches for awicm3. I'd like to have a style review.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -30,7 +34,7 @@ general:
             add_include_models:
             - xios
 
-        version: "1.0-deck"
+        versions: "1.0-deck"
```
mandresm, line 37: We need to define a default `version`. Which version is entirely up to you, but a default needs to be given, otherwise this can result into problems for ESM-Tools I think

denizural, line 37: Does it work really with `version` instead of `versions`?

configs/setups/awicm3/awicm3.yaml
```diff
@@ -294,13 +299,21 @@ fesom:
         # Sets kkp_kv0 in namelist.cvmix to 5e-2 for the 1.0-deck version to be able to reproduce
         # old experiments. 5e-3 is a better value so it is used for XIOS version
         choose_general.version:
+                "1.0-master":
+                        model_tag: "awicm-3-master"
+                        # namelist_dir: "${model_dir}/fesom-2.0-master/config/" TODO: this is the line we actually want to use
```
mandresm: `model_dir` during run time refers to the folder of that particular component (i.e. `model_dir` in this case should be `<path_to_model_folder>/awicm3-1.0-master/fesom-2.0-master`), each component has its own `model_dir`. So I would suggest to write this:
```
namelist_dir: "${model_dir}/config/"
```

configs/setups/awicm3/awicm3.yaml
JanStreffing: no that looks like a typo. strange that is caused no error. but im not using 1.0-deck here. I'll fix it

namelists/fesom2/awicm3/awicm-3-master/namelist.io
```diff
@@ -0,0 +1,45 @@
...
+ldiag_dMOC       =.false.
+ldiag_DVD        =.false.
+ldiag_forc       =.false.
+/
+
+&nml_listsize
+io_listsize=100 !number of streams to allocate. shallbe large or equal to the number of streams in &nml_list
+/
+
+# for sea ice related variables use_ice should be true, otherewise there will be no output
```
denizural: This is a FORTRAN namelist, right? I think you need `!` instead of `#`. Be careful

JanStreffing: In this case it's the namelist I got from the fesom2 source code. Fixed here: https://github.com/FESOM/fesom2/pull/92/files
I'll fix this in the esm-tools after it's merged there.

The idea of this branch is that we can use their namelists directly.

configs/components/oifs/oifs.yaml
```diff
@@ -100,6 +101,15 @@ compile_infos:
             branch: foci_conserv
             comp_command: cd make; ../fcm/bin/fcm make -v -j24 -f oifs.cfg
             destination: 40r1
+          43r3-master:
+            requires:
+            - oasis3mct-4.0-awicm3-frontiers
+            branch: master
+            comp_command: export OIFS_TOPLEVEL_DIR=${model_dir}; cd make; ../fcm/bin/fcm make
+              -v -j8 -f oifs.cfg ; cp -p esm/oifs/bin/master.exe esm/oifs/bin/oifs
```
denizural: I did not understand the copying of the binary `master.exe` to `oifs`. Is there a reason for the `master.exe` to stay in the same directory? What about linking? I never used OIFS, btw.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -17,10 +17,14 @@ general:
                 - rnfmap
                 - oasis3mct
 
-        available_versions:
+        available_version:
```
mandresm: This is incorrect, it should be `available_versions`.

mandresm (changes requested):

configs/components/oifs/oifs.yaml
JanStreffing: probably better to mv actually. Less confusion about why there might be two executable. I'll fix that.

configs/setups/awicm3/awicm3.yaml
denizural: 👍

JanStreffing:
ohh.. I just found something bad.
we are downloading fesom2 from gitlab instead of github. When working on older branches or tags that makes no difference, but for the latest master branch it does.

This master branch is from 25. of February. Let's me see what happens if I change that...

mandresm:
> ohh.. I just found something bad.
> we are downloading fesom2 from gitlab instead of github. When working on older branches or tags that makes no difference, but for the latest master branch it does.
> 
> This master branch is from 25. of February. Let's me see what happens if I change that...

That is not the case for the deck version, there we use `fesom-2.0-jio` version. In this version you see we specifically define `git-repository` to be the GitHub one. I will copy that inside `fesom-2.0-master` version.

JanStreffing:
I would suggest changing the default to github instead. Nobody should start working on gitlab anymore. Only those that specifically need old branches from there should select it.

JanStreffing:
I'm testing again with the suggested changes and trying to run straight from the fesom source code folder namelists, as well as with the latest version from github. 

Ty for the feedback so far.

mandresm:
> I would suggest changing the default to github instead. Nobody should start working on gitlab anymore. Only those that specifically need old branches from there should select it.

I don't agree. I am not sure if there is a particular reason why people is using gitlab in `awicm2` or `awiesm2`. We should not change this kind of defaults without some agreement with the parts involved, cause it can potentially break things that were already working, and we won't realize until we merge into develop, and by that time we probably forgot that we made this major change in fesom.

denizural:
+1 from my side. There should be an agreement on which repository to use. Otherwise, I am sure that there will be a lot of confusion at later stages.

JanStreffing:
Agreed, let's talk about it Monday morning. I think there could be significant benefit for all of us in using the fesom2 repository that the core fesom2 development is taking place on. I see that these is still significant activity regarding fesom2 recom on gitlab as well.

configs/components/fesom/fesom-2.0.yaml
```diff
@@ -13,11 +13,14 @@ choose_version:
...
     git-repository:
-    - https://github.com/FESOM/fesom2.git
     - https://gitlab.dkrz.de/FESOM/fesom2.git
+    - https://github.com/FESOM/fesom2.git
     install_bins: bin/fesom.x
   2.0-master:
     branch: master
+    destination: fesom-2.0
+    git-repository:
+    - https://gitlab.dkrz.de/FESOM/fesom2.git
```
mandresm: Here you want the github one right?

JanStreffing:
Ok i found some time to finish this. I think it's ready for merging.

PR #311, opened 2021-04-09
Aleph machine file

JanStreffing:
I think this should be a clean merge. After that I can continue branching off an awicm-3-frontiers branch that will contain a feature that is very similar to awicm-3-master. That branch can then be used to test PBS on Aleph not just with a testcase but with the real model.

### DIFF
--- a/configs/components/oifs/oifs.env.yaml
+++ b/configs/components/oifs/oifs.env.yaml
@@ -257,10 +257,11 @@ compiletime_environment_changes:
         - 'OIFS_NETCDFF_INCLUDE=-I$NETCDF_DIR/include'
         - 'OIFS_NETCDFF_LIB="-L$NETCDF_DIR/lib -lnetcdff"'
         - 'OIFS_FC=${computer.fc}'
-        - 'OIFS_FFLAGS="-align array32byte -O2 -h cpu=x86-skylake -h byteswapio"'   #TODO: Test: -O3 -hfp3
+        - 'OIFS_FFLAGS="-emf -O2 -h cpu=x86-skylake -h byteswapio"'   #TODO: Test: -O3 -hfp3
+        - 'OASIS_FFLAGS=$OIFS_FFLAGS'
         - 'OIFS_FFIXED="-ffixed"'
         - 'OIFS_FCDEFS="BLAS LITTLE LINUX INTEGER_IS_INT"'
         - 'OIFS_LFLAGS='
         - 'OIFS_CC=${computer.cc}'
-        - 'OIFS_CFLAGS="-O2 -h cpu=x86-skylake"'
+        - 'OIFS_CFLAGS="-emf -O2 -h cpu=x86-skylake"'
         - 'OIFS_CCDEFS="LINUX LITTLE INTEGER_IS_INT _ABI64 BLAS"'
--- a/configs/components/oifs/oifs.env.yaml
+++ b/configs/components/oifs/oifs.env.yaml
@@ -257,10 +257,10 @@ compiletime_environment_changes:
         - 'OIFS_NETCDFF_INCLUDE=-I$NETCDF_DIR/include'
         - 'OIFS_NETCDFF_LIB="-L$NETCDF_DIR/lib -lnetcdff"'
         - 'OIFS_FC=${computer.fc}'
-        - 'OIFS_FFLAGS="-r8 -fp-model precise -align array32byte -O3 -qopenmp -xCORE_AVX2 -g -traceback -convert big_endian"'
-        - 'OIFS_FFIXED=""'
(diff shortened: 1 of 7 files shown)


configs/machines/aleph.yaml
```diff
@@ -0,0 +1,124 @@
...
+        "serial":
+                cores_per_node: 1
+
+logical_cpus_per_core: 2
+
+threads_per_core: 1
+pool_dir: "/proj/awi"
+
+pool_directories:
+        pool: "/proj/awi"
```
denizural: Not only for this file but I think that we need some form of consistency for the directory paths. I opt for using a directory separator (`/`) at the end of the paths. I realized that sometimes there are errors due to a missing slash at the end. What do you think @mandresm?

configs/machines/aleph.yaml
```diff
@@ -0,0 +1,124 @@
...
+hyper_flag: ""
+
+# standard environment setup
+#
+#
+
+useMPI: cray_mpich
+
+module_actions:
+        - "load ${PrgEnv}"
```
denizural, line 48: @JanStreffing, what about starting with a `module purge` here? Maybe it is not necessary but at least it is a measure against conflicting module loads.

configs/machines/aleph.yaml
```diff
@@ -0,0 +1,124 @@
...
+                PrgEnv: PrgEnv-intel
+
+                fc: '"mpiifort -mkl"'
+                f77: '"mpiifort -mkl"'
+                mpifc: mpiifort
+                mpicc: mpiicc
+                cc: mpiicc
+                cxx: mpiicpc
+
+                launcher_flags: "-n @nproc@ -N @nproc_per_node@ -d @cpus_per_proc
```
denizural: **Typo alert!** missing `@`

configs/machines/aleph.yaml
```diff
@@ -0,0 +1,124 @@
...
+                        env OMP_NUM_THREADS=@omp_num_threads@ -cc depth"
+
+                add_export_vars:
+                        MPIROOT: ${I_MPI_ROOT}/intel64
+                        KMP_AFFINITY: disabled
+        cray_mpich:
+                PrgEnv: PrgEnv-cray/6.0.4
```
denizural, line 114: Is `PrgEnv-cray/6.0.4` a valid directory?

configs/components/oifs/oifs.env.yaml
```diff
@@ -232,3 +236,31 @@ compiletime_environment_changes:
         - 'OIFS_CC=${computer.cc}'
         - 'OIFS_CFLAGS="-fp-model precise -O3 -xCORE_AVX2 -g -traceback -qopt-report=0"'
```
denizural, line 237: I see that quite often. What is the point of mixing `-O3` (aggressive optimization) and `-traceback` (runtime error stuff) and `-g` (debug). Any thoughts? Maybe from @dbarbi ?

configs/machines/aleph.yaml
JanStreffing: That wont work on Aleph. There are modules that you cant get back after you purge. I had a lot of trouble with this in the beginning.

configs/machines/aleph.yaml
JanStreffing: Its a module name I guess. right @mandresm ?

configs/components/oifs/oifs.env.yaml
JanStreffing: Make it fast, but if if does crash you know where from. I think -traceback -g are not very expensive options. The -O flags do a lot more in terms of speed. Especially O0 to O2 on intel and O0 to O1 on cray.

JanStreffing:
Ty for the comments @denizural !

configs/components/oifs/oifs.env.yaml
denizural: Here is a nice read: https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/compilation/debugging/debugging-and-optimizations.html

> Note that if you specify an O1, O2, or O3 option with the -g (Linux* and macOS*) or /debug:full (Windows*) option, some of the debugging information returned may be inaccurate as a side-effect of optimization.

configs/machines/aleph.yaml
denizural: My bad, you are right. I had to use the search function before jumping in 🤦‍♂️

configs/machines/aleph.yaml
JanStreffing: Always saver to add them. I'll do so here.

PR #315, opened 2021-04-13
Changeing fesom2 default repo for future branches

### LINKED ISSUE #314: FESOM2 default repository
I posed the question of the FESOM2 default repository on the FESOM2 repo:
https://github.com/FESOM/fesom2/issues/94
To quote @koldunovn:

- Q: Is still this reflective of the policy of the FESOM developers? Has the default changed to github?
A: The policy is now that the default is github, gitlab is only synced sporadically. We have not decided on gitlab's fate yet. I would close it, but there are people who still do work there, so it should be discussed.

- Q: Are there plans to automatically sync the two, such that we would not even have to make a choice?
A: Currently it happens by hand, so not very reliable. Even if it would happen automatically, I would still put github as a default, since the number of lucky people who have DKRZ account in this world is fairly limited :)

- Q: Or are there branches on gitlab that are not supposed to go public?
A: I don't know, but I see that not everyone has moved to github jet, maybe they have their reasons.

From this I would suggest the following: We should probably change the default to github, as the fate of gitlab is in question and it is not considered default by FESOM developers.

@dbarbi You mentioned that setting up automatic synchronization is easy to set up. Maybe @koldunovn would like to know?

For a practical implementation that does not mess with any existing setups I would suggest to:
- Switch the default to github
- Remove github from individual branch checkouts
- Add gitlab to all branch checkouts that did not have github so far

This way nothing should change for any existing esm_tools FESOM2 users. Only when new FESOM2 branch selections are made in fesom-2.0.yaml the default well be github in the future. This way we might be able to avoid having to get feedback from every single fesom2 user.

What do you think?

JanStreffing:
The intent is not to change anything for existing branches. Only the default for future developments. Linked to #314

### DIFF
--- a/configs/components/fesom/fesom-2.0.yaml
+++ b/configs/components/fesom/fesom-2.0.yaml
@@ -13,8 +13,8 @@ choose_version:
   '2.0':
     branch: 2.0.2
     git-repository:
-    - https://gitlab.dkrz.de/FESOM/fesom2.git
     - https://github.com/FESOM/fesom2.git
+    - https://gitlab.dkrz.de/FESOM/fesom2.git
     install_bins: bin/fesom.x
   2.0-esm_interface:
     branch: using_esm-interface_yac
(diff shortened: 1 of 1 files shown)


koldunovn (approved):
Make sense :)

mandresm (approved):

PR #328, opened 2021-05-07
suggested changes for wiso

mandresm:
@xshi-awi , I made some suggestions in this pull request for AWIESM-2.1-WISO. Feel free to merge this into your `wiso` branch if you agree with them. Please, see the comments below.

None of these changes are fixes, what you have done was already working, the changes suggested here are done just with the aim of shortening the runscript.

### DIFF
--- a/configs/components/echam/echam.yaml
+++ b/configs/components/echam/echam.yaml
@@ -231,6 +231,23 @@ choose_version:
                                         lrad_async: "${lrad_async}"
                                         lrestart_from_old: "${lrestart_from_old}"
 
+        6.3.05p2-wiso:
+                namelist_changes:
+                        namelist.echam:
+                                parctl:
+                                        nproca: "${nproca}"
+                                        nprocb: "${nprocb}"
+                                        nprocar: "${nproar}"
+                                        nprocbr: "${nprocbr}"
+                                runctl:
+                                        default_output: True
+                                radctl:
+                                        lrad_async: True
+                                wisoctl:
+                                        lwiso_rerun: False
+                                        lwiso: False
+                                        nwiso: 0
+
         6.3.05p2-concurrent_radiation:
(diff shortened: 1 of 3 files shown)


configs/components/echam/echam.yaml
```diff
@@ -231,6 +231,23 @@ choose_version:
...
+                                        nprocar: "${nproar}"
+                                        nprocbr: "${nprocbr}"
+                                runctl:
+                                        default_output: True
+                                radctl:
+                                        lrad_async: True
+                                wisoctl:
+                                        lwiso_rerun: False
+                                        lwiso: False
+                                        nwiso: 0
```
mandresm: I removed these lines from your runscript and place them into the `echam.yaml`. If this are defaults for `echam-wiso` they should be in the `echam.yaml` under the 6.3.05p2-wiso section, so that the runscript can be a bit shorter. 

Now you can control the nproc variables directly from the runscript without havinf to use namelist_changes in the echam section.

runscripts/awiesm/PI_ctrl_awiesm-2.1-wiso_ollie.yaml
```diff
@@ -58,11 +43,13 @@ fesom:
     ini_parent_exp_id: piw_zstar_swon 
     ini_parent_dir: "/home/ollie/xshi/work/AWIESM_2_1_LR_concurrent_rad/piw_zstar_swon/restart/fesom/"
     ini_parent_date: "2681-12-31"
+
+    climate_data_dir: "/home/ollie/xshi/work/input/hydrography/"
```
mandresm, line 47: The variable climate_data_dir is already given to namelist_changes inside the fesom-2.0.yaml therefore you don't need to include it inside namelist_changes in your namelist.

PR #332, opened 2021-05-11
Default output in awiesm

pgierz:
There has always been ongoing discussion regarding the "default output".

At least for AWI-ESM, the department internal recommendation is to set the following in `namelist.echam`:

```fortran
&runctl
    ...
    default_output = .true.
    ...
/
```
To avoid me needing to fix that by hand for each of the students, it could be default (at least for AWIESM). Clever people can turn it back off again.


dbarbi:
will merge this by hand now

Issue #335, opened 2021-05-14
fesom2 warning namelist.diag somepath pool data not found 


chrisdane:
Hi

**Describe the bug**
Running
```
cd ~/esm/runscripts
esm_runscripts fesom2-mistral-initial-monthly.yaml -e test -c
```
yields the warnings
```
WARNING: File not found: /mnt/lustre01/pf/a/a270073/esm/esm_tools/namelists/fesom2/namelist.diag
WARNING: File not found: /work/ab0246/a270073/out/fesom-2.0/test/run_20010101-20010131/config/fesom/namelist.diag
MISSING FILES:

- missing source: /mnt/lustre01/pf/a/a270073/esm/esm_tools/namelists/fesom2/namelist.diag
- missing target: /work/ab0246/a270073/out/fesom-2.0/test/run_20010101-20010131/config/fesom/namelist.diag

- missing source: /work/ab0246/a270073/out/fesom-2.0/test/run_20010101-20010131/config/fesom/namelist.diag
- missing target: /work/ab0246/a270073/out/fesom-2.0/test/run_20010101-20010131/work/namelist.diag

- missing source: somepath
- missing target: opbndpath (from namelist.config in FESOM)

- missing source: /pool/data/AWICM/FESOM2/hydrography/
- missing target: climatedatapath (from namelist.config in FESOM)

- missing source: /pool/data/AWICM/FESOM2/forcing/
- missing target: forcingdatapath (from namelist.config in FESOM)

- missing source: somepath
- missing target: tideforcingpath (from namelist.config in FESOM)
```

**To Reproduce**
```
cd ~/esm/runscripts
esm_runscripts fesom2-mistral-initial-monthly.yaml -e test -c
```

**Expected behavior**
No warnings.

**System (please complete the following information):**
 - Supercomputer: mistral
 - Version

| package_name        | version   | file                                                          | branch   | tags   |
|---------------------|-----------|---------------------------------------------------------------|----------|--------|
| esm_analysis        | 0.3.0     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |          |        |
| esm_archiving       | 0.1.0     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |          |        |
| esm_calendar        | 5.0.0     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |          |        |
[...]

PR #337, opened 2021-05-17
namelist.diag removed from fesom-2.0.yaml

mandresm:
Solves #335 

@koldunovn , could you please confirm that `namelist.diag` is not anymore need it in FESOM version `2.0`?


koldunovn:
@ yes I can confirm that, at least for stand alone version. I also don't think it's something needed for the coupled, but I would double check with @JanStreffing

koldunovn (approved):

JanStreffing:
There is no additional use of namelist.diag for AWI-CM3.

PR #342, opened 2021-05-18
fix(awiesm): salt restoring should be off in coupled mode

pgierz:
After checking with @dsidoren and @koldunovn, the default switch for coupled mode in AWIESM should be **off**. 

Relevant FORTRAN namelist setting:
```
&oce_tra
    surf_relax_s = 0
```

This PR sets that as a default for AWI-ESM 2.1 and AWI-ESM 2.2.


dbarbi (approved):

dbarbi:
Will merge this by hand now

PR #348, opened 2021-05-28
fix(fesom2): adds GM parameters based upon selected ALE scheme

pgierz:
Based upon email exchange with Xiaoxu Shi (@xshi-awi), I have included the
recommended GM settings from @dsidoren based upon the user selected ALE scheme. 

Xiaoxu, could you check please if that agrees with your version?  @dbarbi, once
this checks out, can we also include a warning in the MOTD to make sure no one
is running with bad settings?


dbarbi:
will merge this by hand now

PR #349, opened 2021-05-31
fix(awiesm): sets nproca/b to 24/24 and turns off fast rad

pgierz:
We have found that the nproca/b settings of 24/18 yield non-reproducible results. This changes the default to a safer 24/24 variant, and turns off the (as of now) broken fast radiation by setting the number of processors to 0 for the radiation cores


dbarbi:
I will merge this by hand now

PR #353, opened 2021-06-03
Awicm 3 frontiers xios

mandresm:
Merging `xios` branch to the main branch.


configs/components/oifs/oifs.yaml
```diff
@@ -146,7 +157,7 @@ compile_infos:
             comp_command: "export OIFS_TOPLEVEL_DIR=${model_dir}; export OIFS_XIOS=enable ; export OIFS_XIOS_DIR=${model_dir}/../xios ; export OIFS_XIOS_INCLUDE=-I/${model_dir}/../xios/inc/ ; cd make; ../fcm/bin/fcm make -v -j8 -f oifs.fcm ; mv esm/oifs/bin/master.exe esm/oifs/bin/oifs"
             git-repository: https://gitlab.dkrz.de/ec-earth/oifs-43r3.git
             install_bins: make/esm/oifs/bin/oifs
-            destination: oifs-43r3
+            #destination: oifs-43r3 TODO: JS: Had to comment this out. When I use destination the $model_dir changes, but $XIOS_TOPLEVEL is still pointing at oifs-43r3-awi-e1 leading to failed oasis lib including
```
JanStreffing, line 160: @mandresm Do you know why I had to do this?

configs/components/xios/xios.yaml
```diff
@@ -36,7 +36,12 @@ choose_version:
     use_oasis: ''
   "2.5":
     git-repository: https://gitlab.dkrz.de/ec-earth/xios-2.5
-    archfile: ESMTOOLS_generic_oasis_intel
+    #choose_computer.useMPI:
+    #  cray_mpich:
+    #    archfile: ESMTOOLS_generic_oasis_cray
+    #  intelmpi:
+    #    archfile: ESMTOOLS_generic_oasis_intel
+    archfile: ESMTOOLS_generic_oasis_cray # TODO: fix and use choose block above
```
JanStreffing, line 44: I think this needs to be worked on, because it's hardcoded for cray maschines at the moment.

configs/components/xios/xios.yaml
```diff
@@ -98,6 +103,12 @@ choose_general.setup_name:
                         file_def: file_def
         awicm3:
                 xml_dir: "${pool_dir}"
+choose_computer.name:
+        ollie:
+                add_environment_changes:
+                        add_module_actions:
+                                - "unload netcdf; load netcdf/4.4.1.1_intel_mpi"
+
```
JanStreffing, line 111: This seems like a strange addition for the XIOS merge, since I did most of my work on XIOS on Aleph. I did first compile XIOS, without integrating it into OpenIFS on Ollie. So maybe this was needed?

runscripts/awicm3/awicm3-aleph-TL159L60-CORE2.yaml
```diff
@@ -2,14 +2,14 @@ general:
...
+    compute_time: "00:30:00"
     initial_date: "1990-01-01"
-    final_date: "1990-01-01"
+    final_date: "1991-01-01"
     base_dir: /scratch/awiiccp2/
     nday: 1
     nmonth: 0
     nyear: 0
-    taskset: false
+    heterogeneous_parallelization: true
```
JanStreffing, line 12: May need a sanity check, if this is the right name for the option in the target branch.

JanStreffing (changes requested):

configs/components/oifs/oifs.yaml
mandresm: Nope, I'll have a look at that before merging.

configs/components/xios/xios.yaml
mandresm: Probably this was needed, but this won't go into the environment of AWICM3 as we decided to go there with an environment that specifically ignores environments in the components and only takes the environments specified in awicm3.yaml. 

We could change that of course by either not ignoring component environments, or (the easiest solution) getting all the environment from `xios.yaml` and put them into an `xios.env.yaml` and then loading it with a `further_reading` in `awicm3.yaml`.

runscripts/awicm3/awicm3-aleph-TL159L60-CORE2.yaml
mandresm: No it's not, it needs to go on computer. I'll take care of that. Should the value be true?

configs/components/xios/xios.yaml
JanStreffing: I think the second solution might be easier for now (moving this to `awicm3.env.yaml`). I suspect XIOS will not be used by any other models on ollie.

runscripts/awicm3/awicm3-aleph-TL159L60-CORE2.yaml
JanStreffing: Yeah I think true a good default for aleph. We know it works well on PBS.

configs/components/oifs/oifs.yaml
JanStreffing: So I can recall that this had something to do with the name of the folder that the component was installed into not matching with where it was looking for it elsewhere. But I forgot what that elsewhere was..

mandresm: Jan, this is not for the branch we typically use for AWICM3 right? (for that we use `oifs-43r3-awi-frontiers`, this one is `43r3-awi-e1`). Is this just for standalone `OpenIFS` with `xios`?

JanStreffing: I think so, yes. It was part of the first commit where I was testing installing XIOS on ollie.

PR #360, opened 2021-06-09
Openmp

mandresm:


configs/components/fesom/fesom-2.0.yaml
```diff
@@ -78,7 +78,7 @@ bin_dir: "${setup_dir}/bin"
 climate_data_dir: "${pool_dir}/hydrography/"
 forcing_data_dir: "${pool_dir}/forcing/"
 ini_data_dir: "${pool_dir}/pool-data/"
-namelist_dir: "${esm_namelist_dir}/fesom2/"
+namelist_dir: "${model_dir}/fesom-2.0/config/"
```
JanStreffing: I think we don't want this as default. Only for the fesom-2.0-master

configs/components/fesom/fesom.yaml
```diff
@@ -65,7 +65,7 @@ bin_dir: "${setup_dir}/bin"
 climate_data_dir: "${pool_dir}/hydrography/"
 forcing_data_dir: "${pool_dir}/forcing/"
 ini_data_dir: "${pool_dir}/pool-data/"
-namelist_dir: "${esm_namelist_dir}/${model}"
+namelist_dir: "${model_dir}/config/"
```
JanStreffing: As above

configs/components/oifs/oifs-43r3-preprocess.sh
```diff
@@ -52,6 +52,7 @@ elif [[ "$(hostname -f)" =~ juwels ]] ; then
     module load ParaStationMPI/5.4.4-1-mt    
     module load Python/3.6.8
     module load imkl/2019.3.199
+    export PATH=$IO_LIB_ROOT/bin:$PATH:/p/project/chhb19/jstreffi/software/cdo/cdo-1.9.8/src/
```
JanStreffing, line 55: We can keep this for now, but the restrictive folder access policy of the JSC means that this will not work for users that are not part of project chhb19. Unfortunately JSC has been uncooperative in installing software packages system-wide for us. There is no better solution at the moment. AFAIK the CDO in the latest stages does not work as we need it.

JanStreffing: Maybe we can add a comment that clarifies to future users, that the may have to install their own cdo and change this path to point towards that installation.

JanStreffing: Alternatively all paths that are not accessible to all users on Juwels can go into the exp definition yaml. 
In fact I like that idea a lot more, since it put the variable that likely needs to be changed at the top user interface level!

configs/machines/juwels.yaml
```diff
@@ -94,7 +112,7 @@ choose_compiler_mpi:
         NETCDF_Fortran_LIBRARIES: $EBROOTNETCDFMINFORTRAN/lib
         ECCODESROOT: /p/project/chhb19/jstreffi/ecmwf/eccodes_intel2020_impi2020
         LD_LIBRARY_PATH: $ECCODESROOT/lib64:$LD_LIBRARY_PATH
-
+        PATH: /p/scratch/chhb19/hhb193/miniconda3/bin/:$PATH
```
JanStreffing, line 115: As above

configs/machines/juwels.yaml
```diff
@@ -114,12 +132,14 @@ choose_compiler_mpi:
...
         CC: mpicc
         CXX: mpic++
+        MAIN_LDFLAGS: -openmp
         MPIROOT: "\"$(mpifort -show | perl -lne 'm{ -I(.*?)/include } and print $1')\""
         MPI_LIB: "\"$(mpifort -show |sed -e 's/^[^ ]*//' -e 's/-[I][^ ]*//g')\""
         NETCDF_Fortran_INCLUDE_DIRECTORIES: $EBROOTNETCDFMINFORTRAN/include
         NETCDF_Fortran_LIBRARIES: $EBROOTNETCDFMINFORTRAN/lib
         ECCODESROOT: /p/project/chhb19/jstreffi/ecmwf/eccodes_intel2020_psi2020
         LD_LIBRARY_PATH: $ECCODESROOT/lib64:$LD_LIBRARY_PATH
+        PATH: /p/scratch/chhb19/hhb193/miniconda3/bin/:$PATH
```
JanStreffing, line 142: As above

JanStreffing (changes requested):

mandresm:
@JanStreffing can you have a look at the changes in the `frontiers` namelists and let me know if you want these changes? Also in the `runscripts/awicm3/awicm3-juwels-TL159L60-CORE2.yaml`.

JanStreffing:
> @JanStreffing can you have a look at the changes in the `frontiers` namelists and let me know if you want these changes? Also in the `runscripts/awicm3/awicm3-juwels-TL159L60-CORE2.yaml`.

I dont think so. The rotation change is because we switched from a rotated to an unrotated mesh between deck and frontiers. So now FESOM2 has to do the rotation.

The cavity stuff is new to fesom and did not exist for deck.

It would be good to have better names for the runscripts.
`awicm3-juwels-TL159L60-CORE2.yaml` is should be `awicm3-deck-juwels-TL159L60-CORE2.yaml`, contrasting with to `awicm3-frontiers-juwels-TL159L60-CORE2.yaml`

Otherwise people will assume that `awicm3-juwels-TL159L60-CORE2.yaml` is the default and start messing around with old scripts.

PR #366, opened 2021-06-15
Oifs eternal restarts

mandresm:
- **eternal restart feature** working at least for daily and monthly restarts, not necessarily 1, but 3 days restarts can also be run
- **cleanup** of the `oifs.yaml` and `awicm3.yaml`:
    - preprocessing is now in `oifs.yaml` (split preprocessing scripts) and removed from `awicm3.yaml`
    - compute recipe in `awicm3.yaml` is now loaded from `oifs.yaml`
- **preprocessing scripts (`change_icm_date.sh`) now load the `env.sh`** produced by `esm_environment`

Tested in `Ollie`

Closes AWI-CM3/project_management#7


configs/components/oifs/change_icm_date.sh
```diff
@@ -63,17 +31,28 @@ cdo -V
...
 echo " Output dir: $outdir "
-echo " Exp ID: $expid "
+echo " InExp ID: $inexpid "
+echo " OutExp ID: $outexpid "
 echo " Start date: $startdate "
 echo " End date: $enddate "
-
+echo " Perturb: $perturb "
+# Miguel: Style is not supported anymore in this preprocessing scripts, I don't
+# understand why it's being removed. @JAN, @JOAKIM?
```
JanStreffing: I think this is because Joakim took back his own long run changes in the source code and decided to use the esm_tools based ones now. Maybe @joakimkjellsson or @seb-wahl can comment?

configs/components/oifs/oifs.yaml
```diff
@@ -1018,20 +1036,81 @@ choose_output:
...
+                        ${oifs.input_expid}
+                        ${initial_date!syear!smonth!sday}
+                        ${start_date!syear!smonth!sday}
+                        ${oifs.wam_number}
+                        ${oifs.perturb}
+                        ${oifs.nx}
+                        ${oifs.ensemble_id}"
+        2:
+                pseudo_initial_date: "${initial_date}"
+                #TODO: prev_pseudo_initial_date: "${initial_date}"
```
JanStreffing: Do we still need this or can this be removed?

configs/components/oifs/oifs.yaml
```diff
@@ -1018,20 +1036,81 @@ choose_output:
...
+                start_ndays_source: "${start_ndays}"
+                preprocess_method: "${general.esm_function_dir}/components/oifs/skip.sh"
+        "*":
+                pseudo_initial_date: "${last_start_date}"
+                #TODO: prev_pseudo_initial_date: "${prev_run.oifs.pseudo_initial_date}"
```
JanStreffing: Same question as above

JanStreffing:
The code changes are quite extensive and it's been a while since I worked on this. I think it's easy to check that what is here works. I guess you have done that extensively. The questions whether all that is here is needed to make it happen is really down to you Miguel, since you are deep in this. Hopefully we will never have to touch this again. ;-)

configs/components/oifs/oifs.yaml
mandresm: I'd like to keep it there. That variable might be needed it in case the legs have different lengths.

JanStreffing: Ah okay, would it be hard to get that in before the merge? I think this is a very common use case. E.g. all of Dima Sidorenkos runs recently have been with changing the length of the run in the middle. I'd like to avoid having to come back to this again as much as possible. It will take the next person a long time to figure out what's going on here.

mandresm:
Very early, I realized I could not work with what was the current state of the eternal runs and restart files. It was too complicated and the information was spread through the whole file. That's why I decided to restructure a lot.

However, most of the changes are mainly changes in the variables names that end up in the `restart` file dictionaries only. There is a lot of moving lines around, so that related lines are together and it is easier to read.

I hope this helps readability not only for me but for the next person that reads `oifs.yaml` (fingers crossed)

configs/components/oifs/oifs.yaml
mandresm: Basically to do a restart with a different leg length? I'll try that then.

JanStreffing: Yes. I think most users will change the length between leg 1 and 2, but it would be nice to also test the cases of changing between leg 2 and 3, as well at between 3 and 4.

JanStreffing:
Midnight commits (⊙＿⊙')
Looks like only the Jesus Style question is still open. I think it can be answered when we merge with GEOMAR. From my side this is cleared.

JanStreffing (approved):

configs/components/oifs/change_icm_date.sh
joakimkjellsson: Correct! "style" was used becuase Jan and I had different ways of coping with the long runs. 
But now that we are both on the same page (i.e. we are both "jesus") I think we can remove the "style" variable. 
Cheers
Joakim

joakimkjellsson (approved):

PR #381, opened 2021-07-05
fixed uninstall command

JanStreffing:


mandresm (approved):

docs/installation.rst
```diff
@@ -54,8 +54,8 @@ $ esm_versions clean
 
 You can also choose to manually uninstall. In order to do that, remove the installed Python packages and delete the ``esm_*`` executables. The following commands will do the trick if you installed with the ``install.sh`` script or installed using ``pip`` with ``user`` mode ::
 
-$ rm ~/.local/bin/esm*
-$ rm ~/.local/lib/python3.6/site-packages/esm*
+$ rm -rf ~/.local/bin/esm*
+$ rm -rf ~/.local/lib/python3.6/site-packages/esm*
```
denizural, line 58: I would rather do `rm -ir` here since `-rf` forcefully removes everything. What if you have some important things starting with `esm*`. This is why I changed the related part in the `esm_versions` because it deleted an unrelated file 😭

PR #397, opened 2021-07-15
fix(install.sh): allows the install to work inside of venv

pgierz:
If you are using a virtual environment, `./install.sh` breaks. This checks and fixes that.


mandresm (approved):

pgierz:
Miguel, when you have a moment, can you put this in as a hot fix to release, prep release, and develop?

Issue #390, opened 2021-07-12
fesom restart only does not work in echam+fesom coupled setup


chrisdane:
**Describe the bug**
In a coupled echam+fesom+recom experiment, I need to restart fesom (and recom) only from a coupled echam+fesom+recom experiment. In the [branching off docu](https://esm-tools.readthedocs.io/en/latest/recipes/how_to_run.html?highlight=restart#how-to-setup-runscripts-for-different-kind-of-experiments) its written that
```yaml
general:
    lresume: 1
    ini_parent_exp_id: <old-experiment-id>
    ini_restart_dir: <path-to-restart-dir-of-old-experiment>/restart/
```
needs to be in the runscript.

With 
```yaml
general:
    lresume: True
    ini_parent_exp_id: test
    ini_restart_dir: /work/ba1103/a270094/AWIESM/test/restart/
echam:
    lresume: False
jsbach:
    lresume: False
hdmodel:
    lresume: False
fesom:
    lresume: "${general.lresume}"
recom:
    lresume: "${general.lresume}"
oasis3mct:
    lresume: False
```
esm_runscripts complains that
```
==============================================================================
::: Executing the step:  set_parent_info    (step [9/15] of the job:  prepare)
==============================================================================
  File "/mnt/lustre01/pf/a/a270073/esm/src/esm_runscripts/esm_runscripts/prepare.py", line 632, in set_parent_info
    config[model]["parent_date"] = config[model]["ini_parent_date"]
KeyError: 'ini_parent_date'
```
When I add
```yaml
general:
    ini_parent_date: 2685
```
the error changes to
```
  File "/mnt/lustre01/pf/a/a270073/esm/src/esm_runscripts/esm_runscripts/prepare.py", line 633, in set_parent_info
    config[model]["parent_restart_dir"] = config[model]["ini_parent_dir"]
KeyError: 'ini_parent_dir'
```
When I add
```yaml
general:
    ini_parent_dir: /work/ba1103/a270094/AWIESM/test/
```
the error changes to
```
  File "/pf/a/a270073/.local/lib/python3.6/site-packages/esm_calendar/esm_calendar.py", line 306, in __init__
    int, ndate
ValueError: invalid literal for int() with base 10: ''
```
When I change `ini_parent_date` to
```yaml
general:
    ini_parent_date: "2685-12-31"
```
the error changes to 
```
[...]

PR #398, opened 2021-07-15
hotffix/fesom restart date

mandresm:
This is the fix we talk about it yesterday in Slack, that should also solve #390. I'm not so convinced anymore that is a good fix, therefore I'd like to get some feedback. The disadvantages of this fix in contrast with Paul's 3b7c824bab65fbc5f5a377dd2bbaeaea805555dc:

1. Too complex, in comparison to the standard naming convention (just using the year) and letting ESM-Tools to time stamp repetitions in its own way (`fesom.YYYY.restart.oce.nc_YYYYMMDD-YYYYMMDD`).
2. The user needs to take care of one extra step if using an old simulation.
3. This only happens for FESOM-1 and RECOM, FESOM-2 still copies with the normal names.

The only advantage I see from keeping the naming convention in the general restart directory as it currently is, is that we won't generate a problem for people already using that. I don't think there are many though and, now that I am more experienced, I really don't like what I did with the fesom and recom restarts renaming... At this point I prefer Paul's solution.

### DIFF
--- a/configs/components/fesom/fesom.yaml
+++ b/configs/components/fesom/fesom.yaml
@@ -128,9 +128,9 @@ restart_in_in_work:
         oce_restart: fesom.${parent_date!syear}.oce.nc
         ice_restart: fesom.${parent_date!syear}.ice.nc
 restart_in_sources:
-        oce_diag: fesom.${parent_date!syear!smonth!sday}.oce.diag.nc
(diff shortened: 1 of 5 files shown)


mandresm:
Fixes by @pgierz and @chrisdane incorporated. I will test fesom-recom-1.4 and merge it if it works.

mandresm:
AWICM compiles and runs with restarts.

PR #408, opened 2021-07-27
removed unneeded temporary namelist.echam files

chrisdane:
ps: `flake8 esm_tools` gives a lot of errors like
```
esm_tools/setup.py:5:1: F401 'setuptools.find_packages' imported but unused
esm_tools/setup.py:19:80: E501 line too long (85 > 79 characters)
```


pgierz (approved):

pgierz:
Flake8 is badly configured for our project, and we don't really have a set code style. We should probably at some point start using one, and I think I always try to put my stuff through one of Python's auto-code formatters (black, in particular). Between release 6 and 7 we wanted to focus more on internal cleanup.

@mandresm can you do the whole tag/patch into develop/etc stuff? Maybe we can also add that whole process to the contributing guide so Chris can do it himself next time.

pgierz:
.....wait, Chris, you are allowed to merge?

chrisdane:
öh ups sorry i dont know o_O

mandresm:
I think he is allowed to merge because you approved the PR and he has `write` permisions. Otherwise he would need admin permissions. But anyway, I agree, merges to `release` and `develop` should be banned from anyone that is not admin.

pgierz:
I'll have a look at the settings, that should be possible to configure

mandresm:
Bumversion is done, as well as merging into `develop` and `prep_release`

PR #411, opened 2021-07-28
Streams from ECHAM Namelist

pgierz:
Both are needed for runscripts PR to work


pgierz:
AWIESM gets its own recipe, and write config happens almost last. If any users start to get problems with AWIESM I will look at it.

PR #421, opened 2021-08-04
Add simple installation/run tests with GitHub actions

koldunovn:
Two versions of the tests are added. Description and discussion in #422

### DIFF
--- a/.github/workflows/esm_tools_actions.yml
+++ b/.github/workflows/esm_tools_actions.yml
@@ -13,6 +13,7 @@ jobs:
       LC_ALL: en_US.UTF-8
       LANG: en_US.UTF-8
     container: koldunovn/esm_test:latest
+      options: --hostname ollie1
 
     # Service containers to run with `gfortran_ubuntu`
     steps:
--- a/.github/workflows/esm_tools_actions.yml
+++ b/.github/workflows/esm_tools_actions.yml
@@ -12,7 +12,8 @@ jobs:
     env:
       LC_ALL: en_US.UTF-8
       LANG: en_US.UTF-8
-    container: koldunovn/esm_test:latest
+    container: 
+      image: koldunovn/esm_test:latest
       options: --hostname ollie1
 
     # Service containers to run with `gfortran_ubuntu`
--- a/.github/workflows/esm_tools_actions.yml
+++ b/.github/workflows/esm_tools_actions.yml
@@ -40,10 +40,6 @@ jobs:
     - name: run esm_plugins
       run: |    
         esm_plugins
-    
-    - name: run esm_plugins
-      run: |    
-        esm_plugins
         
     - name: run esm_versions
       run: |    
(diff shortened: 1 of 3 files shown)


.github/workflows/esm_tools_actions.yml
```diff
@@ -0,0 +1,63 @@
...
+    
+    - name: run esm_master second time
+      run: |    
+        esm_master
+    
+    - name: run esm_plugins
+      run: |    
+        esm_plugins
+    
+    - name: run esm_plugins
```
pgierz: We don't need that one twice

.github/workflows/esm_tools_actions_novm.yml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  basic_tests_no_docker:
+    runs-on: ubuntu-latest
+    env:
+          LC_ALL: en_US.UTF-8
+          LANG: en_US.UTF-8
+    strategy:
+      matrix:
+        python-version: [ '3.6', '3.7', '3.8', '3.9']
```
pgierz, line 16: I think we still have at least one machine using Python 3.5. But we also use f-strings, which aren't in there until 3.6....so, that will fly around the ears at some point.

pgierz (commented):
Hi Nikolay, thanks very very VERY much for putting this together. I will be a great starting place.

.github/workflows/esm_tools_actions.yml
koldunovn: Oh, fixed.

.github/workflows/esm_tools_actions_novm.yml
koldunovn: I added 3.5, and it failed with `SyntaxError: invalid syntax` (https://github.com/esm-tools/esm_tools/runs/3242808547?check_suite_focus=true). I will remove now. It was actually a good test showing it's working :)

pgierz: Yup, we get a big red X. I'll adapt the setup.py to ensure it doesn't even try to install itself if you aren't using 3.6 or newer. 

I at some point for one my private projects also had figured out how to get the CI to automatically publish stuff to PyPI. This might also be something to look into

pgierz:
....hooray we are slowly moving into the modern world 🥳

.github/workflows/esm_tools_actions_novm.yml
koldunovn: @pgierz looks like there is an action for this :) https://github.com/pypa/gh-action-pypi-publish One could probably do it for tags or stable releases, though :)

mandresm:
Thanks a lot Nikolay! :) This is awesome!

mandresm (approved):
I think we can merge this into `develop` and `prep_release` already, so that we have this very basic CI already in place, and we can expand it soon through other PRs.

PR #394, opened 2021-07-13
Hotfix/fesom yearnew branchoff

mandresm:
Fixes for restarting fesom as a branchoff experiment. When branching off `yearnew` in the `namelist.config` cannot be equal to the last number in the `fesom.clock`, otherwise it ignores the fesom restart files and does a spinup.

PR #426, opened 2021-08-11
hotfix/fesom newday

mandresm:
Allows FESOM to do a cold start in the middle of a year.

`daynew` in `fesom*.yaml`, only needs to be defined if the run is a cold start. In that case both `daynew` and `yearnew` should coincide with the clock.

In a restart run (`lresume: true` and `run_number > 1`), the date defined by `daynew` and `yearnew` should be prior to the one defined by the `fesom.clock`. In the yamls it is configured that `daynew` and `yearnew` remain the same, as `initial_date` represents the date in which the whole experiment begins.

In a branchoff experiment (`lresume: true` and `run_number: 1`) `initial_date` matches the `fesom.clock` created by the configuration file which has the potential to run a cold start. To avoid that `yearnew` is defined to be the previous year to the current simulation so that restart occurs.

This follows up on a previous fix #394 for branchoff experiments.

### DIFF
--- a/configs/components/fesom/fesom-2.0.yaml
+++ b/configs/components/fesom/fesom-2.0.yaml
@@ -163,9 +163,13 @@ log_sources:
 # Is it a branchoff experiment?
 branchoff: "$(( ${lresume} and ${general.run_number}==1 ))"
 choose_branchoff:
+        # Makes sure the <units>new in the namelist make an earlier date than that of
+        # of the clock to avoid cold starts for branching off experiments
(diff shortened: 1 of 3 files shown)


koldunovn (approved):
Looks good to me.

pgierz (approved):
I both love and hate that we some invented a domain specific language on top of python and yaml. I am sure there are dumber ways of doing this, but there must also be better ways...

Looks good. 👍

PR #429, opened 2021-08-12
Hotfix/fesom newday

mandresm:
Same as #426

### DIFF
--- a/configs/components/fesom/fesom-2.0.yaml
+++ b/configs/components/fesom/fesom-2.0.yaml
@@ -163,9 +163,13 @@ log_sources:
 # Is it a branchoff experiment?
 branchoff: "$(( ${lresume} and ${general.run_number}==1 ))"
 choose_branchoff:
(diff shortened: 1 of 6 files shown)


pgierz:
We had failing tests on the version bump, is that a problem? (Note the big red X)

mandresm:
Yeah, the problems in `prep_release` is that `esm_runscripts` is not in `prep_release` while `esm_tools` is. We need to come up with an idea on how to coordinate different versions packages with different branches for the testing.

pgierz:
....just one repo? Instead of 5? Just an idea 😉

mandresm:
Yeah, I'm not opposed to that idea, I just don't see it happening before release 6 cause there is so much left to be done. In any case, why were the packages separated on the first place? (I was not there when it happened)

pgierz:
That is a question Dirk can answer better than I can. We at the beginning wanted to (I guess) make very separate tools that could be used without depending on each other, but that is (at least as it is right now) no longer the case.

PR #435, opened 2021-08-24
Feat/better ci

pgierz:
This PR is a start for some CI which deals primarily with style and syntax fixes.

The following are currently automatically checked:
* Does your code pass flake8? (that is for syntax)
* Does your code pass black? (that is for style)

The Flake8 check will catch all sorts of mini annoyances, so, in any case that is good to have. Black is more of a "Paul Favorite" and will help us all to keep our style consistent across the various developers.

In case of a `black` failure, the CI system will automatically (or, magically) reformat your code for you, make a new commit, and push. This is admittedly petty, I know. However, if we all agree on how to format things, merges will in the long run be easier (at least, in my opinion). 

I would from my side encourage more automatic things: we can even include something like releasing to PyPI on every tag, so perhaps something we can discuss in one of the next team meetings or at the mini-retreat.

### DIFF
--- a/.github/workflows/flake8_checks.yml
+++ b/.github/workflows/flake8_checks.yml
@@ -1,10 +1,5 @@
 name: Flake8
 
-on:
-  push:
-    paths:
-      - '*.py'
-
 jobs:
   flake8_py3:
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


pgierz:
...and, I am sorry for the horrible commit messages. It took me a bit to get this off the ground :-(

pgierz:
Superceeded by #462

PR #441, opened 2021-08-27
HIST scenario is working at least for T63

mandresm:
Fix for historical runs with ECHAM


denizural (approved):
I am not sure but for some simulations 2 years before / after is also needed. I made those changes too. You can see them here:
https://github.com/esm-tools/esm_tools/blob/hist_fixed_and_awicm_recom/configs/components/echam/echam.yaml
The problem is I made check runs and short runs and hand them over to the colleagues but follow-up issues come later on. Maybe meeting with Chris can solve some of these issues.

mandresm:
Merged already in #496

PR #448, opened 2021-09-15
Fesom 2.1 for awiesm 2.1 recom

mandresm:
Fesom 2.1 for awiesm 2.1 recom and wiso generalization


configs/setups/awiesm/awiesm-2.1.yaml
```diff
@@ -75,6 +83,11 @@ echam:
         namelist_changes:
                 namelist.echam:
                         runctl:
+                                # NOTE(PG): Almost no one in paleodyn uses the
```
pgierz: Hello,

it turns out I made a mistake here. `default output: true` is a horrible idea, and mixes up the output between snapshots and monthly means.  Please remove that chunk

pgierz (changes requested):
Hi Miguel,

I only checked if the AWI-ESM configs needed any changes, and have one small comment. @mbutzin will know better what exactly we can give the green light to for RECOM, and either @xshi-awi or @mwerner-awi can have a look at the WISO parts.

configs/setups/awiesm/awiesm-2.1.yaml
```diff
@@ -75,6 +83,11 @@ echam:
...
+                                # stream-style output, and most people expect a
```
pgierz: And in fact, even the comment is wrong. We all use the stream-style output, just that the paleodyn workflow expects different streams than are defined by default in echam. I have begun an overall of the namelists for our department, if you are interested, please see here:

https://gitlab.awi.de/paleodyn/Models/namelists

mandresm:
> Hi Miguel,
> 
> I only checked if the AWI-ESM configs needed any changes, and have one small comment. @mbutzin will know better what exactly we can give the green light to for RECOM, and either @xshi-awi or @mwerner-awi can have a look at the WISO parts.

Thanks for the review, just a note, this is a merge for the AWIESM-RECOM group, for the more general merge into `prep_release` I'll get all the interested parts involved :) but thanks for your comments, they will prevent headaches from the very beginning

configs/setups/awiesm/awiesm-2.1.yaml
mandresm: That's great, could this repo be syncronised or included within echam namelists inside `esm_tools` GitHub? Then things could be completely automatic.

pgierz (approved):
👍

configs/setups/awiesm/awiesm-2.1.yaml
pgierz: We can discuss that...we could include it as a submodule for example, but git submodules are one level above the git knowledge of some of our users...

PR #454, opened 2021-09-24
monorepo motd

mandresm:
changes on the motd for the monorepo rework


mandresm:
Can we merge this one now?

PR #462, opened 2021-09-28
Monorepo style checks

pgierz:
Automatically check, and if needed, fix Python code style to conform with black: https://black.readthedocs.io/en/stable/

The second commit, made automatically by the CI system, trigger the chain again, but since it just did black, it will not need to do it again, thus we do not end in an infinite look


mandresm:
Can we avoid merging this right now? There are still some pull requests needed that they will be hapenning previous to Release 6 and that means that if we reformat all the code before doing a code freeze, merges can become difficult. What do you think about that point @pgierz ?

pgierz:
The automatic black is overkill, so I will overly kill this PR. If we really want to enforce this, we can build a check into CI, but having it done automatically is too much I think.

mandresm:
I liked the idea, but the problem I see is that I've been using a lot git diff lately, and running black would have make things difficult for using git diff. I think this is something to be implemented after we do the cleanup, one package at a time.

denizural:
I also think that code formatting should rather be integrated into the editor/IDE rather than the workflow. The workflow should not modify the code but accept it as it is written and test it.

PR #465, opened 2021-09-28
Monorepo hist fixed

pgierz:
@mandresm and/or @denizural can you (once you have time) double check if these are all changes needed for hist_fixed?

Thanks.


mandresm:
I have never looked at this, I think it was @denizural that did most of the work.

mandresm:
Merged already in #496

PR #478, opened 2021-10-15
Hotfix/sbatch mail2general and mistral c

pgierz:


pgierz:
OK, the checks that are failing have nothing to do with this particular change.

No chunky parts? That sounds like @dbarbi

mandresm:
I know what's the problem here. The `choose_` for the partitions is different in this version. I'll have  a look at this soon after I merge in release.

mandresm:
Just remembered, that the testing does not work on `prep_release` branch because the installation of `esm_tools` installs the packages in `release`. And the recipes (contianed in `esm_tools`) call to elements of `esm_runscripts` that do not exist in `release`. So I think you can go ahead and merge this. I would only care about the tests in the monorepo.

mandresm (approved):

PR #480, opened 2021-10-19
Prep release

mandresm:
Fixed computer files other than `ollie` for `prep_release`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(
     test_suite="tests",
     tests_require=test_requirements,
     url="https://github.com/esm_tools/esm_tools",
-    version="5.1.23",
+    version="5.1.24",
     zip_safe=False,
 )
 
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(
     author_email="dirk.barbi@awi.de",
     python_requires=">=3.6",
     classifiers=[
-        "Development Status :: 3 - Beta",
+        "Development Status :: 3 - Alpha",
         "Intended Audience :: Science/Research",
         "License :: OSI Approved :: GNU General Public License v2 (GPLv2)",
         "Natural Language :: English",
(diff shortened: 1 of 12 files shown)


setup.py
```diff
@@ -31,7 +31,7 @@
     author_email="dirk.barbi@awi.de",
     python_requires=">=3.6",
     classifiers=[
-        "Development Status :: 3 - Beta",
+        "Development Status :: 3 - Alpha",
```
pgierz, line 34: This shouldn't matter I think...?

setup.py
```diff
@@ -46,7 +46,7 @@
     long_description=readme + "\n\n" + history,
     include_package_data=True,
     keywords="esm_tools",
-    name="esm_tools",
+    name="esm-tools",
```
pgierz, line 49: Out of curiosity, what is wrong with underscore vs minus? Might be a nice thing to have in my notes for the future...

pgierz (approved):

setup.py
mandresm: It seems to matter, it won't allow me to upload as Status 3 classifier only exist in PyPI for Alpha, Beta is Status 4

pgierz:
Ah sorry, I only looked at about half of this: the diff I looked at only showed your last commit. The other ones seem to just be things for the version number (small peanuts) and the partition stuff. I guess that was part of the work needed for @chrisdane? Or no?

setup.py
mandresm: That is the name the package is going to have in PyPI and the name of the project.

pgierz: So, then, that means we do `pip install esm-tools` instead of `pip install esm_tools`? Only important insofar as we write the correct hints into the readme. Thanks for testing it all out!!

mandresm:
Correct, this fixes all computers for `prep_release`

setup.py
pgierz: OK, I found a list of all the things we are allowed to use: https://pypi.org/classifiers/

Either we do
```
Development Status :: 4 - Beta
```
or
```
Development Status :: 3 - Alpha
```

I would argue we got past the Alpha stage though ;-)

setup.py
mandresm: It seems to cover both, `pip install esm_tools` and `pip install esm-tools` for reasons I don't fully understand XD

pgierz:
RE the failing test: 

> AttributeError: module 'esm_runscripts' has no attribute 'chunky_parts'

That is very likely a version mismatch. I would at this stage suggest we ignore any failing test until after everything is merged, but once we all have one single version to work from we should ensure that they all go through

mandresm:
This is `prep_release` installing `esm_runscripts` in `release` branch and trying to use the method `chunky_parts` that does not exist in `release` but in `prep_release`, so yes, we should ignore all this failing messages until we have one single repo where all the versions are consistent.

PR #482, opened 2021-10-19
Mistral openmpi

mandresm:
Working `openMPI` environment for `AWICM3` on `mistral` (environment taken from the @hegish ksh environment).

If `useMPI` is not specified in the `runscript` or through the `esm_master` using the `-m` flag, everything remains as before.


JanStreffing (approved):
I figure this is tested, so I have no objections.

JanStreffing:
> If `useMPI` is not specified in the `runscript` or through the `esm_master` using the `-m` flag, everything remains as before.

Btw: Is is possible to set a default esm_master env for a specific combination of machine and model?

mandresm:
Yes, simply adding the following to the model yaml:
```
computer:
    choose_computer.name:
        mistral:
            useMPI: openmpi
```

JanStreffing:
Ok good to know. I guess it more applies to HLRN where the MPI is not a free choice yet. For this PR on mistral everything looks good.

PR #492, opened 2021-11-03
Monorepo prep release update

mandresm:
Merges all up-to-date `prep_release` branches of all the packages (except `esm_tests`, that one still needs some work before merging).

I made sure all fixes that @pgierz did for the packages in the `monorepo` branch to work properly are preserved.

Tests were run for fesom2 and awicm2.

This merge can delete `prep_release` branch in `esm_tools`, as it has been merged to this branch.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 graft configs
 graft namelists
 graft runscripts
+graft stuff
--- a/configs/machines/blogin.yaml
+++ b/configs/machines/blogin.yaml
@@ -21,18 +21,22 @@ batch_system: "slurm"
 jobtype: compute
 sh_interpreter: "/bin/bash"
 
-choose_jobtype:
-        tidy_and_resubmit:
-                partition: standard96
-        post:
-                partition: standard96
-        compute:
-                partition: standard96
 
+partition: standard96
 
 choose_partition:
         standard96:
-                cores_per_node: 96
+                partition_name: standard96
+                partition_cpn: 96
+
+partitions:
+        compute:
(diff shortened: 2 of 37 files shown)


pgierz:
For the part: tests were run for "fesom2 and awicm2." I am still running into hiccups for fully automatic testing of *everything*. But, might it be a good idea rather than doing everything at once, slowly start to integrate what we know should work? It would be in that case be more of a build over time rather than all at once, but maybe that is easier.

if you agree, make a new pull request with a mini-run. We can sit together if you want as well, as I think there might be some tweaking needed in esm-runscripts to get that to work nicely

pgierz:
All tests passed, merging. Nice work Miguel! :-)

mandresm:
> For the part: tests were run for "fesom2 and awicm2." I am still running into hiccups for fully automatic testing of _everything_. But, might it be a good idea rather than doing everything at once, slowly start to integrate what we know should work? It would be in that case be more of a build over time rather than all at once, but maybe that is easier.

I agree, in fact in `prep_release` branch I only tested FESOM2, FESOM-1.4-RECOM and AWICM2. I am doing tests also in parallel for the `prep_release` branches that contain specific features, such as `prep_release_awicm3` and `awiesm2-wiso-recom_prep_release` so I say we implement their tests as they get merged. I am currently running all the tests in `esm_tests` and trigger them manually as I already have a lot of information about many different compilations and runs there. We could even still use that for all the merges, and once everything is merged and passing all the manual tests with `esm_tests`, then we can work our way through the CI of everything.

PR #494, opened 2021-11-08
renaming of the .sad file to .run

mandresm:
Changes the `.sad` extension by `.run`.


seb-wahl (changes requested):
You should remove the .batch_system.py.swp (vim swap file) from the commit

mandresm:
Done, and thanks!

denizural (approved):
sad file is gone, now I am very sad 😭 . Cold jokes aside, it looks fine. I assume that you grepped all sads in the code and the docs. Approve from my side. 👏 👏

denizural:
Btw, if you use neovim, it will create the `.swp` file in another directory and won't mess up your repo.

src/esm_runscripts/backup/sadfile.py
```diff
@@ -1,6 +1,6 @@
-class sadfile:
+class runfile:
```
pgierz, line 1: Not for now, but we should stick to common standards for class names. Normally they are capitalised. One more for the cleanup list.

pgierz (approved):

src/esm_runscripts/backup/sadfile.py
denizural: Ah, I missed that. Here is a good read from PEP
https://www.python.org/dev/peps/pep-0008/#class-names

CamelCase is the naming convention for the classes. I remember that some classes in `esm_master` also violate that. Not too bad but the style police might be angry.

PR #514, opened 2021-12-02
fix/update bins

### LINKED ISSUE #513: model bin not updated on esm_runscripts -U -update
**Describe the bug**
Continuing a run with `esm_runscripts runscript -e expname -U` updates the esm-* files but not the - potentially changed - model binaries.

**To Reproduce**
Stop a run. Recompile the model. Continue the run with `esm_runscripts runscript -e expname -U -c`. The experiment will continue with the old model binaries.

**Expected behavior**
The new model binaries should be used when submitted with `-U`. 

Alternatively I would like to use another argument to explicitly update the model bins, e.g. `-update_bin`, as I also see good reasons to not update the model bins on `-U`.  

 - Supercomputer: mistral
<details>
<summary>esm_versions check</summary>

```
+---------------------+-----------+---------------------------------------------------------------+----------------------+----------------------------+
| package_name        | version   | file                                                          | branch               | tags                       |
|---------------------+-----------+---------------------------------------------------------------+----------------------+----------------------------|
| esm_analysis        | 0.3.0     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_archiving       | 5.0.0     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_calendar        | 5.0.0     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_database        | 5.0.0     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_environment     | 5.1.3     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_master          | 5.1.6     | /mnt/lustre01/pf/a/a270073/esm/src/esm_master                 | develop              | v5.1.6-10-g96c2b66         |
| esm_motd            | 5.0.2     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_parser          | 5.1.12    | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_plugin_manager  | 5.0.1     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_profile         | 5.0.0     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_rcfile          | 5.1.0     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
| esm_runscripts      | 5.1.35    | /mnt/lustre01/pf/a/a270073/esm/src/esm_runscripts             | hist_fixed_issue_485 | v5.1.35-4-g5154c21         |
| esm_tools           | 5.1.0     | /mnt/lustre01/pf/a/a270073/esm/esm_tools                      | ssp585_awicm-recom   | v5.0.17-89-g2ad1cfd4-dirty |
| esm_version_checker | 5.1.5     | /mnt/lustre01/pf/a/a270073/.local/lib/python3.6/site-packages |                      |                            |
+---------------------+-----------+---------------------------------------------------------------+----------------------+----------------------------+
```
</details>

mandresm:
Closes #513

### DIFF
--- a/src/esm_runscripts/helpers.py
+++ b/src/esm_runscripts/helpers.py
@@ -208,7 +208,7 @@ def update_reusable_filetypes(config, reusable_filetypes=None):
             reusable_filetypes = list(set(reusable_filetypes))
             # Do the removal
             if config["general"].get("verbose", False):
-                print(f"Removing f{update_filetype}")
+                print(f"Removing {update_filetype}")
             reusable_filetypes.remove(update_filetype)
         elif config["general"].get("verbose", False):
             print(
--- a/src/esm_runscripts/helpers.py
+++ b/src/esm_runscripts/helpers.py
@@ -163,11 +163,11 @@ def update_reusable_filetypes(config, reusable_filetypes=None):
     Returns
     -------
     dict or list
-        If the user gives no ``reusable_filetypes`` as input. A dict 
-        ``config`` is returned, which contains the changes requested 
+        If the user gives no ``reusable_filetypes`` as input. A dict
+        ``config`` is returned, which contains the changes requested
         by the user with the ``--update-flag``.
-        
-        If the user gives ``reusable_filetypes`` as input, a list is 
+
+        If the user gives ``reusable_filetypes`` as input, a list is
         returned. This list ``reusable_filetypes`` contains the changes
         requested by the user with the ``--update-flag``.
     """
--- a/src/esm_runscripts/helpers.py
+++ b/src/esm_runscripts/helpers.py
@@ -162,12 +162,14 @@ def update_reusable_filetypes(config, reusable_filetypes=None):
 
     Returns
     -------
-    config : dict
-        If the user gives no ``reusable_filetypes`` as input. ``config`` contains the
-        changes requested by the user with the ``--update-flag``.
-    reusable_filetypes : list
-        If the user gives ``reusable_filetypes`` as input. ``reusable_filetypes``
-        contains the changes requested by the user with the ``--update-flag``.
+    dict or list
+        If the user gives no ``reusable_filetypes`` as input. A dict 
+        ``config`` is returned, which contains the changes requested 
+        by the user with the ``--update-flag``.
+        
+        If the user gives ``reusable_filetypes`` as input, a list is 
+        returned. This list ``reusable_filetypes`` contains the changes
+        requested by the user with the ``--update-flag``.
     """
     # Set the logic for what to return
     if reusable_filetypes:
--- a/src/esm_runscripts/helpers.py
+++ b/src/esm_runscripts/helpers.py
@@ -151,7 +151,7 @@ def update_reusable_filetypes(config, reusable_filetypes=None):
 
     This method takes care of removing the reusable file types from
     ``reusable_filetypes``, specified by the user in the ``esm_runscripts``
-    call with the ``--update-files`` flag.
+    call with the ``--update-filetypes`` flag.
 
     Parameters
     ----------
@@ -179,18 +179,22 @@ def update_reusable_filetypes(config, reusable_filetypes=None):
     update_filetypes = (
         config["general"].get("command_line_config", {}).get("update_filetypes", [])
     )
+    if config["general"].get("verbose", False) and update_filetypes:
+        print("User requests that the following filetypes be updated:")
+        [print(f"* {filetype}") for filetype in update_filetypes]
+
     # NOTE(MAM, PG): Originally defined in prepare.py
     # https://tinyurl.com/2p8awzsu
     potentially_reusable_filetypes = config["general"]["potentially_reusable_filetypes"]
 
-    # Loop through the file types specified by the user with the ``--update-files``
+    # Loop through the file types specified by the user with the ``--update-filetypes``
     # flag
     for update_filetype in update_filetypes:
         # Check if that file type exists/makes sense. Otherwise, through an error
         if update_filetype not in potentially_reusable_filetypes:
             esm_parser.user_error(
-                "update-files",
-                f"``{update_filetype}`` specified by you in ``--update-files`` is not a "
+                "update-filetypes",
+                f"``{update_filetype}`` specified by you in ``--update-filetypes`` is not a "
                 + "ESM-Tools file type. Please, select one (or more) of the "
                 + "following file types:\n\t- "
                 + "\n\t- ".join(potentially_reusable_filetypes),
@@ -201,13 +205,21 @@ def update_reusable_filetypes(config, reusable_filetypes=None):
             # Remove duplicates just in case
             reusable_filetypes = list(set(reusable_filetypes))
             # Do the removal
+            if config["general"].get("verbose", False):
+                print(f"Removing f{update_filetype}")
             reusable_filetypes.remove(update_filetype)
-        elif config["general"]["verbose"]:
+        elif config["general"].get("verbose", False):
             print(
                 f"- The file type ``{update_filetype}`` you are trying to update was not "
                 "reusable accross runs in the first place, so it's been always "
                 "updated with the external source, and it will still be."
             )
+
+    if config["general"].get("verbose", False):
+        print("The following filetypes will be re-used from already copied sources:")
+        for reusable_filetype in reusable_filetypes:
+            print(f"* {reusable_filetype}")
+    config["general"]["reusable_filetypes"] = reusable_filetypes
 
     if return_config:
         return config
--- a/src/esm_runscripts/helpers.py
+++ b/src/esm_runscripts/helpers.py
@@ -1,10 +1,9 @@
 import sys
 from datetime import datetime
 
-import esm_rcfile
-
 import esm_parser
 import esm_plugin_manager
+import esm_rcfile
 import esm_tools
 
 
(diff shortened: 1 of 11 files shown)


src/esm_runscripts/helpers.py
```diff
@@ -136,6 +137,78 @@ def assemble_log_message(
...
+    potentially_reusable_files = config["general"]["potentially_reusable_files"]
+
+    # If there are file types to update specified by the user, do so
+    if update_files:
+        # Loop through the file types specified by the user with the ``--update-files``
+        # flag
+        for ufile in update_files:
+            # Check if that file type exists/makes sense. Otherwise, through an error
+            if ufile not in potentially_reusable_files:
+                esm_parser.user_error("update-files",
```
pgierz: Maybe an idea for the code cleanup phase: we could move things like `user_error` into a separate module for error reporting or small utilities that we use across the various packages.

src/esm_runscripts/helpers.py
```diff
@@ -136,6 +137,78 @@ def assemble_log_message(
     return line
 
 
+def update_reusable_files(config, reusable_filetypes=None):
+    """
+    Reusable files, in general (``config["general"]["reusable_filetypes"]``), or model
+    specific (``config[<model>]["reusable_filetypes"]``) forces files of a given type,
+    such as ``bin``, ``input``, ``forcing``, etc., to be copied only once into the
+    general experiment folder, and then **reused** them through out every run. If a
```
pgierz: Isn't this supposed to be the default behavior for some of the filetypes?

src/esm_runscripts/helpers.py
```diff
@@ -136,6 +137,78 @@ def assemble_log_message(
...
+    given file type is not present in this list, that file type will be copied from
+    its external source (i.e. for binaries from the folder where the model is compiled).
```
pgierz:
```suggestion
    its external source at the beginning of every run (i.e. for binaries from the folder where the model is compiled).
```

src/esm_runscripts/helpers.py
```diff
@@ -136,6 +137,78 @@ def assemble_log_message(
...
+        List of reusable file types that can be specified by the user.
+
+    Returns
+    -------
+    config : dict
+        If the user gives no ``reusable_filetypes`` as input. ``config`` contains the
+        changes requested by the user with the ``--update-flag``.
+    reusable_filetypes : list
+        If the user gives ``reusable_filetypes`` as input. ``reusable_filetypes``
+        contains the changes requested by the user with the ``--update-flag``.
```
pgierz: Only a minor thing: check in Sphinx just to be sure, it might be that this is documented as returning two things. If that is the case, you could try:

```suggestion
    dict or list
        If the user gives no ``reusable_filetypes`` as input. A dict 
        ``config`` is returned, which contains the changes requested 
        by the user with the ``--update-flag``.
        
        If the user gives ``reusable_filetypes`` as input, a list is 
        returned. This list ``reusable_filetypes`` contains the changes
        requested by the user with the ``--update-flag``.
```

src/esm_runscripts/helpers.py
```diff
@@ -136,6 +137,78 @@ def assemble_log_message(
...
+    """
+    # Set the logic for what to return
+    if reusable_filetypes:
+        return_config = False
+    else:
+        return_config = True
+        reusable_filetypes = config["general"].get("reusable_filetypes", [])
+
+    update_files = config["general"].get("command_line_config", {}).get("update_files")
+    potentially_reusable_files = config["general"]["potentially_reusable_files"]
```
pgierz: I'm not sure I understand this: is there a default for `config["general"]["potentially _reusable_files"]`?

src/esm_runscripts/helpers.py
```diff
@@ -136,6 +137,78 @@ def assemble_log_message(
...
+                print(
+                    f"- The file type ``{ufile}`` you are trying to update was not "
+                    "reusable accross runs in the first place, so it's been always "
+                    "updated with the external source, and it will still be."
+                )
+
+    if return_config:
+        return config
+    else:
+        return reusable_filetypes
```
pgierz:
```suggestion
    if return_config:
        return config
    return reusable_filetypes
```

pgierz (commented):
I had a few minor change suggestions and a few questions, but otherwise this looks good. I will add a unit test for it on this branch, @mandresm maybe you can have a look at that once I commit it to make sure I am testing the correct behaviour.

src/esm_runscripts/helpers.py
mandresm: Agreed. In that moment we can even make that `user_error` uses the `logger`.

src/esm_runscripts/helpers.py
mandresm: Yes, it is the default for `src` and `bin`, but you can redefine `reusable_filetypes` from the runscript, either in `general` or in the model sections.

src/esm_runscripts/helpers.py
mandresm: Yes, it is defined here:
https://github.com/esm-tools/esm_tools/blob/590a2181ea95c867c79f5a956ad8b56169fb1d83/src/esm_runscripts/prepare.py#L497-L500

src/esm_runscripts/helpers.py
pgierz: OK, I made a note so we don't forget #515

src/esm_runscripts/helpers.py
pgierz:
```suggestion
    # NOTE(MAM, PG): Originally defined in prepare.py
    # https://tinyurl.com/2p8awzsu
    potentially_reusable_files = config["general"]["potentially_reusable_files"]
```

pgierz:
Alright, I did a bit of refactoring, and I still have one failing test. It seems that if you define the reusable files at the model level, this does not work. I'm not sure if that is ever needed, though.

src/esm_runscripts/prepare.py
```diff
@@ -494,6 +493,12 @@ def _add_all_folders(config):
     config["general"]["reusable_filetypes"] = config["general"].get(
         "reusable_filetypes", ["bin", "src"]
     )
+    # Define the files that could be reusable accross runs (external files)
+    config["general"]["potentially_reusable_files"] = (
```
pgierz: I think we should here be consistent and always use `filetypes`. From my understanding of the functionality, you aren't able to change on particular file to update, but an entire category.

pgierz (changes requested):
Just a small wording or naming idea, let me know what you think.

src/esm_runscripts/prepare.py
pgierz: OK, done, I actually changed it in the other file and forgot ti here.

denizural (approved):
👍

src/esm_runscripts/prepare.py
mandresm: Perfect, thanks!

mandresm:
> Alright, I did a bit of refactoring, and I still have one failing test. It seems that if you define the reusable files at the model level, this does not work. I'm not sure if that is ever needed, though.

Model specific handling of `reusable_filetypes` is used in `AWICM3` and `FOCI`-like setups, where `OpenIFS` considers different `reusable_filetypes` than the other models (it reuses as well `input` filetypes).

I cannot reproduce the `esm_master get-fesom-2.0` error reported by GitHub actions... I'm a bit lost here, do you have an idea what might be going on?

mandresm:
@pgierz , there is something very weird going on with the testing system. If you look at the `prepare.py` file used by the GitHub Actions (`/work/ollie/abtci/github_actions/esm_tools/esm_tools/src/esm_runscripts/prepare.py`) you'll see that it's stil defining the `potentially_reusable_files` key, instead `potentially_reusable_filetypes` one:
```
    # Define the files that could be reusable accross runs (external files)
    config["general"]["potentially_reusable_files"] = ( 
        all_filetypes + config["general"]["in_filetypes"]
    )   
```

However, if you look at the last version of the `prepare.py`, that line is correct:
https://github.com/esm-tools/esm_tools/blob/14b6315a85feb0ec8cd31d9a09bb93252362f0f2/src/esm_runscripts/prepare.py#L496-L499

Somehow it looks like GitHub actions is not cleaning up after a failed test, and the follow up test runs the old code? I'm kind of perplex...

pgierz (approved):

tests/test_esm_runscripts/test_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+               reusable_filetypes: ["input", "forcing", "extra"]
+        """
+        self.config["general"]["command_line_config"]["update_filetypes"] = ["extra"]
+        self.config["general"]["potentially_reusable_filetypes"].append("extra")
+        self.config["foo_model"] = {}
+        self.config["foo_model"]["reusable_filetypes"] = ["input", "forcing", "extra"]
+        out_config = esm_runscripts.helpers.update_reusable_filetypes(
+            self.config, self.config["foo_model"]["reusable_filetypes"]
+        )
+        self.assertNotIn("extra", out_config)
```
pgierz: Shouldn't this be:
```suggestion
        self.assertNotIn("extra", out_config["genereal"]["reusable_filetypes"])
```

pgierz (changes requested):
I found (maybe?) an incorrect test. Can someone double check?

tests/test_esm_runscripts/test_helpers.py
mandresm: I think that is correct, and that was the problem why the previous test was failing. When you give the `reusable_filetypes` as an input to the `update_reusable_filetypes` method it returns the list, not a dictionary.
https://github.com/esm-tools/esm_tools/blob/aa221c98f9a0d90511a57a1bf9abf0f04aadd76b/src/esm_runscripts/helpers.py#L174-L179
https://github.com/esm-tools/esm_tools/blob/aa221c98f9a0d90511a57a1bf9abf0f04aadd76b/src/esm_runscripts/helpers.py#L226-L228
This is because this method is used in different places in the code:
1. To change the general one (no `reusable_filetypes` provided)
2. In `filelists.py` where it is used for each model and for that case, the files that are to be reused are calculated on the fly and don't make it into the config dictionary.

pgierz: OK, I renamed the output variable of the function to hopefully make more sense. Once the tests go through, I'll merge this.

pgierz (approved):

PR #519, opened 2021-12-10
Fix/vilma pism

mandresm:
Including `vilma-pism` to the testing software. Changes are only into the `esm_tests` so this should not affect any model.


mandresm:
Wrong target branch, and therefore PR closed

Issue #533, opened 2021-12-17
Failing CI after Release 6


pgierz:
Hello,

This issue is for discussion and does not really report a "problem", but maybe something we should be aware of. The following models and components are currently failing the compile tests. In some cases, it is because my CI dummy account doesn't have access everywhere, in others, I am a bit confused:

## Full Setups
**AWIESM**
+ [x] `AWIESM 2.1`: Could not read username (but why then does it work below?)
+ [x] `AWIESM 2.1 Wiso`: As above
+ [x] `AWIESM 2.1 Recom`: As above

## Standalone Components

**ECHAM**
+ [ ] `6.3.04p1-esm_interface`: Tries to compile as standalone, but probably is not a standalone valid component
+ [ ] `6.3.04p1-recom-awicm`: As above
+ [ ] `6.3.05p2-concurrent_radiation`: Can't tell yet why
+ [x] `6.3.05p2-concurrent_radiation-paleodyn`: Can't tell yet why
+ [ ] `6.3.05p2-foci`: Permission error of git clone, and not a standalone model
+ [x] `6.3.05p2-wiso`: No Username for "https://gitlab.awi.de". That's from the same repository as the paleodyn variant, so, why is it able to download one and not the other?

**PISM**:
+ [x] `Snowflake 0.7`: No longer exists after `swrepo.awi.de` retired
+ [x] `Snowflake 1.0`: As above

**FESOM**
+ [ ] `fesom-1.4-recom-awicm`: Not a standalone model
+ [ ] `fesom-1.4-recom-mocsy-slp`: Not a standalone model
+ [ ] `fesom-2.0-jio`: FORTRAN internal error?
+ [ ] `fesom-2.0-o`: Error during get phase (permission denied?)
+ [ ] `fesom-2.0-paleodyn`: Fortran interal error (or, model not designed as standalone?)

## General ideas
One idea. The following:
```
python/2020.2.902/lib/python3.7/subprocess.py", line 512, in run
119
    output=stdout, stderr=stderr)
120
subprocess.CalledProcessError: Command '['git', 'clone', '-b', 'oifs-deck', 'https://github.com/FESOM/fesom2.git', 'fesom-2.0-o']' returned non-zero exit status 128.
121
Error: Process completed with exit code 1.
```
is super unhelpful. It just says "ok, you got back first a 128 and then a 1". I have absolutely no way of seeing what didn't work without more specific info. I guess we should make esm-master spit out something more sensible.

[...]

PR #534, opened 2021-12-20
adjusting the AWIESM recipes to the new recipe logic

mandresm:
Adjusts the AWIESM recipes to the new recipe logic.

It still needs to be tested. But it should work... (something similar works for `OpenIFS`).

### DIFF
--- a/.github/workflows/esm_tools_awicm.yml
+++ b/.github/workflows/esm_tools_awicm.yml
@@ -20,8 +20,10 @@ jobs:
     compile-ollie:
         name: Compile Test AWICM (Ollie)
         # Pattern for uses: github-org/repo/some/path/in/repo@commit_or_branch
-        # (if @commit_or_branch is omitted, it assumes the default)
-        uses: esm-tools/esm_tools/.github/workflows/esm_tools_actions_hpc_awi_ollie.yml
+        # If @commit_or_branch is ommited, github actions will break!
+        # For now, I tie it to release, but there might be better ways of doing
+        # that, see for instance here: https://tinyurl.com/4x6h373j
+        uses: "esm-tools/esm_tools/.github/workflows/esm_tools_actions_hpc_awi_ollie.yml@release"
         secrets:
           dkrz-token: ${{ secrets.GITLAB_DKRZ_TOKEN }}
           awi-token: ${{ secrets.GITLAB_AWI_TOKEN }}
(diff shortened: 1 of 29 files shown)


pgierz (commented):
Alright, the changes themselves look good. I have one question, and then a few things I need to check:

Question: Where is the `venv_bootstrap` step?

For me to look into:
It says "all checks passed", but if you looks a bit more carefully at the checks tab, it seems as if some tests did not even get triggered, see here that the circle is only halfway green:
https://github.com/esm-tools/esm_tools/pull/534/checks

mandresm:
I don't understand the outcome of those checks... I still need to get familiar with those.

The tests I run on ollie for AWIESM worked well and they also include the `hosing_files` operations (at least ESM-Tools is reporting that that step is run).

As now each simulation consists of several tasks, each task has it's own recipe. I think the old recipe in AWIESM was simply trying to add the `add_hosing_files` step, that would go into the recipe of the `prepcompute` task. `venv_bootstrap` is part of `prepexp` recipe:
https://github.com/esm-tools/esm_tools/blob/bc3cb779c34c50d244a7742f5561e7ac76421b67/configs/esm_software/esm_runscripts/esm_runscripts.yaml#L40-L47

pgierz:
Ah, alright, and since we only override specific recipes from the setup to the task flow, if we don't define one (such as `prepexp`) it just takes the default? 

For the missing checks, I can give an overview in our first monday meeting after the holidays

mandresm:
> Ah, alright, and since we only override specific recipes from the setup to the task flow, if we don't define one (such as `prepexp`) it just takes the default?

Correct!

pgierz:
OK, I found out why those other checks were not triggered. It seems that you need to specify a version. I put a notes and reference to the GitHub docs in my commit 88dae62a2f4a1e.

This may result in some of the same failures as documented in #533

pgierz (approved):
@mandresm go ahead and merge this with version bump and all that. The failing CI tests are the same as the one I reported in the issue, so, that isn't related to this PR.

How did we normally do this? Bump the version and add the tag in the PR branch, or merge, and then do that in the release branch?

mandresm:
> @mandresm go ahead and merge this with version bump and all that. The failing CI tests are the same as the one I reported in the issue, so, that isn't related to this PR.
> 
> How did we normally do this? Bump the version and add the tag in the PR branch, or merge, and then do that in the release branch?

The way I do it is the way you taught me some time ago :) : first bumpversion, and then merge into release and develop.

PR #545, opened 2022-01-17
hotfix/awicm3 hlrn

mandresm:
- catches an additional exception type reported by HLRN when using pipwithout --user
- updates some configurations needed by AWICM3 for HLRN

### DIFF
--- a/configs/components/xios/xios.yaml
+++ b/configs/components/xios/xios.yaml
@@ -36,6 +36,7 @@ choose_version:
     use_oasis: ''
   "2.5":
     git-repository: https://gitlab.dkrz.de/ec-earth/xios-2.5.git
+    archfile: ESMTOOLS_generic_oasis_intel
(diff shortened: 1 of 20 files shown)


configs/components/xios/xios.yaml
```diff
@@ -43,6 +43,8 @@ choose_version:
         archfile: ESMTOOLS_generic_oasis_intel
       parastationmpi:
         archfile: ESMTOOLS_generic_oasis_intel
+      "*":
+        archfile: ESMTOOLS_generic_oasis_intel
```
mandresm: This was needed since HLRN does not have the `useMPI` variable, and I did not want to add it. @JanStreffing , @sveta-loza , is `ESMTOOLS_generic_oasis_intel` the correct archfile to use for `xios` in HLRN?

JanStreffing: Sounds right. The default is intel compiler. Only Aleph uses something different so far with cray ftn.

JanStreffing (approved):

PR #546, opened 2022-01-17
hotfix/choose_with_int_and_floats

### LINKED ISSUE #543: esm_tools 6: Problem to start PISM "Invalid numeric value 0 for --ntasks"
**Describe the bug**
I get this error message when I want to submit a PISM job:

```
resubmit tasks: 0
Submitting jobscript to batch system...

Output written by slurm:
sbatch: error: Invalid numeric value "0" for --ntasks.
```

 - Supercomputer: mistral
 - Version: esm_tools version 6

**Additional context**
Here is a path to an experiment: `/work/ba0989/a270124/PalModII/experiments/pism_test03`

mandresm:
Resolves the problems with integers and floats (i.e. version numbers) inside choose_blocks.

For example, let's say you have this choose (in `pism.yaml`):
```
choose_computer.partitions.compute.cores_per_node:
    24: 
        nnodes: 4
        nproca: 96
        nprocb: 1
    36: 
        nnodes: 2
        nproca: 72
        nprocb: 1
```

and `computer.partitions.compute.cores_per_node` is defined to be something like this:
```
computer.partitions.compute.cores_per_node: "24"
```

This will not be resolved correctly. The opposite case won't resolve either (choice is an `int` while the possible choices are `str`).

This fix transforms the variable to be evaluated and the keys of the choose into strings, at the moment of evaluation, if they are `int`, `float` and not `bool`.

Closes #543

It needs extensive testing as this is major change in the behavior of the parser. I'll be doing that in the following days.


src/esm_parser/esm_parser.py
```diff
@@ -1468,9 +1468,19 @@ def resolve_basic_choose(config, config_to_replace_in, choose_key, blackdict={})
         choice = do_math_in_entry([False], choice, config)
     logging.debug(choice)
 
-    if choice in config_to_replace_in.get(choose_key):
+    # This allows users to use version numbers (floats) and integers as choices inside
+    # the choose_blocks, instead of having to specify the choices as strings
+    choices_available = {}
+    for ckey, cval in config_to_replace_in.get(choose_key, {}).items():
```
denizural, line 1482: I am afraid that this part can cause some issues. `items()` method works on dictionaries. What happens if `config_to_replace_in['choose_key']` is not a dictionary?

Maybe it would be a better idea to place an `isinstance` check like
`if isinstance(config_to_replace_in.get(choose_key, {}), dict)`
It looks like the previous code did not crash so far since it has 
`six.iteritems(config_to_replace_in[choose_key][choice])` on line 1473. So `config_to_replace_in[choose_key][choice]` yields a dictionary so far.

mandresm: You are right. But bare in mind that the contents of a `choose_` block need to always be a dictionary, that's how our `choose_` syntax works:

```
choose_<var>:
        <choice_1>: <var/dict/list>
        <choice_2>: <var/dict/list>
```

What do you think about placing a check here to return a user-friendly error if a `choose` is not a dictionary instead?

```
 if not isinstance(config_to_replace_in.get(choose_key, {}), dict):
    user_error("choose_ block", "``choose_`` blocks need to be defined as ``dictionaries``. And some more text that tells the user what is going on and what exact choose is giving problems")
```

denizural: Yes, this is a good solution. Good error messages prevent headaches later on 😄

mandresm: I'll do that then.

pgierz: Yes to good error messages. 👍🏻

src/esm_parser/esm_parser.py
```diff
@@ -1468,9 +1468,21 @@ def resolve_basic_choose(config, config_to_replace_in, choose_key, blackdict={})
...
+        if isinstance(ckey, (int, float)) and not isinstance(ckey, bool):
+            choices_available[str(ckey)] = cval
+        else:
+            choices_available[ckey] = cval
+
+    # Resolve the choose variables
+    if choice in choices_available:
         for update_key, update_value in six.iteritems(
```
pgierz, line 1494: Cleanup: we need to get rid of six at some point. It doesn't hurt, of course, but one more thing that is divergent everywhere in the code.

 I think I will ask Natalja is I can rent a cabin in northern Sweden for a week and just tear apart everything to do the basic cleaning

src/esm_parser/esm_parser.py
pgierz: ...but, I would still like to find a way to force strings for version numbers. It is (in my opinion) cleaner. What if you get something like:

```
echam v1.6.04p2rc-alpha2
```

I have never seen that "in the wild" of course, but some elements should just be strings, always. I would count version info to that list.

pgierz (commented):
Looks good Miguel! From my side only a small idea about version numbers and the re-emphasis that we need a cleanup phase soon...

src/esm_parser/esm_parser.py
mandresm: But that should go somewhere else right? I mean, this is specific for choices in `choose_` blocks. While what you are proposing is more of a syntax check/transformation that should go for example in the `yaml_to_dict` method, right?

src/esm_parser/esm_parser.py
mandresm: Can it be a cabin for 2 + cat?

pgierz: Well, then of course for all the esm team. But I am sure we need to first write a 5 page document outlining out covid safety concept.

src/esm_parser/esm_parser.py
mandresm: @pgierz , I think that `echam v1.6.04p2rc-alpha2` will be interpreted by the yaml parser as a `str` anyway, so it will stay being a string

mandresm:
All tests passed for all the computers (I'm not trying with Juwels), and with some exceptions that are source code related (AWIESM-RECOM and AWICM3).

There were no significant differences in the configuration files that suggested a malfunctioning of the `choose_` blocks. Therefore, I consider this PR ready to merge, I'd only need an approval :)

pgierz (approved):
One general comment or idea: it's a bit ugly, but maybe we can find a unique file extension name for the test files. If I use the github File filter thingy in the online Gui, it is difficult to focus on the actual changes to python and (maybe) config yaml. Many things are changed to keep esm-tests up to date and passing (which is good), but I cannot seem to easily filter them out (which is not so good)

mandresm:
I agree 100% with that. I am thinking of having a separate repository for those files because I don't like this at all. And sorry for the inconvenience...

PR #551, opened 2022-01-21
Hotfix/jobid in log name

mandresm:
- Brings back the jobid to the names of the general logs
- Catches errors happening during model run on PBS so that ESM-Tools does not move to the next runs if an error occurred during the simulation.


mandresm:
@pgierz , would you agree that we only need to test the runs in `ollie` and `aleph`?

mandresm:
Tested for `ollie` and `aleph`. @pgierz , if you approve I'll merge it to release on monday

pgierz:
Failing tests are the same as in #533. Approve, one less PR. A good start to Monday morning 😄

pgierz (approved):

PR #559, opened 2022-01-27
hotfix/filelist wild cards in target folder

mandresm:
In the past, if `*` was included in the source of a file, and there was no `/` in the string (the `esle` in line 339 of `filelists.py`), renaming of the file during coyping/moving/linking was permitted, as long as the target name contained a pattern equivalent to the original one:
```
source: foo.*.nc
target: bar*.something_here.nc
```

The problem was that this was not contemplated for an instance in which `/` was included (lines below 333), so that the following case was not working:
```
source: fesom_ice_restart/foo.*.nc
target: fesom_ice_restart/bar*.something_here.nc
```

This PR puts the wildcarding-renaming operations in the `get_target_name_from_wild_card` method, and uses it for both cases, fixing the problem of renaming files inside folders that contain wildcards.

**TODO**
- [x] Test for all computers
- [x] Comment and write docstrings
- [x] Correct based on comments

### DIFF
--- a/src/esm_tests/resources/last_tested/ollie/run/awicm3/awicm3-frontiers-TCO159L91-CORE2_initial/config/awicm3-frontiers-TCO159L91-CORE2_initial_finished_config.yaml
+++ b/src/esm_tests/resources/last_tested/ollie/run/awicm3/awicm3-frontiers-TCO159L91-CORE2_initial/config/awicm3-frontiers-TCO159L91-CORE2_initial_finished_config.yaml
@@ -25,7 +25,7 @@ computer:
         'srun: error:':
             file: stdout
             frequency: 30
-            message: 'SLURM ERROR: slurm endet with an error, exiting.'
+            message: 'SLURM ERROR: slurm ended with an error, exiting.'
             method: kill
     choose_heterogeneous_parallelization: *id001
     config_files:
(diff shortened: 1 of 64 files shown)


mandresm:
This is a fix requested by @mathanase and @sveta-loza aimed to make it easy to branchoff from a given FESOM simulation at different stages, copying time-stamped parallel restarts contained in `restart/fesom/fesom.<YEAR>.ice.restart` and `restart/fesom/fesom.<YEAR>.oce.restart` into the work folder, removing the time stamp.

FESOM/fesom2#279

denizural (approved):
Looks good. Another thumbs up for turning the long lines of code into a function. Such changes will make the code more readable and managable.

mandresm:
> Looks good. Another thumbs up for turning the long lines of code into a function. Such changes will make the code more readable and managable.

Thanks, it could be better written, more general, but this will be part of a next step: cleaning up (I promise :) )

src/esm_tests/resources/last_tested/ollie/run/awicm3/awicm3-frontiers-TCO159L91-CORE2_initial/config/awicm3-frontiers-TCO159L91-CORE2_initial_finished_config.yaml
```diff
@@ -25,7 +25,7 @@ computer:
         'srun: error:':
             file: stdout
             frequency: 30
-            message: 'SLURM ERROR: slurm endet with an error, exiting.'
+            message: 'SLURM ERROR: slurm ended with an error, exiting.'
```
denizural, line 28: I always wanted to change this Denglisch line 🤣

pgierz (approved):
Yes to smaller functions! I was reading some software books over the winter holiday, and one big thing for the cleanup phase we should think about is making sure functions are doing exactly only one thing (if that one thing is to wrap up a bunch of small steps, ok too)

And a huge thumbs up for doc strings!

pgierz:
I think this can go in. Can someone do the whole bump version stuff? (We really should automate the bump version stuff...)

mandresm:
I will :) sorry, this got left behind...

PR #561, opened 2022-01-28
fix/awiesm address changes

pgierz:
## Summary:
This PR allows AWIESM versions to be pulled from github.com for FESOM and
gitlab.awi.de for oasis, thus it is no longer required for users to have access
to gitlab.dkrz.de

## Commits:
- fix(awiesm): change the OASIS3-mct URL to point to AWI gitlab instead of DKRZ gitlab
- fix(oasis3mct.yaml): correct the url name
- fix(fesom.yaml): change the url address in various fesom versions to use github instead of dkrz gitlab

## Follow up questions:

We currently have this "couplings" folder, which has always a folder name, and
then the exact same yaml file name inside of it. What is the reason here? Could
we simply it? (For the cleanup phase)

### DIFF
--- a/configs/components/echam/echam.yaml
+++ b/configs/components/echam/echam.yaml
@@ -52,7 +52,7 @@ compile_infos:
             branch: "ice"
             clean_command: rm -rf src/echam/bin; rm -rf bin; make clean
             comp_command: ./config/createMakefiles.pl; autoreconf -i --force; mkdir -p src/.deps yaxt/src/.deps yaxt/tests/.deps; ./configure $configure_opts --with-fortran=intel INSTALL='/usr/bin/install -p'; make -j `nproc --all`; make install -j `nproc --all`; mkdir -p src/echam/bin; cp  bin/echam6 src/echam/bin/echam6
-            git-repository: https://gitlab.awi.de/pgierz/echam6.git
+            git-repository: https://gitlab.awi.de/paleodyn/Models/echam6.git
           6.3.05p2-concurrent_radiation-paleodyn-hammoz:
             branch: "hammoz"
             clean_command: rm -rf src/echam/bin; make clean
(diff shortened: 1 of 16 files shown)


denizural (approved):

mandresm:
I agree, we could get rid of it. I'll write it in a ticket on Jira.

configs/components/echam/echam.yaml
```diff
@@ -73,7 +73,7 @@ compile_infos:
             branch: "wiso"
             clean_command: rm -rf src/echam/bin; rm -rf bin; make clean
             comp_command: ./config/createMakefiles.pl; autoreconf -i --force; mkdir -p src/.deps yaxt/src/.deps yaxt/tests/.deps; ./configure $configure_opts --with-fortran=intel INSTALL='/usr/bin/install -p'; make -j `nproc --all`; make install -j `nproc --all`; mkdir -p src/echam/bin; cp  bin/echam6 src/echam/bin/echam6
-            git-repository: "https://gitlab.awi.de/pgierz/echam6.git"
+            git-repository: "https://gitlab.awi.de/paleodyn/Models/echam6.git"
```
mandresm, line 76: Are WISO people working on this repo already? Otherwise, we need to let them know that they should be changing repos.

configs/components/fesom/fesom-2.1.yaml
```diff
@@ -19,8 +19,6 @@ choose_version:
     branch: "2.1.0"
   2.1-wiso:
     branch: wiso
-    git-repository:
-    - https://gitlab.dkrz.de/FESOM/fesom2.git
```
mandresm, line 23: Same as my previous comment. Are WISO people working already on FESOM github?

mandresm (commented):
Thanks for making some sense out of the repos. I only have a few questions but other than that, this can go in. Let me know if you'd like me to bumpversion and merge.

mandresm:
Okay, it's merged already XD

mandresm:
@pgierz , sorry, I've just seen the email that targets my concerns. I should start from bottom to top reading emails, not the other way around...

PR #563, opened 2022-01-31
Feat/dual hemisphere

ackerlar:
I added a dual hemisphere setup, in which awiesm-pism can be run with two seperate pism instances: pism_nh (pism northern hemisphere) and pism_sh (pism southern hemisphere). The setup is like a three model setup. 
Also, I included some changes to the coupling scripts in order to neglect small icebergs, that are instead treated vie the surface runoff (as it is done in a coupling without interactive icebergs)


mandresm (approved):
All fine by me. I think the most important thing is to get to something that is working, then we can work on reducing the duplication of information and generalization. @pgierz , if you approve as well, I will merge into `feat/awiesm-pism`.

PR #566, opened 2022-02-02
fix/openmp in hlrn

mandresm:
Before, runs with heterogeneous parallelization (some executables run with MPI and OpenMP, and others only with MPI) in `HLRN` systems were not possible. This PR fixes the problem.

Successful runs were carried out with `AWICM3` where `OpenIFS` ran with 2 OpenMP threads, and therefore might be of interest to @JanStreffing , @sveta-loza , @mathanase and @helgegoessling

### DIFF
--- a/configs/machines/blogin.yaml
+++ b/configs/machines/blogin.yaml
@@ -11,9 +11,27 @@ choose_use_hyperthreading:
         True:
                 hyperthreading_flag: ""
         "0":
-                hyperthreading_flag: "--ntasks-per-core=1"
+                choose_heterogeneous_parallelization:
+                        False:
+                                hyperthreading_flag: "--ntasks-per-core=1"
+                        True:
+                                hyperthreading_flag: ""
+                                add_unset_vars:
+                                        - "SLURM_DISTRIBUTION"
+                                        - "SLURM_NTASKS"
+                                        - "SLURM_NPROCS"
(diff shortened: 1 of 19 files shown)


configs/machines/blogin.yaml
```diff
@@ -11,9 +11,27 @@ choose_use_hyperthreading:
         True:
                 hyperthreading_flag: ""
         "0":
```
seb-wahl, line 13: I would use False here for consistency. May confuse new users who don't know that we treat 0 and false the same

seb-wahl: Also in the lines above we have 1 and True which seems a duplication which seems left from times where we weren't sure if 1 and true really do the same...

seb-wahl (dismissed):
looks good, see my comments

configs/machines/blogin.yaml
mandresm: The "duplication" here was added one year ago for safety to avoid problems that users have been experiencing (I think it was something in the lines that someone wrote `use_hyperthreading: '"0"'` and then this was a string and evaluated as true. Therefore, I would not remove yet this safety feature until the background problem is resolved.

mandresm: If you are okay with that then let me know and I can go ahead and merge :)

PR #567, opened 2022-02-03
ssp awicm recom

chrisdane:
Hi

I added the CMIP6 SSP scenarios SSP126, SSP245, SSP534OS and SSP585 (official cmip6 names ssp126, ssp245, ssp534-over and ssp585). The changes are mainly added paths to forcing files for echam and jsbach.

I tested those 4 scenarios from model years 2015 to 2100 on mistral with the setup `awicm-1.0-recom`. It should also work for any echam/jsbach setup.

There are two rather ugly things.

1. Some forcing files in the pool on mistral are in a format which is not readable for echam/jsbach. For example, the ndepo files for jsbach must have either 1 (climatology) or 12 (monthly) time points per file per year. The files in the pool, however, are multi-year files. I splitted these multi-year files to annual files with 12 time points each and saved them in my local work dir and documented this here
```
/work/ba1103/a270073/forcing/JSBACH/input/r0010/T63/Ndepo/README.md
```
Thus, my local work path is in `jsbach.datasets.yaml`:
```yaml
r0010:
    add_forcing_sources:
        "Ndepo_ssp126":
            #"${input_dir}/${echam.resolution}/${echam.resolution}_ndepo.input4MIPs_surfaceFluxes_ScenarioMIP-NCAR-CCMI-ssp126.nc": # nyear x nmonth
            "/work/ba1103/a270073/forcing/JSBACH/input/r0010/${echam.resolution}/Ndepo/ssp126/${echam.resolution}_ndepo.input4MIPs_surfaceFluxes_ScenarioMIP-NCAR-CCMI-ssp126_@YEAR@.nc": # 12 months; see README.md in Ndepo/ dir
```

The population density and the historical greenhouse gases are similarly affected, I documented this here:
```
/work/ba1103/a270073/forcing/JSBACH/input/r0010/T63/spitfire/README.md
/work/ba1103/a270073/forcing/ECHAM6/input/r0007/README.md
```

2. To model the year 2100, a small change is needed in the `set_time_weight` function of `src/echam/mo_simple-plumes.f90`. The changes can be seen here: https://gitlab.dkrz.de/modular_esm/echam6/-/commit/a1118c8bba91c9a1f74ca79ecfdab3d5f6f4d000.

Cheers,
Chris


pgierz:
Hi @chrisdane, thanks! One step closer to my pipe dream of "a release version that just works".

> I tested those 4 scenarios

Which ones?

>  I splitted these multi-year files to annual files with 12 time points each and saved them in my local work dir and documented this here

How "long-lived" is that directory? I would prefer to put "new" pool files directly in the pool. There are a few options here. @christian-stepanek  maintains a paleodyn specific pool on mistral, or maybe @tsemmler05 has a way of getting that into the large AWICM pool on mistral

>  To model the year 2100, a small change is needed in the set_time_weight function of src/echam/mo_simple-plumes.f90

Can you explain a bit why that needs to be done? Also, I have tried to shift all of ECHAM code here instead: https://gitlab.awi.de/paleodyn/Models/echam6. It would be super if you could migrate the change there as well, ideally still in a branch for now. I still need to check at some point how far out of sync the version on DKRZ, my private copy on AWI Gitlab, and the paleo group one is. Once all the repos are all in one place, then I can start to tame the branches. At the end, I want to have just one single one: one repo, one branch. That will make maintaining tasks much much easier....

denizural:
👍 from my side @pgierz. Since the datasets take a lot of space, I think at some point model users need to come together and decide on the pool strategy. And this is my pipe dream 😄

pgierz:
@denizural the "lets have a clean pool" discussion has been going for almost a decade ;-)

chrisdane:
The 4 scenario names are given in the first sentence of my initial post :p

The pool files should go into the pool, right. I guess someone needs to ask someone from the dkrz pool. Who? Its a bit weird since there are already SSP runs of MPI-ESM in the CMIP6 pool, so someone must have had the same problem and either created similar files locally or changed the echam code.

I documented the necessary echam changes in the [code](https://gitlab.dkrz.de/modular_esm/echam6/-/commit/a1118c8bba91c9a1f74ca79ecfdab3d5f6f4d000):
```
    ! does not work for 31. Dec 2100:
    ! some value above `year_fr`=2100.999???? results in `iyear`=252 > `nyears`=251 and triggers STOP
...
        time_weight(1,iplume) = year_weight(iyear-1,iplume) * ann_cycle(1,iweek,iplume) ! use values from 2100
``` 
At some time of the model date 2100-12-31 echam thinks that its already the year 2101 which is not defined for this routine. I added this case for which the model should re-use the values from the year before, i.e. 2100, the values it should use anyway. Of course this is not a general solution but there is no SSP forcing beyond 2100 so it should work for this case. @tsemmler05 did a similar thing for this AWI-CM-MR-* simulations. Same thing here: Someone must have already encountered the same problem since there are SSP data for the year 2100 from MPI-ESM simulations in the CMIP6 pool. But I could not find a related change in this routine in the latest mpi-esm tarball.

pgierz:
Ugh sorry for the dumb questions. I just hate looking at ECHAM code, and your comment, while surely correct, does not explain why the behaviour was that way before and why it needs to be changed. ECHAM just does not have very clean code, it seems.

christian-stepanek:
I could make the files available here if this is wished:

```
/work/ab0246/paleo_pool
```
Just send me a link. On another note, as these files are probably most relevant for Climate Dynamics people, we could, of course, also ask Tido to put them in their pool.

Does that help for the moment?

mandresm:
Hi all, before starting resolving conflicts in this direction, I would strongly suggest to merge `release` into `ssp585_awicm-recom`. I am suggesting this because release 6 contains all the functionality packages inside a `src` folder in `esm_tools` (finally one single repo for all the ESM-Tools packages). If we don't do first this merge, we won't be able to really test AWICM-REcoM until it is merged to release, and we won't be able to perform any of our new sanity checks. 

That inverse merge should be straight forward and I can take care of it tomorrow morning if @chrisdane gives me the green light.

pgierz:
Friendly reminder to @chrisdane:
> That inverse merge should be straight forward and I can take care of it tomorrow morning if @chrisdane gives me the green light.

pgierz:
@chrisdane can you look at this again in the coming week, and rebase it vs the current release branch? I am trying to get less PRs floating around in "zombie modus". Thanks! :-)

chrisdane:
On some point Deniz made another PR with additional changes for the transition for esm_tools v6 (https://github.com/esm-tools/esm_tools/pull/665). So this PR is not needed anymore.

denizural:
Yes, this one is obsolete. We have a new one which contains the Levante support. I think the first step of clean-up should be going over the branches on the non-monorepo tools.

PR #568, opened 2022-02-04
added missing config

denizural:
This is a nasty bug. When you run `esm_runscripts` after a clean install, Python raises
```python
esm_runscripts/dataprocess.py:79: SyntaxWarning: list indices must be integers or slices, not str; perhaps you missed a comma?
  return os.path.join(["general"]["started_from"], filename)
/mnt/lustre01/pf/a/a271096/esm_tools_R6/esm_tools/src/esm_runscripts/dataprocess.py:82: SyntaxWarning: list indices must be integers or slices, not str; perhaps you missed a comma?
  return os.path.join(["general"]["started_from"], filename)
```
But in the following ones it does not. But clearly it is bug.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.1.3
+current_version = 6.1.4
 commit = True
 tag = True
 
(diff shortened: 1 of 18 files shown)


mandresm (approved):
All good from my side all `esm_tests` passed on `ollie`. Thanks for that fix!

PR #576, opened 2022-02-14
ci: allow manually selecting test jobs

pgierz:
This should allow to select CI Jobs for manual testing. It will not be active until it is in the default branch though...


mandresm (approved):
Let's give it a shot then :)

PR #580, opened 2022-02-16
ci: allow for skipping model compilation

pgierz:
Adding a tag in your commit message `[skip-ci]`, `[skip-model-comp]` will (should) not run ollie compilation. Probably it makes more sense to customize it for certain models, but this will save at least a little bit of electricity.

### DIFF
--- a/.github/workflows/esm_tools_actions_hpc_awi_ollie.yml
+++ b/.github/workflows/esm_tools_actions_hpc_awi_ollie.yml
@@ -38,6 +38,16 @@ defaults:
     shell: bash -leo pipefail {0}
 
 jobs:
+    skip_ci:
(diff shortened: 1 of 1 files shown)


mandresm (approved):
Only one question, this can only be controlled from the commit message? Does that mean that we have to add the `[skip-model-comp]` for all of our commits in branches we don't want tests?

pgierz:
Yes. It might make sense to invert it, and "opt in" for testing on HPC rather than "opt out".

mandresm:
I have something on the #bump job that can give you info about the PR affected, maybe we can add the labels at the PR description so that we don't have to add them everytime to the commits?

pgierz:
No version bumping here, as it just effects CI.

Issue #154, opened 2020-10-09
ECHAM-6.3.05p2 Runtime Error with Shared Libraries


pgierz:
**Describe the bug**
When running `AWIESM-2.1` on Ollie, I immediately get an error:
```
  27: /work/ollie/pgierz/test_esmtools/experiments/forced_init_test_003/run_20000101-20001231/work/./echam6: error while loading shared libraries: libsupport.so.0: cannot open shared object file: No such file or directory
```

I preformed a clean install of all the components this morning, and re-installed the model from scratch.

**To Reproduce**
You can use this runscript to reproduce the problem:
```yaml
general:
    setup_name: "awiesm"
    version: "2.1"
    compute_time: "01:00:00"
    initial_date: "2000-01-01"
    final_date: "2300-12-31"
    base_dir: "/work/ollie/pgierz/test_esmtools/experiments/"
    nmonth: 0
    nyear: 1

awiesm:
    postprocessing: false
    model_dir: "/work/ollie/pgierz/test_esmtools/model_codes/awiesm-2.1/"
    scenario: "PALEO"

echam:
    nproca: 24
    nprocb: 18
    nprocar: 24
    nprocbr: 18
    npromar: 8
    restart_rate: 1
    restart_unit: "years"

fesom:
    version: "2.0"
    res: "CORE2"
    pool_dir: "/work/ollie/pool/FESOM/"
    mesh_dir: "/work/ollie/dsidoren/input/fesom2.0/meshes/mesh_CORE2_finaltopo_mean/"
    restart_rate: 1
    restart_unit: "y"
    restart_first: 1
    lresume: 0
```

**System (please complete the following information):**
 - Supercomputer: ollie
 - Version:
```
You are using the following esm_tools versions:
-----------------------------------------------
esm_archiving : unknown version!
esm_autotests : unknown version!
esm_calendar: 4.0.1
esm_database: 4.0.0
esm_environment: 4.2.0
esm_master: 4.2.3
esm_parser: 4.2.1
esm_profile: 4.0.0
esm_rcfile: 4.0.0
esm_runscripts: 4.2.1
esm_tools: 4.2.6 (development install, on branch: release, describe=v4.2.6-1-g4116b68)
esm_plugin_manager: 4.2.0
esm_version_checker: 4.5.0
```

**Additional Info**

[...]

PR #587, opened 2022-02-23
changes in the default variant, now gamma

mandresm:
@fernandadialzira pointed out yesterday that the default `general.variant: beta` in `aiwesm-2.1.yaml` is incorrect and it should be `gamma`. I changed that. However, when looking at the `fesom` section of the `awiesm-2.1.yaml` we can notice the following:

https://github.com/esm-tools/esm_tools/blob/17ad18361418606fac4e394f2d6a2fa08abe2130/configs/setups/awiesm/awiesm-2.1.yaml#L298-L300

This lines control whether `K_gm_max` is removed from the namelist (for `beta`) or not (any other variant). The problem I see is that `K_gm_max` is also controlled here:

https://github.com/esm-tools/esm_tools/blob/17ad18361418606fac4e394f2d6a2fa08abe2130/configs/setups/awiesm/awiesm-2.1.yaml#L271-L296

**There is therefore a** `choose_` **conflict here, where there is no way of knowing what value** `fesom.K_gm_max` **will take.**

The current options (a written summary of the current conflicts) are as follow:

1. If AWI-ESM is any variant of version `2.1`, for the `ALE_schemes` `zstar` and `linfs` the `K_gm_max` in the namelist is switch to `2000.0`
2. If `default.variant` is `beta` removes `K_gm_max` from the namelist.

My question for @pgierz and @fernandadialzira: is it possible that someone wants to run `zstar` or `linfs` with a `2.1` version in the `beta` variant (removing `K_gm_max`)? If that's the case I would suggest the following:
1. Remove `K_gm_max` from the `choose_version_2_1.True` since there it is always by default `2000.0`
2. Add a the following `gamma` option:
```
        choose_general.variant:
            beta:
                K_gm_max: "remove_from_namelist"
            gamma:
                K_gm_max: 2000.0
```


fernandadialzira:
If I am not mistaken, the beta and the gamma variant are from two different fesom versions, not only the `ALE scheme`. So if one would work with the `beta` variant, he/she would be working with `fesom2.0` and the `linfs` `ALE scheme`. On the other hand, for the `gamma` variant, the fesom version is 2.1 and the `ALE scheme` is `zstar`. 

In my opinion, we would not need to include this variable variant, either gamma or beta, neither the choose block for it, as in the choose block for the model the ALE scheme is defined. Perhaps all would be solved if this part would be:  

`variant: beta` - delete

>                          version_2_1: "$(( '2.1' in '${general.version}' ))"
>                                  choose_version_2_1:
>                                          True:
>                                                  # NOTE: move potentially to AWIESM 2
>                                                  ## Settings for Gent/McWilliams based upon the selected ALE scheme.
>                                                  K_gm: 3000.0
>                                                  K_hor: 3000.0
>                                                  K_gm_max: 2000.0
>                          
>                                                  choose_ALE_scheme:
>                                                      linfs:
>                                                          K_gm: 5000.0
>                                                          K_hor: 5000.0
>                                                          K_gm_max: remove_from_namelist
>                                                      zstar:
>                                                          K_gm_max: 2000.0
>                                                          K_hor: 3000.0
>                                                          K_gm: remove_from_namelist
>                                                      "*":
>                                                          is: invalid
>                          
>                                                  add_namelist_changes:
>                                                      namelist.oce:
>                                                          oce_dyn:
>                                                              K_gm: ${K_gm}
>                                                              K_gm_max: ${K_gm_max}
>                                                          oce_tra:
>                                                              K_hor: ${K_hor}

   And delete the following.

>                             choose_general.variant:
>                                      beta:
>                                          K_gm_max: "remove_from_namelist"

However, as the zstar only has K_hor and K_gm_max, and both are the default values already detailed in `fesom2.1`'s `namelist.oce`, could it also work?:

`variant: beta` - delete

>                         version_2_1: "$(( '2.1' in '${general.version}' ))"
>                                  choose_version_2_1:
>                                          True:
>                                                  choose_ALE_scheme:
>                                                      linfs:
>                                                          K_gm: 5000.0
>                                                          K_hor: 5000.0
>                                                      "*":
>                                                          is: invalid
> 
>                                                  add_namelist_changes:
>                                                      namelist.oce:
>                                                          oce_dyn:
>                                                              K_gm: ${K_gm}
>                                                          oce_tra:
>                                                              K_hor: ${K_hor}

   And delete the following:

>                             choose_general.variant:
>                                      beta:
>                                          K_gm_max: "remove_from_namelist"

So that only the values would change if we would be specifying the `linfs` `ALE scheme`?

mandresm:
I agree that if `gamma` and `beta` are specific to the fesom version, and all of our fesom versions in `awiesm-2.1.yaml` are `2.1` then there is no point of having such an option: it's all `gamma` and we can delete the `variant` and the `choose_general.variant` block.

About the second part of the email:

> However, as the zstar only has K_hor and K_gm_max, and both are the default values already detailed in `fesom2.1`'s `namelist.oce`, could it also work?:

Are you suggesting to remove the `zstar` case because these options are the ones provided with the default `fesom` namelist? If that's the case, I will still leave it there. The reason for that is that imagine someone is using their own namelists and those variables deviate from the default, and then they try to set `ALE_scheme: zstart`. Instead of getting `zstar` they will get whatever they put in their namelist and that will be confusing. Would you agree @fernandadialzira ?

fernandadialzira:
I completely agree! Thank you for the clarification. I also set the `ALE scheme` in previous simulations, and I understand it can create trouble if it is not well defined. So perhaps we edit this part of `awiesm-2.1.yaml` to my first suggestion if you and @pgierz agree. In this way we delete the blocks to choose the variant and only add a `k_gm_max: remove_from_namelist` on the `linfs` `ALE scheme` section.

pgierz:
Apologies for the late response:

> is it possible that someone wants to run zstar or linfs with a 2.1 version in the beta variant (removing K_gm_max)?

Short answer: no. `zstar` should actually only be possible via `gamma` without serious under the hood changes. That requires namelist changes, model changes, and more than I would be comfortable giving a default user to work on...

Shorter answer: I need to give a talk about version numbers to paleodyn and how to write version numbers down. Let's say this is my next windmill to charge against....

mandresm:
Okay, so I will then proceed to do what was agreed with @fernandadialzira : remove the variant option.

mandresm:
Testing now. @pgierz , if you agree with it would you approve the PR? I will merge after successful tests.

pgierz:
Which PR is that?

mandresm:
This one :)

pgierz:
And as to failing checks....I am on that. There is a fix after a long combat with the GitHub Actions, but I want to cover all cases listed in #533 fixed before that goes in all the way

pgierz:
@mandresm sorry, I thought I was in an issue thread. too many thing are happening at once...

pgierz (approved):
👍🏻

Before this actually goes in and creates a new version number, we are sure it works, right? (@fernandadialzira your run functions with this version?) The config changes seem harmless enough, but in honesty I have not run a real production model in a while thus I am only doing "visual checks"....

mandresm:
I am running `awiesm-2.1-wiso` and `awiesm-2.1-recom` monthly restarts, 2 months (the runscripts I have implemented in esm_tests). I'll also wait for @fernandadialzira for her own tests.

fernandadialzira:
I am currently stuck on [this problem](https://github.com/esm-tools/esm_tools/issues/154#issuecomment-1048969256) referenced here in #154. So when I try to run any test I am getting the same error regarding the echam libraries. If you @mandresm and @pgierz have any ideas I could try. As in my explanation there, I tried several ways to handle the library problem, but so far no resolution.

mandresm:
Currently running esm_tests in ollie

fernandadialzira:
@mandresm and @pgierz

After Paul found out that to run the bootstrap simulations we needed to switch to echam `ice` branch, I ran this with the esm-tools `release` and `fix/awiesm_default_variant` branches and both simulations worked out. 
In fact, by using this branch I am also able to run rho-amoc simulations which require wiso turned off. With release, it would still find the isotopes fields, and I would have to comment them out by hand in `echam.yaml`. 

So from my tests, this could be merged to the main branch

pgierz:
I think this PR will be superseded by #623?

pgierz:
I think this PR will be superseded by #623?

fernandadialzira:
Yes, at least from what I have seen of #623, the changes in this will be applied there, but meanwhile I am still using this branch

mandresm:
> I think this PR will be superseded by #623?

I'll be closing it then. I'll keep the branch alive for @fernandadialzira

PR #596, opened 2022-03-01
Feature/add_foci_default_oasismct4

seb-wahl:
Adds a new FOCI configuration that uses OASIS3MCT4 instead of OASIS3MCT2.8
all coupling fields are interpolated using gausweights as some of the old methods are not supported anymore.
scientific validation still ongoing. Technical stable in a 50 year run


pgierz (approved):

pgierz:
@seb-wahl, does this count as a "new feature" or rather as a fix? 

I don't want to trigger @mandresm CI magic just yet, I still want to use the afternoon to attempt to figure out why the PyPI publishing is broken.

seb-wahl:
It's a feature as we introduce coupling with the latest version of OASIS3MCT4 with FOCI. The new pull request which I'll put together is based on this one, so this one has to be merged first.

pgierz:
Alright, then let's do the magic again:

#bump

pgierz:
#bump

@mandresm does the CI magic need to be on the first line? Or did it get stuck because there were conflicts?

pgierz:
@mandresm: Nevermind, it showed up. @seb-wahl: `6.2.0` now has this in it. I will upload to PyPI by hand, that is still having strange things happening if you do it automatically.

mandresm:
We should probably not bumping minors so easily. From yesterday night until today we are 2 minors up. Users might not like that. I fixed a couple of things in the pre-merge magic so that now, whatever you specify in your comment wins over the branch name, for example, here I would have done <hash>bump <hash>patch.

seb-wahl:
Well I don't think we do not add normally new features that often as we do at the moment. Maybe we should discuss our strategy w.r.t semantic versioning next Monday which boils down to the question: Does every new feature really deserve a new version number increase, i.e. such as the increase from `6.1.0` to `6.2.0`?

mandresm:
Noted, I'll add that to the agenda

PR #599, opened 2022-03-03
Add new FOCI config, make changing coupling freq in echam within FOCI

seb-wahl:
The changes here already include the #596 so #596 has to be merged first. I'm sure the merge conflict will then disappear

PR #601, opened 2022-03-03
fix for the new nemo version coupling

mandresm:
Fixes the bugs in `esm_master` introduce in #599


pgierz:
Is this what I just emailed Laura about?

mandresm:
I guess, everyone is having problems apparently... I'm also getting a bunch of emails

seb-wahl (approved):

seb-wahl:
I don't understand why a merge broke this since it worked for me. Maybe some merge magic overwrote things the way it shouldn't. And the basic test the automatic tests do should have caught this I think.

pgierz:
#bump

pgierz (approved):

PR #602, opened 2022-03-04
fix(setup.py): removes Python 3.10 compatibility

pgierz:
Python 3.10 has features that currently break the parser, so we disallow this version.


pgierz:
See also here for more info on the `python_requires` argument in `setup.py`

https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

pgierz:
#bump

mandresm (approved):

PR #603, opened 2022-03-04
Fix/icb fesom namelist suggestions

mandresm:
So this are all my suggestions for the branch.

The only thing I see it could destroy iceberg simulations right now with this PR is the `icb_code` switch. Once it is clear to us in which `fesom-2.1` versions we should include `icb_code: true` then we can overcome the problem in an automatic manner. For the time being, @ackerlar will need to add to his runscript the following:
```
fesom:
    icb_code: true
```


mandresm:
@ackerlar feel free to merge it to your branch (`fix/icb_fesom_namelist`) if you agree with the changes.

ackerlar (approved):
All changes look good to me!

PR #608, opened 2022-03-10
ci: enables unit tests to be written in docstrings as examples

pgierz:


pgierz:
We still should at some point clarify what to do with changes that only influence CI. I would say that this should not influence the version number, as the actual software the user downloads is not modified.

mandresm (approved):
I say `docs` and `CI` don't bump. You can go ahead and merge

PR #611, opened 2022-03-16
fix(fesom2.1): fix for pool directories, forcing, and mesh rotation in FESOM 2.1

pgierz:
The standard runscript provided for FESOM 2.1 does not work on Ollie. I modified a few things so that it runs out of the box again.


mandresm:
I thought the recommended force_rotation is true. By the way FESOM2.1 works for me out of the box in ollie. Can you provide the runscript path so that I can compare it with mine?

pgierz:
Sure, the path is here:

```
/work/ollie/pgierz/SciComp/HPC_Tests/run_configs
```

Also, the diff to the supplied one:

```
$ diff /work/ollie/pgierz/SciComp/HPC_Tests/run_configs/fesom2-ollie-initial-monthly.yaml /work/ollie/pgierz/SciComp/HPC_Tests/ssoftware/github.com/esm-tools/esm_tools/runscripts/fesom2/fesom2-ollie-initial-monthly.yaml
```
Yields:

```diff
6c6
<         base_dir: "/work/ollie/pgierz/SciComp/HPC_Tests/experiments"
---
>         base_dir: "/work/ollie/dbarbi/esm_yaml_test/"
12,13c12,14
<         version: 2.1
<         model_dir: "/work/ollie/pgierz/SciComp/HPC_Tests/model_codes/fesom-2.1/"
---
>         version: 2.0
>         model_dir: "/work/ollie/dbarbi/modelcodes/fesom-2.0/"
>         pool_dir: "/work/ollie/pool/FESOM/"
```

mandresm:
My runscript and yours (`work/ollie/pgierz/SciComp/HPC_Tests/run_configs/fesom2-ollie-initial-monthly.yaml`) are identical except that mine has no `mesh_dir`.

`fesom-2.1.yaml` sets up the default mesh used by fesom developers automatically. In your script, you are using the old rotated core mesh, and that's why is not working for you out of the box. In my opinion, what we need is a fesom-2.1 runscript.

configs/components/fesom/fesom-2.1.yaml
```diff
@@ -63,7 +63,9 @@ pool_dir: "${computer.pool_directories.pool}"
 
 choose_computer.name:
     ollie:
-        pool_dir: "${computer.pool_directories.pool}/../clidyn/"
+        pool_dir: "${computer.pool_directories.pool}/FESOM/"
+        climate_data_dir: "${pool_dir}/hydrography/"
+        forcing_data_dir: "${computer.pool_directories.projects}/clidyn/forcing/"
```
mandresm:
```suggestion
        pool_dir: "${computer.pool_directories.pool}/../clidyn/"
```
I suggest to bring back the old configuration as it is the default provided by FESOM developets and then tested with the new `fesom-2.1` runscript I added. This mesh is unrotated in contrast with the one you added on your first commit.

namelists/fesom2/2.1/namelist.config
```diff
@@ -38,7 +38,7 @@ cartesian=.false.
 fplane=.false.
 cyclic_length=360       ![degree]
 rotated_grid=.true.     !option only valid for coupled model case now
-force_rotation=.true.
+force_rotation=.false.
```
mandresm:
```suggestion
force_rotation=.true.
```
@koldunovn can you confirm that I am understanding correctly? Default FESOM mesh is unrotated, and `force_rotation` should be set to `.true.` by default?

mandresm (changes requested):

configs/components/fesom/fesom-2.1.yaml
mandresm: Additionally, the new runscript will also work in mistral (and in the future in Levante) because fesom-2.1 already handles correctly the pools there too.

namelists/fesom2/2.1/namelist.config
koldunovn: Yes, by default meshes in FESOM2.1 are unrotated (or rather not rotated :)) and `force_rotation` therefore should be `true`.

mandresm:
#bump

mandresm (approved):

Issue #612, opened 2022-03-16
Bug for new switch with_wiso in awiesm-2.1-wiso for esm_tools 6.x


mwerner-awi:
There seems to be a severe bug with the new switch `with_wiso`, which was introduced for awiesm-2.1-wiso in esm_tools 6.x:

My yaml file for running awiesm-2.1-wiso does not contain an entry for `with_wiso` in the block `general`. But I switch off water isotopes with `lwiso: False` entries in both the `echam` and `fesom` block. 

In esm_tools version 5.x, calling esm_runscripts resulted in a correct entry `lwiso = .false.` in the FESOM namelist.config file.
In esm_tools version 6.3.3, calling esm_runscripts results in a **wrong** entry `lwiso = .true.` in the FESOM namelist.config file.

Apparently, the default setting `with_wiso: true` in `esm_tools/configs/setups/awiesm/awiesm-2.1.yaml` "overrides" the setting in my own yaml file. 

**Important to know:** The awiesm-2.1-wiso model runs with this erroneous configuration without any error messages(!). As expected, no water isotope files are generated. However, default ECHAM and FESOM variables differ from a simulation with water isotopes switched on, which must not be the case.

The error was found running awiesm-2.1-wiso on ollie.

**Suggestion:**
When fixing this bug, I suggest a default setting `with_wiso: false` in `esm_tools/configs/setups/awiesm/awiesm-2.1.yaml` for all model versions. Water isotopes should only be included in a simulation if they are explicitly switched on by a user.

mandresm:
Yesterday, while working on awiesm-2.1.yaml improvements I also detected this bug. I am working already in a solution.

mandresm:
@mwerner-awi , I am also doing some cleanup of the wiso switches (`wiso_code` and `with_wiso`). It would be useful for me to have the path of your runscript so that I can test it after I fix the bug and add your suggestions to the yamls. Could you please provide that path?

mwerner-awi:
@mandresm: You can use the following yaml files for testing:
/home/ollie/mwerner/runscripts/awiesm-2.1-wiso/PI_ctrl_awiesm-2.1-wiso_ollie.10nodes.yaml
/home/ollie/mwerner/runscripts/awiesm-2.1-wiso/PI_ctrl_awiesm-2.1-wiso_ollie.10nodes.no_wiso.yaml

mandresm:
[...]

PR #615, opened 2022-03-17
Fix/find variable transformations and choose booleans

mandresm:
This PR solves the following problem:
```
general:
    a_var_in_general: true
fesom:
    a_var_in_fesom: ${general.a_var_in_general}
    choose_a_var_in_fesom:
         True:
             a_variable: true
```

Under this situation, `general.a_var_in_general` is interpreted in the `config` as a boolean, however, when searched from the fesom using the `esm_parser.find_variable` method it returns `"True"` as a string. This is because of a bug in `find_variable` that apparently was known to exist (there was a big BUG comment in the lines affected). With this bug, the correct option in the choose is not selected because `True` is not `"True"`.

Additionally to this fix, this also solves:
- renames the misleading `ok_part` and `new_raw` variables to `prefix` and `suffix` variables
- solves a subsequent bug introduce by this fix, so that 0 and 1 are understand as True or False as well
- solves a subsequent bug introduce by this fix that affects `PrevRunInfo` classes when using `find_variable`

**TODO**

- [ ] Heavy testing with `esm_tests`. **Very important** to make sure nothing changes in the configuration files or namelists in terms of variable types. Full compilation and runs will be run in all computers.

### DIFF
--- a/src/esm_parser/esm_parser.py
+++ b/src/esm_parser/esm_parser.py
@@ -1494,8 +1494,8 @@ def resolve_basic_choose(config, config_to_replace_in, choose_key, blackdict={})
     for ckey in choices_available:
         choices_bool &= isinstance(ckey, bool)
     # If the choices are booleans and the ``choice`` is a string, try to transform the
-    # try to transform the string in an integer (that will be able to select a choice
-    # from the boolean choices
+    # string in an integer (that will be able to select a choice from the boolean
+    # choices)
     if choices_bool and isinstance(choice, str):
         if choice == "0" or choice == "1":
             choice = int(choice)
--- a/src/esm_parser/esm_parser.py
+++ b/src/esm_parser/esm_parser.py
@@ -1971,8 +1971,11 @@ def find_variable(tree, rhs, full_config, white_or_black_list, isblacklist):
             # If the substituted variable is not a list, and there is either a
             # preceding (``prefix``) or following (``suffix``) string, then add up
             # the parts, making sure that other variables (``${}``) are also
-            # substituted
-            if type(var_result) not in [list] and (prefix or suffix):
+            # substitute. The "NONE_YET" part is to handle PrevRunInfo class correctly
+            if (
+                (not isinstance(var_result, list) and (prefix or suffix)) or
+                (isinstance(var_result, dict) and "NONE_YET" in var_result)
+            ):
                 prefix, var_result, more_rest = (
                     str(prefix),
                     str(var_result),
--- a/src/esm_parser/esm_parser.py
+++ b/src/esm_parser/esm_parser.py
@@ -1489,6 +1489,17 @@ def resolve_basic_choose(config, config_to_replace_in, choose_key, blackdict={})
         else:
             choices_available[ckey] = cval
 
+    # Are choices all booleans?
+    choices_bool = True
+    for ckey in choices_available:
+        choices_bool &= isinstance(ckey, bool)
+    # If the choices are booleans and the ``choice`` is a string, try to transform the
+    # try to transform the string in an integer (that will be able to select a choice
+    # from the boolean choices
+    if choices_bool and isinstance(choice, str):
+        if choice == "0" or choice == "1":
+            choice = int(choice)
+
     # Resolve the choose variables
     if choice in choices_available:
         for update_key, update_value in six.iteritems(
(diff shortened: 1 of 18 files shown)


mandresm:
This is part of the solution for #612 and a requirement for keep moving forward the `feat/awiesm_final` branch.

src/esm_parser/esm_parser.py
```diff
@@ -1926,8 +1937,8 @@ def find_variable(tree, rhs, full_config, white_or_black_list, isblacklist):
     if not tree[-1]:
         tree = tree[:-1]
     if isinstance(raw_str, str) and "${" in raw_str:
-        ok_part, rest = raw_str.split("${", 1)
-        var, new_raw = rest.split("}", 1)
+        prefix, rest = raw_str.split("${", 1)
+        var, suffix = rest.split("}", 1)
```
denizural, line 1941: 👍 for choosing good variable names.

src/esm_parser/esm_parser.py
```diff
@@ -1943,7 +1954,7 @@ def find_variable(tree, rhs, full_config, white_or_black_list, isblacklist):
                         isblacklist,
                     )
 
-                if "$((" in var_result:
+                if isinstance(var_result, str) and "$((" in var_result:
```
denizural, line 1957: Do we really need to check the `str` instance here? It is already inside the `if type(var_result) == str:` block

src/esm_parser/esm_parser.py
```diff
@@ -1489,6 +1489,17 @@ def resolve_basic_choose(config, config_to_replace_in, choose_key, blackdict={})
...
             choices_available[ckey] = cval
 
+    # Are choices all booleans?
+    choices_bool = True
+    for ckey in choices_available:
+        choices_bool &= isinstance(ckey, bool)
+    # If the choices are booleans and the ``choice`` is a string, try to transform the
+    # try to transform the string in an integer (that will be able to select a choice
+    # from the boolean choices
+    if choices_bool and isinstance(choice, str):
```
denizural: I am not sure but it might be redundant to check for `if choices_bool`. Simply checking ` if choice == "0" or choice == "1":` might be sufficient since both of these are evaluated as `True`.

denizural (approved):
Apart from the comments I made it looks very nice. 👏 👏

src/esm_parser/esm_parser.py
mandresm: Yes, `var_result` that is a boolean, integer or float some times reaches this line, resulting in a crash.

src/esm_parser/esm_parser.py
mandresm: We want to check if `choices_bool` are boleans because you could have a choose for version numbers:
```
version: 1.0
choose_version:
     1.0:
           some_variables: here
     2.0:
           some_other_vars: here
```
In this case, we should not transform the `version` variable into an integer because or set of choices are actually version numbers.

mandresm:
> Apart from the comments I made it looks very nice. 👏 👏

Thanks for your fast review :) please, can you resolve the conversations where the answer was satisfactory for you or further comment if it was not? :)

src/esm_parser/esm_parser.py
mandresm: However, if we have the following:
```
with_wiso: 1
choose_with_wiso:
     true:
           some_variables: here
     false:
           some_other_vars: here
```
Then we want to transform `with_wiso` into an `int` cause an `int` can resolve the boolean choice.

src/esm_parser/esm_parser.py
denizural: Ah, I see that it is returned again inside the if block. Sorry

src/esm_parser/esm_parser.py
```diff
@@ -1489,6 +1489,17 @@ def resolve_basic_choose(config, config_to_replace_in, choose_key, blackdict={})
         else:
             choices_available[ckey] = cval
 
+    # Are choices all booleans?
+    choices_bool = True
```
pgierz: Small suggestion: all_choices_are_bool

pgierz (changes requested):
One super tiny suggestion only :-)

src/esm_parser/esm_parser.py
mandresm: Apparently I cannot push my commit with your suggestion for some weird reason...

mandresm:
All tests passed. Time to #bump

Issue #616, opened 2022-03-18
It is not possible to use a manual Recom namelist


denizural:
**Describe the bug**
It is described in the documentation (https://esm-tools.readthedocs.io/en/latest/recipes/use_own_namelist.html?highlight=namelist_dir) that it should be possible to use a manual namelist via the required changes in the runscript.

**To Reproduce**
`esm_runscripts -e test fesom-recom-ollie-initial-control.yaml`
Same behavior during `Fesom-Recom` or `AWICM-Recom` on both `Mistral` and `Ollie`

**Expected behavior**
Manual namelist gets copied to the `workdir`

**System (please complete the following information):**
 - Supercomputer: Mistral, Ollie
 - Version `HEAD`

**Additional context**
Bug has been fixed, I will shortly make a pull request

mandresm:
This is not a bug.

EDIT: @denizural proposed a more specific solution below.

`namelist_changes` is doing here what it supposed to do: changing the namelist, no matter where it comes from.

@chrisdane , if you wish to change this behaviour it is already possible by using the `remove` feature in the runscript:
```
<model>:
    remove_namelist_changes.<namelist_of_your_choice>:
        - <section1_of_the_namelist>
        - <section2_of_the_namelist>
```
This will remove all the default changes ESM-Tools do in `section1` and `section2` of the namelist of your choice.

If you want to only remove specific variable `<var1>` changes in the namelist then you can go one level further:
```
<model>:
    remove_namelist_changes.<namelist_of_your_choice>.<section_of_your_choice>:
        - <var1>
```

I believe @denizural is working on a feature where, when you indicate a different path for the namelist other than the default, ESM-Tools warns you about what is being changed by default at the end of the esm_runscript submission.

@chrisdane, any suggestion for improvement is welcome, however bare in mind that `namelist_changes` is core functionality of ESM-Tools and controls of certain namelist variables, such as restart switches, run length control, etc. are important features for simulations to be able to run in batches of successive smaller runs.

denizural:
Adding these lines to the user runscript (to `Recom` section) solves the issue:
```yaml
namelist_dir: "/path/to/manual/namelist/directory/"
add_config_sources:
    # user-given namelist whose name differs from namelist.recom
    recom: ${namelist_dir}/namelist.recom_co2_coupling 
config_in_work:
    # user-given namelist will be renamed correctly in the work/ directory
    recom: namelist.recom
    
# rest of the namelist changes
namelist_changes:
[...]

PR #617, opened 2022-03-18
solves issue esm-tools/esm_tools/issues/616

denizural:
This bugfix solves the issue https://github.com/esm-tools/esm_tools/issues/616

When I debug the code, I can see that the parser adds `add_config_sources` to the target but since it is appended it does not have any effect:
```python
target_config['recom']['config_sources']
{'recom': '${namelist_dir}/namelist.recom', 
 'namelist.recom': '/mnt/lustre01/pf/a/a271096/Solutions/Christopher/2022/namelist.recom'
}
```
A simple solution is to modify the Recom YAML file, so that the parser can safely overwrite the `namelist.recom` key. Another solution is also possible but I wanted to make it look similar to Echam example in the docs so that it is consistent.

### DIFF
--- a/configs/components/recom/recom.yaml
+++ b/configs/components/recom/recom.yaml
@@ -106,11 +106,8 @@ choose_lresume:
                                 pavariables:
                                         recom_restart: false
 
-config_files:
-        recom:  recom
-
 config_sources:
-        recom:  "${namelist_dir}/namelist.recom"
+        "namelist.recom":  "${namelist_dir}/namelist.recom"
 #        io.recom.ciso: "${namelist_dir}/namelist.io.recom.ciso"
 
 restart_name: ""


mandresm:
This does not solve the problem and breaks other setups. It won't be merged in this state. See issue #616 for more details.

denizural:
Adding these lines to the user runscript (to `Recom` section) solves the issue:
```yaml
namelist_dir: "/path/to/manual/namelist/directory"
add_config_sources:
    # user-given namelist whose name differs from namelist.recom
    recom: ${namelist_dir}/namelist.recom_co2_coupling 
config_in_work:
    # user-given namelist will be renamed correctly in the work/ directory
    recom: namelist.recom
    
# rest of the namelist changes
namelist_changes:
    namelist.recom:
        pavariables:
            REcoMDataPath: '/work/ba1103/a270094/MarESys/forcing/CORE2mesh/'
```

Issue #593, opened 2022-02-28
Problem on AWICM3 installation on JUWELS


shijian0702:
Hi all,

I have some problems when installing AWICM3 on JUWELS, which I think may be related to the netcdf-Fortran configuration.
I will be grateful if anyone can have a look at it. By the way, this issue cannot be reproduced in other HPCs, like ollie, mistral.

Best,
Jian

**System**
 - Supercomputer: JUWELS (Stages 2022, Python/3.9.6, netCDF-Fortran/4.5.3-serial)
 - ESM-Tools 6.1**
 
I used the command: **esm_master install-awicm3-deck**, and it shows errors after asking for permission of all the components: 
(I also alter different stages, python, netcdf versions, but the same errors remain)

-- setting default build type: Release
-- The C compiler identification is Intel 2021.4.0.20210910
-- The CXX compiler identification is Intel 2021.4.0.20210910
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /p/software/juwels/stages/2022/software/OpenMPI/4.1.2-intel-compilers-2021.4.0/bin/mpicc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /p/software/juwels/stages/2022/software/OpenMPI/4.1.2-intel-compilers-2021.4.0/bin/mpic++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- The Fortran compiler identification is Intel 2021.4.0.20210910
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Check for working Fortran compiler: /p/software/juwels/stages/2022/software/OpenMPI/4.1.2-intel-compilers-2021.4.0/bin/mpifort - skipped
-- Checking whether /p/software/juwels/stages/2022/software/OpenMPI/4.1.2-intel-compilers-2021.4.0/bin/mpifort supports Fortran 90
-- Checking whether /p/software/juwels/stages/2022/software/OpenMPI/4.1.2-intel-compilers-2021.4.0/bin/mpifort supports Fortran 90 - yes
-- Looking for Fortran NCCRE
-- Looking for Fortran NCCRE - not found
-- Looking for nc__create
-- Looking for nc__create - not found
-- Configuring done
[...]

PR #618, opened 2022-03-18
fix/path in juwels awicm3

mandresm:
Fixes AWICM3 not running in Juwels due to an improper ordering in the PATH environment variable.

#593 is probably not solved previous to this fix. This PR will fix the problem.


JanStreffing (approved):

mandresm:
#bump

PR #622, opened 2022-03-23
feat/esm_tests

mandresm:
Again, not a finished feature, but a small development that will allow the use of esm_tests by different people because now it does not save the local paths anymore in the `last_tested` files. This will be bumped as a `patch`.


pgierz (approved):
I don't know enough about the internals to give a good feedback, to be honest. But, from what I can tell, looks good.

mandresm:
#bump #patch

PR #625, opened 2022-03-25
fix/echam cleanup wiso and concurrent radiation code

mandresm:
This PR cleans up the ECHAM versions in the `echam.yaml` by creating a new switch: `concurrent_radiation_code`. 

Advantages:
- reduces the length and complexity of the AWIESM file
- sets a common framework for versions with concurrent radiation code, so that setting up a new version is easier
- should reduce the length of runscripts, that are not typically having a lot of `namelist_changes`. 

This switch follows the phylosophy of the `wiso_code` and `icebergs_code`: it is bounded to the source-code versions and sets up special options for the code to run successfully, independently of whether you are trying to run with or without concurrent radiation.

### DIFF
--- a/configs/components/echam/echam.yaml
+++ b/configs/components/echam/echam.yaml
@@ -201,6 +201,7 @@ choose_version:
         6.3.05p2-awiesm-2.1:
                 # TODO MA: this problably needs much more stuff from the wiso version
                 wiso_code: true
+                concurrent_radiation_code: true
                 repo_tag: not_applicable
                 dataset: r0007
         6.3.04p1:
@@ -230,18 +231,7 @@ choose_version:
                 lrad_async: "remove_from_namelist"
 
         6.3.05p2-concurrent_radiation-paleodyn:
-                choose_computer.name:
-                    mistral:
-                        compiletime_environment_changes:
-                            add_export_vars:
-                                NETCDFROOT: "/sw/rhel6-x64/netcdf/netcdf_c-4.3.2-gcc48"
-                                LD_LIBRARY_PATH: "/sw/rhel6-x64/netcdf/parallel_netcdf-1.6.1-impi-intel14/lib/:$LD_LIBRARY_PATH"
-                        runtime_environment_changes:
-                            add_export_vars:
-                                NETCDFROOT: "/sw/rhel6-x64/netcdf/netcdf_c-4.3.2-gcc48"
-                                LD_LIBRARY_PATH: "/sw/rhel6-x64/netcdf/parallel_netcdf-1.6.1-impi-intel14/lib/:$LD_LIBRARY_PATH"
-                test1:
-                    test2: true
+                concurrent_radiation_code: true
                 repo_tag: not_applicable
                 dataset: r0007
                 add_streams:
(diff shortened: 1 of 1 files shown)


configs/components/echam/echam.yaml
```diff
@@ -230,18 +231,7 @@ choose_version:
...
-                    mistral:
-                        compiletime_environment_changes:
-                            add_export_vars:
-                                NETCDFROOT: "/sw/rhel6-x64/netcdf/netcdf_c-4.3.2-gcc48"
-                                LD_LIBRARY_PATH: "/sw/rhel6-x64/netcdf/parallel_netcdf-1.6.1-impi-intel14/lib/:$LD_LIBRARY_PATH"
-                        runtime_environment_changes:
-                            add_export_vars:
-                                NETCDFROOT: "/sw/rhel6-x64/netcdf/netcdf_c-4.3.2-gcc48"
-                                LD_LIBRARY_PATH: "/sw/rhel6-x64/netcdf/parallel_netcdf-1.6.1-impi-intel14/lib/:$LD_LIBRARY_PATH"
-                test1:
```
pgierz, line 243: thank you for finally throwing this out.

configs/components/echam/echam.yaml
```diff
@@ -332,6 +275,31 @@ choose_version:
...
+                        namelist.echam:
+                                parctl:
+# kh 30.04.20 nprocrad is replaced by more flexible partitioning using nprocar and nprocbr
+#                                       nprocrad: "${nprocrad}"
+                                        nprocar: "${nprocar}"
+                                        nprocbr: "${nprocbr}"
+                                radctl:
+                                        lrad_async: "${lrad_async}"
+        "*":
+                from_concurrent_radiation_code_choose: <NO_CONCURRENT_RADIATION_CODE>
```
pgierz, line 300: this one is just for debugging?

configs/components/echam/echam.yaml
```diff
@@ -529,6 +497,24 @@ choose_wiso_code:
                                 add_coupling_fields:
                                         "[[wiso_fields-->FIELD]]":
                                                 grid: atmo
+                                add_streams:
+                                        - wiso
+                                        - accw_wiso
+                                        - sf_wiso
+                                        - rad
+                                add_forcing_files:
```
pgierz, line 508: I am not sure here: is the ozone[a,b,c] specific to wiso?

configs/components/echam/echam.yaml
```diff
@@ -529,6 +497,24 @@ choose_wiso_code:
...
+                                        ozonec: piozonec
+                                add_forcing_in_work:
+                                        ozonea: "ozon1988"
+                                        ozoneb: "ozon1989"
+                                        ozonec: "ozon1990"
+                                restart_in_sources:
+                                        "[[streams-->STREAM]]": restart_${parent_expid}_${parent_date!syear!smonth!sday}_STREAM.nc
+                                restart_out_sources:
+                                        "[[streams-->STREAM]]": restart_${general.expid}_${other_date!syear!smonth!sday}_STREAM.nc
+
```
mandresm, line 516: @xshi-awi , @mwerner-awi I moved all this from the `choose_version.6.3.05p2-wiso` to be under the `with_wiso` switch. Is that correct? (i.e. a peson is running echam with wiso code but don't want WISO in their simulation, then all these variables here are not needed for that)

configs/components/echam/echam.yaml
mandresm: Is just so that even if the `concurrent_radiation_code` doesn't exist ESM-Tools does not complain. Because this is a dictionary `"*":` cannot be empty, therefore I place that dummy variable.

configs/components/echam/echam.yaml
mandresm: me neither, but it was only under the version `6.3.05p2-wiso`.

configs/components/echam/echam.yaml
mandresm: This is the whole block for which I would like to have your feedback:
```yaml
                choose_with_wiso:
                        True:
                                nwiso: 3
                                lwiso_rerun: ${lresume}
                                add_coupling_fields:
                                        "[[wiso_fields-->FIELD]]":
                                                grid: atmo
                                add_streams:
                                        - wiso
                                        - accw_wiso
                                        - sf_wiso
                                        - rad
                                add_forcing_files:
                                        ozonea: piozonea
                                        ozoneb: piozoneb
                                        ozonec: piozonec
                                add_forcing_in_work:
                                        ozonea: "ozon1988"
                                        ozoneb: "ozon1989"
                                        ozonec: "ozon1990"
                                restart_in_sources:
                                        "[[streams-->STREAM]]": restart_${parent_expid}_${parent_date!syear!smonth!sday}_STREAM.nc
                                restart_out_sources:
                                        "[[streams-->STREAM]]": restart_${general.expid}_${other_date!syear!smonth!sday}_STREAM.nc
```

pgierz: I think the `restart_in_sources` and `restart_out_sources` can be outside of the `choose_with_wiso` block. The only important part (as far as I know) is the add stream and add coupling field part.

pgierz:
Failing tests are already on my radar; this has something to do with the automatic sending of username and passwords. I am almost positive I already solved that, but it must have gotten reverted somehow.

configs/components/echam/echam.yaml
mandresm: you are right about the `retart_in_sources` and `restart_out_sources`, it's all happening from the `add_stream`. What about the `forcing_files` then? is this something `with_wiso` specific? @xshi-awi , @mwerner-awi

mwerner-awi:
Only flies labeled _wiso are related to the WISO code, everything else is not.

PR #629, opened 2022-03-29
Preview: ESM Parser

pgierz:
This is a technical preview of what the future of ESM Parser may eventually look like after a long refactoring overhaul. You can subscribe to this PR to have a look at the progress if you are interested.

You will see that it is much more object oriented and (hopefully) clearer and more Pythonic 😉 

@mandresm @denizural @seb-wahl


denizural:
Looks great @pgierz. 👏 I am also trying to finish some work for the retreat.
I see that you are now using ruamel yaml instead of the older one. We may need to update some other tools for consistency and also the `setup.py` requirements.
I am looking forward to testing these. Maybe you can give us a quick introduction next week.

src/esm_parser/esm_parser.py
```diff
@@ -201,13 +223,22 @@ def _filter_variable(value):
...
-                    address=ConfigAddress(result[0]),
-                    should_resolve_from=result[1],
-                    is_resolved=False,
+            logger.debug(f"Full result: {result}")
+            re_pattern = re.escape(variable_start) + r"(.*?)" + re.escape(variable_end)
+            logger.debug(rf"Pattern to search with: {re_pattern}")
+            logger.debug(f"Result: {result[1]}")
+            regex_results = re.findall(re_pattern, result[1])
+            logger.debug(regex_results)
+            for regex_result in regex_results:
```
denizural, line 232: It could be good idea to check if `regex_results` contain any data or not. If it is empty then an exception can be thrown.

pgierz: Possibly, yes. But, if it's empty, that just means you don't have any variables at all in the config you passed in, so the valid answer would be an empty list, which is what you get. I can write a test for that though

PR #634, opened 2022-03-31
added missing esm_parser statements to avoid a crash

seb-wahl:
tiny fixes to avoid a crash when the `while True` loop is actually used.


mandresm (approved):

mandresm:
#bump

PR #635, opened 2022-03-31
Add Support for NEMO4.2

seb-wahl:
Big step for us at GEOMAR: run a NEMO4.2 (which will be the basis for the next version of FOCI) test case within ESM-Tools.
Includes simplifications in `glogin.yaml` and `blogin.yaml` based on the layout used in `juwels.yaml`, so I know it's robust.


mandresm (approved):
None of these changes will affect other setups other than NEMO so thumbs up from my side. I can still run `esm_tests` in blogin, for the GEOMAR setups if you want @seb-wahl . Let me know if you'd like me to merge it straight away and run the testing.

configs/other_software/vcs/svn.yaml
```diff
@@ -1,2 +1,7 @@
-nothing:
-        - so far
+get_command: "svn co ${define_options} ${define_revision} ${repository}"
```
pgierz, line 1: Hooray for a finally useful `svn.yaml`! Most things are in git, but nevertheless, good that we can add it to the list of things we support

configs/components/nemo/nemo.yaml
```diff
@@ -609,18 +648,41 @@ namelist_changes:
...
+                        #nn_msh: ${nn_msh} 
+                        #rn_rdt: ${time_step}
                 namsbc:
                         nn_fsbc: ${coupling_freq_in_steps}
                 nammpp:
                         jpni: ${jpni}
                         jpnj: ${jpnj}
-                        jpnij: ${nproc}
+
+choose_generation:
```
pgierz, line 660: Is generation a well known term in the NEMO world?

configs/components/nemo/nemo.yaml
```diff
@@ -609,18 +648,41 @@ namelist_changes:
...
                         ln_rstart: ${nemo.lresume}
                 namtsd:                                                
-                        ln_tsd_tradmp: ${ln_tsd_tradmp} 
+                        #ln_tsd_tradmp: ${ln_tsd_tradmp} 
                         ln_tsd_init: ${ln_tsd_init} 
-                namdom:
-                        nn_closea: 1
-                        nn_msh: ${nn_msh} 
-                        rn_rdt: ${time_step}
+                #namdom:
```
pgierz: I would always advocate for deleting code rather than just commenting it. If needed we can always go into the git history. But, that's just my personal style.

runscripts/nemo/nemo4-GYRE_PISCES.yaml
```diff
@@ -0,0 +1,82 @@
+# runscript to run the GYRE_PISCES demo case that comes with NEMO4.2
```
pgierz, line 1: If this is a runscript that could (at some point in my dreamy-future) be included in CI, please let me know.

pgierz (approved):

configs/components/nemo/nemo.yaml
seb-wahl: You are right. I'll push another commit

configs/components/nemo/nemo.yaml
seb-wahl: I don't know, but we need something to distinguish between 3.6.x and 4.2.x

runscripts/nemo/nemo4-GYRE_PISCES.yaml
seb-wahl: Not at the moment, as we are still in the "alpha" stage with NEMO4.2.x

configs/other_software/vcs/svn.yaml
seb-wahl: which we don't need anymore, as the NEMO people finally switched to git a few weeks ago, hooray

configs/components/nemo/nemo.yaml
pgierz: How about version?

seb-wahl:
Since we add a new feature, and if we follow our versioning idea, this one should be `6.4.0`

mandresm:
#bump!!!

PR #637, opened 2022-04-06
Cleanup/remove rcfile

seb-wahl:
tested on blogin on a fresh install with FOCI with `.esmtoolsrc` removed.

PR #638, opened 2022-04-06
Cleanup/remove shell runscript support including rcfile removal

seb-wahl:
This pull request is a cleanup and removes the unused shell support, and hopefully makes the tools a little more robust. Since this branch is based on https://github.com/esm-tools/esm_tools/pull/637 is also closes #637 (i.e. it contains the changes from #637)
contact @seb-wahl if questions remain.


pgierz:
I think I will need to go through this commit by commit, since 700 deleted lines are too much for my brain to handle all at once. I will (very probably) give a thumbs up by tomorrow lunch time: unless I see something absolutely weird but I do not foresee that :-)

And for the others seeing the PR: Good work for everyone at the retreat! I think we did a really great job cleaning up, making new ideas, and getting the new machine supported :-)

pgierz:
Note to whoever merges this later: Please put in **both** #637 as well as this one, I just updated some documentation in the wrong branch.

src/esm_environment/esm_environment.py
```diff
@@ -87,7 +86,7 @@ def __init__(self, run_or_compile, complete_config=None, model=None):
                 self.apply_config_changes(run_or_compile, complete_config, model)
 
         # Add the ENVIRONMENT_SET_BY_ESMTOOLS into the exports
-        self.add_esm_var()
+        # self.add_esm_var()
```
pgierz: can we just directly remove this line? Along with the corresponding method?
```suggestion

```

src/esm_environment/esm_environment.py
```diff
@@ -476,7 +469,7 @@ def get_shell_commands(self):
 
         environment = []
         # Write module actions
-        if "module_actions" in self.config:
+        if self.config.get("module_actions", None) is not None:
```
pgierz: `get` by default returns `None` if it is not set.

```suggestion
        if self.config.get("module_actions") is not None:
```

src/esm_environment/esm_environment.py
```diff
@@ -486,7 +479,7 @@ def get_shell_commands(self):
                     environment.append(f"module {action}")
         # Add an empty string as a newline:
         environment.append("")
-        if "export_vars" in self.config:
+        if self.config.get("export_vars", None) is not None:
```
pgierz:
```suggestion
        if self.config.get("export_vars") is not None:
```

src/esm_environment/esm_environment.py
```diff
@@ -522,7 +515,7 @@ def get_shell_commands(self):
                     environment.append("export {str(var)}")
         environment.append("")
         # Write the unset commands
-        if "unset_vars" in self.config:
+        if self.config.get("unset_vars", None) is not None:
```
pgierz:
```suggestion
        if self.config.get("unset_vars") is not None:
```

src/esm_runscripts/batch_system.py
```diff
@@ -144,7 +144,7 @@ def get_batch_header(config, cluster):
                     all_flags.append(flag)
 
         # some items in `all_values` list might be lists, so flatten it
-        all_values = [this_batch_system[flag] for flag in all_flags]
+        all_values = [this_batch_system.get(flag,None) for flag in all_flags]
```
pgierz:
```suggestion
        all_values = [this_batch_system.get(flag) for flag in all_flags]
```

src/esm_runscripts/batch_system.py
```diff
@@ -156,7 +156,8 @@ def get_batch_header(config, cluster):
         for value in all_values_flat:
             for (tag, repl) in replacement_tags:
                 value = value.replace(tag, str(repl))
-            header.append(this_batch_system["header_start"] + " " + value)
+            if this_batch_system.get("header_start",None) is not None:
```
pgierz:
```suggestion
            if this_batch_system.get("header_start"):
```

src/esm_tools/__init__.py
```diff
@@ -363,14 +363,14 @@ def copy_runscript_folder(dest_path):
 
 
 @caller_wrapper
-def get_namelist_filepath(namelist):
+def get_namelist_filepath(namelist=""):
```
pgierz, line 366: Why is this one needed in this function? (and below?

pgierz (changes requested):
Hi @seb-wahl,

looks good. I have some questions for the namelist part in `esm_tools/__init__.py`, and I had some suggestions to remove some redundant parts.

src/esm_tools/__init__.py
seb-wahl: Short answer: Because it works. Long answer: Because the underlying function that `get_namelist_path` wraps require the argument `nml` to be present, and I wanted to be as little invasive as possible, especially in this part of the parser.

src/esm_environment/esm_environment.py
seb-wahl: solved in latest commit

src/esm_environment/esm_environment.py
seb-wahl: solved in latest commit

src/esm_runscripts/batch_system.py
seb-wahl: solved in latest commit

src/esm_runscripts/batch_system.py
seb-wahl: solved in latest commit

seb-wahl:
Latest changes are here: https://github.com/esm-tools/esm_tools/pull/638/commits/5220a595533eae897e47634a6ab48b2a170dde40

src/esm_environment/esm_environment.py
```diff
@@ -86,8 +85,6 @@ def __init__(self, run_or_compile, complete_config=None, model=None):
             for model in complete_config:
                 self.apply_config_changes(run_or_compile, complete_config, model)
 
-        # Add the ENVIRONMENT_SET_BY_ESMTOOLS into the exports
-        self.add_esm_var()
```
mandresm: I have always assumed that some model might have used that. This introduces the following lines in the `.run` file:

```
export ENVIRONMENT_SET_BY_ESMTOOLS=TRUE
```

Does anyone know if that's the case? @pgierz @denizural

pgierz: ECHAM via auto-tools uses that. Does the `add_esm_var` do **only** the job of setting the variable? If yes, then I would propose to leave it in.

pgierz: To be more specific: PaleoDyn's ECHAM via Autotools uses that...The Geomar ECHAM might be different.

denizural: I have not seem anyone using `ENVIRONMENT_SET_BY_ESMTOOLS`. It is still nice as a debug feature though.

mandresm:
> Note to whoever merges this later: Please put in **both** #637 as well as this one, I just updated some documentation in the wrong branch.

I'm gonna merge that branch here. I rather review exhaustively one PR, plus this branch has advance much more and might contain important changes for the removal of the rcfile that are not present yet in `cleanup/remove_rcfile`

src/esm_environment/esm_environment.py
```diff
@@ -476,7 +470,7 @@ def get_shell_commands(self):
 
         environment = []
         # Write module actions
-        if "module_actions" in self.config:
+        if self.config.get("module_actions") is not None:
```
mandresm, line 473: Talking about style, this is the only place in the code where we have this syntax. Normally we would do:
```
if self.config.get("module_actions"):
```
because `get` if there is not second argument, returns `None` if the variable is not found, and the `if` condition is therefore not satisfied. Is there a specific reason why we want this syntax here?

pgierz: There is no specific reason. It may become relevant if you provide something like an empty list or empty dict, but for module actions (which should, if I remember correctly, always give back a list) it does not matter.

mandresm: Good point. So this will stay as it is now.

mandresm (commented):
All looks good to me. I am running `esm_tests` right now in Ollie. I'll give the thumbs up by tomorrow morning if everything worked fine.

My only concern is that the `EsmToolsDir` class was important for some steps in the virtual environment, so we might want it back if the virtual environment breaks. I'll be testing that too and bringing `EsmToolsDir` back (this time in `esm_runscripts`) only if needed.

mandresm (approved):
All tests are working for `esm_tests` in ollie (except for some problems in the pool with `awiesm2.1` with wiso, but that is not releated to this PR). Also I tried the virtual environment and works well, and it takes the namelists, config files, etc from the local venv as it should.

mandresm:
#bump #minor

PR #639, opened 2022-04-11
some fixes for esm_tests repos

mandresm:
Some fixes for the latest version of `esm_tests`

### DIFF
--- a/src/esm_tests/cli.py
+++ b/src/esm_tests/cli.py
@@ -25,6 +25,12 @@ def main():
         filter={"": "WARNING", "esm_tests": "DEBUG"},
         format="<level>{message}</level>",
     )
+
+    logger.info("")
+    logger.info("Welcome to ESM-Tests!")
+    logger.info("=====================")
+    logger.info("")
+
     if os.environ.get("CI", False):
         logger.add(
             "out.log",
@@ -48,6 +54,14 @@ def main():
         default=False,
         help="Check mode on (does not compile or run, but produces some files that can "
         + "be compared to previous existing files in 'last_tested' folder)",
+        action="store_true",
+    )
+    parser.add_argument(
+        "-u",
+        "--update",
+        default=False,
+        help="Updates the resources with the release branch, including runscripts"
+        + "and last_tested files",
         action="store_true",
     )
     parser.add_argument(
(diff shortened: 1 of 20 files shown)


src/esm_tests/repos.py
```diff
@@ -1,33 +1,56 @@
...
+        sh("git submodule init", verbose=verbose)
+        sh("git submodule update", verbose=verbose)
+        os.chdir(f"{info['script_dir']}/resources")
+
+        # Pull the desired branch of resources
+        try:
+            sh(f"git checkout {resources_branch}")
+            sh("git pull")
+        except:
+            logger.warning(f"Pull of {resoureces_branch} branch not possible")
```
denizural: I think this is more severe than a warning, right? Maybe `logger.error`. But then, a follow-up action should be taken. Eg. `sys.exit()`. I don't know about the details much though.
Also, be careful about the typo `resoureces_branch`

denizural: I always use `CTRL+n` in vim for autocomplete btw. It helps to prevent typos.

src/esm_tests/cli.py
```diff
@@ -25,6 +25,12 @@ def main():
         filter={"": "WARNING", "esm_tests": "DEBUG"},
         format="<level>{message}</level>",
     )
```
denizural, line 27: It might be a good idea to initialize the logging level. Eg `"INFO"` or `"DEBUG"` in `logger.add`. The default is DEBUG in loguru.
Eg.
```Python
logging_level = "INFO"
logger.add(..., level=logging_level, ...) 
```

denizural (changes requested):

src/esm_tests/repos.py
mandresm: thanks, I did not know that!

src/esm_tests/cli.py
mandresm: That it's done on the `cli.py` already.

mandresm: I keep the default format for now on levels that are not `"DEBUG"`. I won't add this now, as I am trying to speedup and I need this for testing levante.

denizural (approved):

mandresm:
#bump

PR #642, opened 2022-04-17
fix/additional flags list

mandresm:
all the aditional_flags variables in slurm.yaml, pbs.yaml and machine files turned into a list, so that add_additional_flags can be used from the runscript to expand the job manager flags

This will allow us in the workshop to add the following to the runscript and show how to use `additinal_flags`:
```
computer:
    add_additional_flags:
        - "--reservation=esmtools"
```


denizural (approved):
👍

mandresm:
#bump

PR #643, opened 2022-04-19
added support for parallel I/O in ECHAM6

seb-wahl:
has been used in FOCI for years and can significantly speed up simulations especially on HLRN4
To use this feature add
```yaml
echam:
   nprocio: 12 # or any other integer >= 4 (fewer processors for I/O don't work)
   namelist_changes:
      namelist.echam:
         parctl:
            iomode: 2
            nprocio: ${nprocio}
```
to your runscript.


mandresm (commented):
Wouldn't it be useful to add the following to the `echam.yaml`?
```
nprocio: 0
choose_nprocio:
   0:
       nprocio_option: False # An option cannot be empty, so a variable is needed here
   "*":
       add_namelist_changes:
          namelist.echam:
             parctl:
                iomode: 2
                nprocio: ${nprocio}
```

pgierz (approved):
Code change is feasible, but I would still like to include @mandresm's changes to `echam.yaml`

mandresm:
`esm_tests` are showing problems with this last commit

pgierz:
It gives back "cyclic dependency", but I do not understand why.

mandresm:
Solved in #781

PR #651, opened 2022-05-02
adds missing variable initializations

denizural:
The arrays `changes_list` and `add_list` are defined in the if block before but they are not reachable from the rest of the `check_changes_duplicates` function. This bugfix declares them at the top. Otherwise this causes the error
```python
UnboundLocalError: local variable 'changes_list' referenced before assignment
```


mandresm:
What are `add_list` and `changes_list` variables? Are those new functionalities?

mandresm (approved):

mandresm:
#bump

mandresm:
#bump

Issue #652, opened 2022-05-02
Virtual env on juwels 


JanStreffing:
I would like to make a longer production run on juwels. I got the following error message:

```
? Which of the following packages do you want to install in 'editable mode'? End with <Enter> done
Building virtual env, please be patient (this takes about 3 minutes)...
...finished 0:04:16.006684, restarting your job in the virtual env
Traceback (most recent call last):
  File "/p/scratch/chhb19/streffing1/runtime/awicm3-frontiers-xios/pi_controll_3-1-TCO95_CORE2/.venv_esmtools/bin/esm_runscripts", line 8, in <module>
    sys.exit(main())
  File "/p/scratch/chhb19/streffing1/runtime/awicm3-frontiers-xios/pi_controll_3-1-TCO95_CORE2/.venv_esmtools/lib/python3.9/site-packages/esm_runscripts/cli.py", line 264, in main
    Setup = SimulationSetup(command_line_config)
  File "/p/scratch/chhb19/streffing1/runtime/awicm3-frontiers-xios/pi_controll_3-1-TCO95_CORE2/.venv_esmtools/lib/python3.9/site-packages/esm_runscripts/sim_objects.py", line 42, in __init__
    self.config = prepare.run_job(self.config)
  File "/p/scratch/chhb19/streffing1/runtime/awicm3-frontiers-xios/pi_controll_3-1-TCO95_CORE2/.venv_esmtools/lib/python3.9/site-packages/esm_runscripts/prepare.py", line 13, in run_job
    helpers.evaluate(config, "prepare", "prepare_recipe")
  File "/p/scratch/chhb19/streffing1/runtime/awicm3-frontiers-xios/pi_controll_3-1-TCO95_CORE2/.venv_esmtools/lib/python3.9/site-packages/esm_runscripts/helpers.py", line 50, in evaluate
    framework_recipe = esm_plugin_manager.read_recipe(
  File "/p/scratch/chhb19/streffing1/runtime/awicm3-frontiers-xios/pi_controll_3-1-TCO95_CORE2/.venv_esmtools/lib/python3.9/site-packages/esm_plugin_manager/esm_plugin_manager.py", line 10, in read_recipe
    recipe = yaml_file_to_dict(recipe)
[...]

PR #653, opened 2022-05-02
Fix/slash in recipes4venv

mandresm:
Solves #652


pgierz (approved):
wait wait wait....a slash? seriously?

We need to think of an internal convention for this one. My opinion: every path we give as a variable **must** end in a slash, so you can just keep writing the rest of the path. If you give an extra slash, no one cares, if you do not, nothing breaks.

and before someone throws the book at me: yes this would all be better solved if we used `Pathlib`, but, let's stay pragmatic.

pgierz:
#bump

denizural:
Yes, exactly. I always use `Pathlib` in my personal projects since it uses objects and the `/` operator is overloaded. You don't need to worry about anything. 
Another thing that really annoys me is string concatenations for path constructions. These should be avoided. This is what I try to accomplish in the code cleaning phase. We need to come up with a coding style and stick to that.

PR #654, opened 2022-05-02
fix/esm_tests empty resources

mandresm:
- if the resources submodule for esm_tests (esm_tests_info) are not installed, asks the user for the installation
- and a bonus to fix the automatic bumping issues: changes in the pre-merge actions to use bumpversion while avoiding the unsafe repository issue

### DIFF
--- a/.github/workflows/pre-merge_actions.yml
+++ b/.github/workflows/pre-merge_actions.yml
@@ -38,7 +38,6 @@ jobs:
         run: |
           git config --global user.name "${GITHUB_ACTOR}";
           git config --global user.email "${GITHUB_ACTOR}@users.noreply.github.com"
-          git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
       # Checkout PR-branch (I could not find an easier way to do this and it seems
       # many people are struggling with the same problem when your event is triggered
@@ -76,7 +75,7 @@ jobs:
             )
           )
           }}
-        uses: jaumann/github-bumpversion-action@master
+        uses: jasonamyers/github-bumpversion-action@v1.0.4
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DEFAULT_BUMP: patch
@@ -98,7 +97,7 @@ jobs:
             )
           )
           }}
-        uses: jaumann/github-bumpversion-action@master
+        uses: jasonamyers/github-bumpversion-action@v1.0.4
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DEFAULT_BUMP: minor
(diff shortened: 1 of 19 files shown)


.github/workflows/pre-merge_actions.yml
```diff
@@ -38,7 +38,6 @@ jobs:
         run: |
           git config --global user.name "${GITHUB_ACTOR}";
           git config --global user.email "${GITHUB_ACTOR}@users.noreply.github.com"
-          git config --global --add safe.directory "$GITHUB_WORKSPACE"
```
pgierz: why is this taken out? what did it do in the first place?

src/esm_tests/repos.py
```diff
@@ -30,6 +35,33 @@ def update_resources_submodule(info, verbose=True):
...
 
+def check_resources(info, verbose=True):
+    resources_folder = f"{info['script_dir']}/resources/"
+    dir_resources = os.listdir(resources_folder)
+    if not (
+        "last_tested" in dir_resources
+        and "runscripts" in dir_resources
+        and not info.get("repo_update", False)
+    ):
+        install_resources = questionary.select(
```
pgierz: I think `questionary` has a True/False selector, so you might not need to look below for the explicit string `"Yes"`. That's small stuff though, if it works, leave it alone :-)

.github/workflows/pre-merge_actions.yml
```diff
@@ -76,7 +75,7 @@ jobs:
             )
           )
           }}
-        uses: jaumann/github-bumpversion-action@master
```
pgierz: There is not an official `bumpversion` action from GitHub, no? As long as everything works the same it really does not matter. It would only be nice to check if we rely on any feature (e.g. `env.DEFAULT_BUMP`) which may specifically rely on the subscribed action. 

I do however like that it is now tired to a version and not just something like `master` which might change over time 👍🏻

src/esm_tests/repos.py
```diff
@@ -1,14 +1,19 @@
 import os
+import questionary
 
 from loguru import logger
 
 from .test_utilities import sh
+from esm_parser import user_error
```
pgierz, line 7: We might consider moving this into `esm_utilities` as is it used often enough everywhere else. It isn't really a parsing thing.

pgierz (approved):
@mandresm, looks good from my side. I have a few minor comments, most of which are actually just stuff we can put on the cleanup list. Nice work!

.github/workflows/pre-merge_actions.yml
mandresm: That was introduced this morning by me when I was trying to solve the same problem. It's basically what git was suggesting to do in the error for the bumpversion, but adding that line did not work.

.github/workflows/pre-merge_actions.yml
mandresm: It's essentially the same action, but fixed by another developer.

src/esm_tests/repos.py
mandresm: Yeap, I agree, but that's a lot of changes in many places. I think this topic will come up again when we cleanup the parser, so we can leave it for that time?

pgierz: yes. in any case that will be a separate PR, and should not be done here.

src/esm_tests/repos.py
mandresm: solved :)

mandresm:
#bump

PR #664, opened 2022-05-10
Feat/levante

mandresm:
Testing that nothing breaks for Ollie but will finish soon.

For a list of what is supported check this https://github.com/esm-tools/esm_tools/discussions/663

Waiting on AWI-ESM-2.1 (will be soon added).

@tsemmler05 @vkolatschek @helgegoessling @mathanase


pgierz:
Note for AWI-ESM: standard compiling works; compiling with optimised flags suggested by DKRZ does not yet work. I would recommend handing this task off to DKRZ, putting this in and then performing an update once DKRZ gets back to us.

configs/components/fesom/fesom-2.0.yaml
```diff
@@ -110,7 +110,16 @@ namelists:
         - namelist.oce
         - namelist.ice
         - namelist.io
+
+# FIXME(KH, PG): This will need to be put in correctly.
+# NOTE(PG): For awiesm test on levante --> cvmix namelist back in, recom namelist out!
+# kh 21.02.22 orig (disabled for a special test)
```
pgierz, line 116: This still needs a general solution before it goes into the main branch. I would for the `AWIESM` side recommend something similar to `use_cvmix` as we already have for other "optional" features. However, `cvmix` capability depends on the version of `FESOM` being used (I think??), and I am not yet sure if the one we give with `AWIESM 2.1` has that.

configs/components/fesom/fesom-2.1.yaml
```diff
@@ -38,6 +38,75 @@ choose_version:
...
+        # kh 21.02.22
+        asforcing: CORE2
+        namelists:
+                - namelist.config
+                - namelist.forcing
+                - namelist.oce
+                - namelist.ice
+                - namelist.io
+        # kh 21.02.22 temporary disabled to get fesom-2.1-recom-par_tracers running on levante
+        #       - namelist.cvmix
```
pgierz: Same as above; we need here some generalised solution, or ideally a consolidated FESOM codebase.

configs/components/fesom/fesom-2.1.yaml
```diff
@@ -38,6 +38,75 @@ choose_version:
...
+        #       cvmix:   cvmix
+        #       icepack: icepack
+        # kh 21.02.22 these entries are required in combination with the corresponding entries in "config_sources:" to copy/use the defaults like namelist.recom, ...
+                recom:   recom
+                io.recom.ciso: io.recom.ciso
+        # kh 21.02.22 these entries are required in combination with the corresponding entriesin "config_files:" to copy/use the defaults like namelist.recom, ...
+        add_config_sources:
+                recom:   "${namelist_dir}/namelist.recom"
+                io.recom.ciso: "${namelist_dir}/namelist.io.recom.ciso"
+#-----------END OF the SPECIAL CHANGES FOR FESOM-PAR-TRACERS
```
pgierz, line 87: General idea: we should probably try to move towards putting comments and such in commit messages, and (maybe) not (or as well?) in the code. It makes it cluttered to read, in my opinion.

configs/components/oasis3mct/oasis3mct.env.yaml
```diff
@@ -1,5 +1,9 @@
-environment_changes:
+add_compiletime_environment_changes:
         choose_computer.fc:
                 ftn:
                         add_export_vars:
                                 - 'OASIS_FFLAGS=-emf'
+        choose_computer.name:
+                levante:
+                        add_export_vars:
+                                OASIS_FFLAGS: '"-march=core-avx2 -mtune=core-avx2"'
```
pgierz, line 9: These flags are then passed to `cmake` somehow, right?

configs/machines/generic.yaml
```diff
@@ -3,6 +3,8 @@
 name: generic
 accounting: false
 
+
+hetjob_flag: "?" # either hetjob or packjob
```
pgierz, line 7: We should consider turning the `generic.yaml` file into something we can include in the documentation page with comments that show what you would need to do to add a new machine.

As things such as `hetjob` may depend on the scheduler version, maybe we can also include a variable for `batch_system_version`. It will probably never be actually used, but it might be good as a place to look as a reminder.

namelists/fesom2/2.1-recom-par_tracers/namelist.config
```diff
@@ -0,0 +1,95 @@
+! This is the namelist file for model general configuration
```
pgierz, line 1: What is the consensus here: do we ship namelists? Or do we simply refer to the ones given to us by the model developers? My idea would be:

* Models that ship example namelists with their source code use those. Then it is up to the model developers or maintainers to keep those working.
* Models that do not ship example namelists with their source code, we shall provide example namelists that we know work in conjunction with our example run configurations.

pgierz (changes requested):
My only thing that still should be changed is finding a generalised way to handle some of @khimstedt's changes he needed to make before this goes out to users.

I have some other small ideas as well, let's please discuss about those.

configs/components/fesom/fesom-2.0.yaml
pgierz: It seems the version given to you with `awiesm-2.1` does indeed include cvmix. https://github.com/FESOM/fesom2/tree/2.1-awiesm-2.1/src

namelists/fesom2/2.1-recom-par_tracers/namelist.config
pgierz: See also: https://github.com/esm-tools/esm_tools/discussions/669

pgierz:
For the FESOM part of AWI-ESM, please also see here: https://github.com/FESOM/fesom2/pull/345

That would provide optimised compiletime flags.

configs/components/fesom/fesom-2.1.yaml
khimstedt: For a general solution of the namelist reference problem I see two possibilities:
a) A consolidated FESOM code base (as already suggested by Paul).

b) A support for adding/removing namelist references via an appropriate data structure, replacing the current commenting in or out of the namelist references in the yaml code. If, for example, a base namelist is maintained that initially contains all possible namelist references, the appropriate deletion of unneeded references would be sufficient for a concrete FESOM variant. If necessary, more sophisticated tree-like structures would also be conceivable, so that common namelist references of similar FESOM variants could also be factored out (like in a typical object-oriented class design).

Variant b) probably offers more potential for future similar cases but variant a) seems better maintainable and more stable.

configs/components/fesom/fesom-2.1.yaml
khimstedt: After the problem of the namelist references in general way is solved, also the still existing comments in the yaml code become obsolete. Until then, I find it quite helpful - in a feature branch - to have them available directly in the yaml code (without the need to use git diff or similar). Partly the comments also reflect (after a previous analysis) how the ESM Tools "work". In principle, this can also be done through the documentation, but having it at hand in the exact applicable place is sometimes also helpful.

namelists/fesom2/2.1-recom-par_tracers/namelist.config
khimstedt: I totally agree with Paul's suggestion how to ship the namelists. The goal should be to be able to perform representative simulation runs "out of the box" for each (coupled) model.

configs/components/fesom/fesom-2.1.yaml
pgierz: Making the online documentation more clear was something we also discussed during our retreat. That comment was just here as a reminder to myself that we should not forget about it.

pgierz:
AWIESM 2.1 Compiling with optimised flags for Levante should now work for ECHAM6 and OASIS. The above issue for FESOM is still outstanding, but FESOM compiles with default, non-optimized flags.

A runtime test is underway, I’ll report back later on. It would be great if this could go in before the paleodyn retreat on May 30/Jun 1

pgierz:
AWIESM 2.1 now also runs, I had to put `$HD5ROOT/lib` earlier in the `LD_LIBRARY_PATH` to avoid a shared library not being found. I am unsure if this affects other models.

This is now ready to be put into release, from the AWIESM perspective. Note I have tested only the aso version, so, no isotopes and no pico. However, if all the input and forcing files would be in place, I would not foresee any challenges there.

configs/components/fesom/fesom-2.0.yaml
mandresm: AWIESM uses `fesom-2.1.yaml`, not `fesom-2.0.yaml`. I'll get rid of these comments.

configs/components/fesom/fesom-2.1.yaml
mandresm: Option b already exists. I'll commit a change using the `remove` feature.

mandresm: This is how it looks now and how one can use `remove_` to remove elements of dictionaries and lists:

https://github.com/esm-tools/esm_tools/blob/e0d47a21bbab3c5ed8653c871be2e195bc046db9/configs/components/fesom/fesom-2.1.yaml#L68-L78

configs/components/fesom/fesom-2.1.yaml
mandresm: I agree with both, in my opinion, as this block will disappear in the future, we can leave the comments there as they are useful to @khimstedt

configs/components/oasis3mct/oasis3mct.env.yaml
mandresm: Some versions of OASIS have that on their cmake files, others not. This was written here for the one from AWICM3 which can handle this. I am not sure about the one for AWIESM2

configs/machines/generic.yaml
mandresm: I think this is better handled manually. Adding another choose for something that only changes effectively between computers seems a bit of an overkill to me.

namelists/fesom2/2.1-recom-par_tracers/namelist.config
mandresm: For FESOM namelists, we had a discussion about this some time ago with @koldunovn and other FESOM users. The conclusion was that for the `master` branch (which changes through time) the namelists should be taken from the source code, while for stable branches/tags (`2.0`, `2.1`, ...) the namelists will be taken from ESM-Tools.

I don't have an opinion on what should be the general approach, as both can trigger an error in the namelist that ends in our discussion/mail-box (the code moves on but ESM-Tools namelists don't/someone changes the namelists in the FESOM source code and what was once working in ESM-Tools is not working anymore, or even worse, it works but the physics are different)

For the particular case of the par_tracers I agree with @khimstedt, that they can be taken from the source code as very advance users are using this development branch, so I have adjusted that accordingly in 8278f0b8

configs/components/fesom/fesom-2.1.yaml
```diff
@@ -38,6 +38,54 @@ choose_version:
...
+#       - recom
+# YY: 29.04.22 to compile wiso code
+    wiso_code: true
+
+# MA: changes specific to PAR-TRACERS go here so they don't destroy other working versions of FESOM-2.1)
+# TODO: remove this block when PAR-TRACERS is generalized!
+par_tracers: "$(( 'par_tracers' in  '${fesom.version}' ))"
+choose_fesom.par_tracers:
+    True:
+        namelist_dir: "${model_dir}/config/"
```
pgierz, line 63: This will also override other namelists (e.g. `namelist.oce` or `namelist.config`) which will then be taken from the `${model_dir}/config/`. That follows what the poll voted for, so I am ok with that, but we need to ensure the namelists given there work (or just tell the FESOM guys)

configs/machines/generic.yaml
pgierz: I meant here more as documentation for ourselves; not to put the `hetjob_flag` behind a choose.

pgierz (approved):
If AWICM1 and AWICM3 still compile, I'm OK with this.

It does look like the hetjob/packjob feature got mixed in with this one. Was that on purpose, or just a bad rebase/cross-merge? I don't particularly care, but it would make reverting this PR commit a bit more handwork if we ever need to go backwards.

mandresm:
> If AWICM1 and AWICM3 still compile, I'm OK with this.
> 
> It does look like the hetjob/packjob feature got mixed in with this one. Was that on purpose, or just a bad rebase/cross-merge? I don't particularly care, but it would make reverting this PR commit a bit more handwork if we ever need to go backwards.

It was in purpose. But you are right... bad practice... I'll make a good mental note about it.

mandresm:
#bump and crossing fingers that the auto-bumping action still works...

mandresm:
So it still works!

Issue #384, opened 2021-07-06
need reduced echam and jsbach output for spinup


chrisdane:
Hi

**Is your feature request related to a problem? Please describe.**
The default `namelist.echam` for a `PI-CTRL` experiment, which is (almost always?) also used for spinups, creates a lot of data on a high temporal interval (< month) due to the namelist blocks
```C
# cd to esm_tools
git checkout develop
git fetch
# namelists/echam/6.3.04p1/PI-CTRL/namelist.echam:
&mvstreamctl
    interval = 1, 'days', 'last', 0
    target = 'glday'
    source = 'gl'
    variables = 'q:mean'
/
&mvstreamctl
    target = 'g3bday'
    interval = 1, 'days', 'last', 0
    source = 'g3b'
    meannam = 'tslm1', 'tsi'
/
&mvstreamctl
    interval = 1, 'days', 'last', 0
    target = 'g3bid'
    source = 'g3b'
    variables = 'u10:mean', 'v10:mean', 'temp2:mean', 'relhum:mean', 'albedo:mean',
                'dew2:mean', 'ws:mean', 'sn:mean', 'wimax:max', 't2min:min', 't2max:max'
/
&mvstreamctl
    interval = 1, 'days', 'last', 0
    target = 'jsbid'
    source = 'jsbach'
    variables = 'layer_moisture:mean'
/
&mvstreamctl
    interval = 6, 'hours', 'last', 0
    target = 'sp6h'
    source = 'sp'
    variables = 'st:mean', 'svo:mean', 'lsp:mean', 'sd:mean'
/
&mvstreamctl
    interval = 1, 'hours', 'last', 0
    target = 'g3b1hi'
    source = 'g3b'
    variables = 'u10:inst', 'v10:inst', 'wimax:max'
/
```
For a spinup, this is unnecessary and bad practice since nobody will ever need this data but the disks are full with it.

**Describe the solution you'd like**
As far as I understand the esm tools, I would like to have a `esm_tools/namelists/echam/<version>/PI-CTRL-SPINUP/namelist.echam` which is the same as  `esm_tools/namelists/echam/<version>/PI-CTRL/namelist.echam` but with only a few important variables on monthly output frequency, or similar.

I tried to achieve this. I ran a default echam-only `PI-CTRL` experiment on ollie:
```
esm_tools branch: develop
runscript: /home/ollie/cdanek/esm/runscripts/echam-ollie-initial-monthly.yaml 
[...]

PR #665, opened 2022-05-10
Feat/ssp585 awicm recom v6 with release

denizural:
Hi folks,
This PR will add `awicm-recom`. It already contains one of the recent `release` of `ESM-Tools`.
@seb-wahl, could you please have a look if the ECHAM changes clash with your FOCI setups?

Some features:
- default runscripts for HIST and PI-CTRL
- Runs for cold starts and restarts
- Output configurations (eg. Fesom) are moved out of the basic configuration file to enhance readability.
- Some minor clean code practices with YAML and Python codes (eg. better variable names, consistent indentations)

NOTE:
- `fesom-recom` may not run at this point but I will solve this later on.


chrisdane (approved):

configs/components/echam/echam.datasets.yaml
```diff
@@ -30,93 +35,159 @@ forcing_sources:
...
+                        from: 2015
+        
         # greenhouse
-        "histgreenh": "${greenhouse_dir}/greenhouse_historical.nc"
-        "greenhouse_hist_scenario":
+        "histgreenh": 
                 "${greenhouse_dir}/greenhouse_historical.nc":
+                        from: 1
+                        to: 2013
+                "/work/ba1103/a270073/data/ECHAM6/input/r0007/greenhouse_historical.nc": # see README.md in this dir
```
mandresm: Can we make this path to not be hard coded? (make a dir variable and include in a choose_block of computer.machine)?

pgierz:
I can't give a green light for this until some explains to me what changed here:
```
awiesm:
[5220](https://github.com/esm-tools/esm_tools/runs/6978790186?check_suite_focus=true#step:6:5221)
|   2.1:
[5221](https://github.com/esm-tools/esm_tools/runs/6978790186?check_suite_focus=true#step:6:5222)
|       bootstrap:
[5222](https://github.com/esm-tools/esm_tools/runs/6978790186?check_suite_focus=true#step:6:5223)
|           ollie:	comp files identical	run files differ
[5223](https://github.com/esm-tools/esm_tools/runs/6978790186?check_suite_focus=true#step:6:5224)
|       awiesm-2.1_icebergs:
[5224](https://github.com/esm-tools/esm_tools/runs/6978790186?check_suite_focus=true#step:6:5225)
|           ollie:	comp files identical	run files differ
[5225](https://github.com/esm-tools/esm_tools/runs/6978790186?check_suite_focus=true#step:6:5226)
|       pi:
[5226](https://github.com/esm-tools/esm_tools/runs/6978790186?check_suite_focus=true#step:6:5227)
|           ollie:	comp files identical	run files differ
[5227](https://github.com/esm-tools/esm_tools/runs/6978790186?check_suite_focus=true#step:6:5228)
|       pico:
[5228](https://github.com/esm-tools/esm_tools/runs/6978790186?check_suite_focus=true#step:6:5229)
|           ollie:	comp files identical	run files differ
```
If something modifies AWI-ESM benchmarks I get the hair on the back of my neck standing up a little bit. 


And an important note: smaller changes. I don't want to review thousands of lines to see what is happening. I would prefer several small PRs instead of one giant one, then I can better wrap my head around what was fixed or changed. Based on the PR summary this does 4 things at once.

configs/components/echam/echam.datasets.yaml
```diff
@@ -30,93 +35,159 @@ forcing_sources:
...
+                        from: 2014
                         to: 2014
-                "${greenhouse_dir}/greenhouse_${scenario}.nc":
-                        from: 2015
-
```
mandresm: @seb-wahl , @pgierz , @chrisdane , I can understand the syntax but I don't understand what are the file rules for the models that you are using/developing, so can you all carefully check that all this is correct for your models (or is not used)? I think this is the most "dangerous" part of this PR.

configs/components/echam/echam.yaml
```diff
@@ -385,23 +370,80 @@ choose_scenario:
...
+                        aerofarir: piaerofarir
+                        aerofin: rcp45aerofin
+                        ozone: rcp45ozone
+                        greenhouse: rcp45greenh
+                        volcir: histvolcir
+                        volcsw: histvolcsw
+                        swflux: histswflux
+                add_input_files:
+                        MAC-SP: MAC-SP
+        RCP85:
```
mandresm: Killing the `scenario_type` seems like a bad idea. This was implemented 11 months ago, I believe by @seb-wahl in agreement with other ECHAM developers. It is a shame we are working with such old stuff @denizural

configs/components/echam/echam.yaml
```diff
@@ -149,7 +148,7 @@ input_dir: "${pool_dir}/input/${dataset}"
 adj_input_dir: "${input_dir}/${resolution}"
 forcing_dir: "${input_dir}/${resolution}"
 greenhouse_dir: "${pool_dir}"
-namelist_dir: "${general.esm_namelist_dir}/echam/${version}/${scenario_type}"
+namelist_dir: "${general.esm_namelist_dir}/echam/${version}/${scenario}"
```
mandresm: Same as below for `scenario_type`...

configs/components/echam/echam.yaml
```diff
@@ -596,24 +638,39 @@ forcing_additional_information:
         aerocoarse:
                 - need_year_before
                 - need_year_after
+                - need_2years_before
+                - need_2years_after
```
mandresm, line 583: does this break anything for the other models?

configs/components/echam/echam.yaml
```diff
@@ -675,15 +743,12 @@ outdata_sources:
...
-        #"[[streams-->STREAM]]": ${general.expid}_${start_date!syear}*_STREAM
-        #"[[streams-->STREAM]]_codes": ${general.expid}_${start_date!syear}*_STREAM.codes
-        #"[[streamsnc-->STREAM]]_nc": ${general.expid}_${start_date!syear!smonth}_STREAM.nc
 
 restart_in_files:
         "[[streams-->STREAM]]": STREAM
 
 restart_in_sources:
-        "[[streams-->STREAM]]": restart_${parent_expid}_STREAM_${parent_date!syear!smonth!sday}.nc
+        "[[streams-->STREAM]]": restart_${parent_expid}_${parent_date!syear!smonth!sday}_STREAM.nc
```
mandresm: this will break the copying of the restarts of the rest of the models using ECHAM

configs/components/echam/echam.yaml
```diff
@@ -693,7 +758,7 @@ restart_out_files:
         "[[streams-->STREAM]]": STREAM
 
 restart_out_sources:
-        "[[streams-->STREAM]]": restart_${general.expid}_STREAM_${other_date!syear!smonth!sday}.nc
+        "[[streams-->STREAM]]": restart_${general.expid}_${other_date!syear!smonth!sday}_STREAM.nc
```
mandresm: this will break the copying of the restarts of the rest of the models using ECHAM

configs/components/echam/echam.datasets.yaml
```diff
@@ -30,93 +35,159 @@ forcing_sources:
...
+                                "${forcing_dir}/aero/${resolution}_aeropt_kinne_sw_b14_fin_@YEAR@.nc":
+                                        from: 1865
+                                        to: 2000
+                                "${forcing_dir}/aero/${resolution}_aeropt_kinne_sw_b14_fin_rcp85_@YEAR@.nc":
+                                        from: 2001
+                                        to: 2014
+                        # greenhouse
+                        "rcp26greenh": "${greenhouse_dir}/greenhouse_rcp26.nc"
+                        "rcp45greenh": "${greenhouse_dir}/greenhouse_rcp45.nc"
+                        "rcp85greenh": "${greenhouse_dir}/greenhouse_rcp85.nc"
```
mandresm: same as above, what is the impact in other models?

configs/components/echam/echam.yaml
```diff
@@ -342,22 +341,8 @@ choose_scenario:
...
-                        #aerofarir_1849: piaerofarir
-                        #aerofarir_1850: piaerofarir
-                        #aerofarir_1851: piaerofarir
-                        #aerofin_1849: piaerofin
-                        #aerofin_1850: piaerofin
-                        #aerofin_1851: piaerofin
-                        #
-                        # with iaero = 3 use te following
+        1950:
+                forcing_filslurm_cronjob.loges:
```
mandresm: What does this syntax do?

configs/components/fesom/fesom.yaml
```diff
@@ -261,26 +260,45 @@ config_sources:
...
+daily_outputs: []
+12_step_outputs: []
 
-yearly_outputs: [lwrd,olat,olwout,osen,relax_salt,shum,tair,uwind,vwind]
-
-monthly_outputs: [evs,fsitherm,hfds,opottemptend,pbo,prlq,prsn,rsdo,siarean,siareas,sidmassevapsubl,sidmasssi,sidmassth,sidmasstranx,sidmasstrany,siextentn,siextents,sifllatstop,sisnconc,sisnmass,sisnthick,sispeed,sivol,sivoln,sivols,soga,thetaoga,u2o,uo,uso,uto,v2o,vo,volo,vso,vto,w2o,wfo,wo,wso,wto,zossq,rho,uhice,uhsnow,urho,uv,vhice,vhsnow,virtual_salt,vrho]
-
-daily_outputs: [mlotst,omldamax,sic,sistrxdtop,sistrxubot,sistrydtop,sistryubot,sithick,sitimefrac,siu,siv,so,sos,tauuo,tauvo,thetao,tos,zos,flice,wnet,evap,runoff,thdgr,thdgrsn]
-
-12_step_outputs: [tso]
```
mandresm: This will affect `AWICM1` `CMIP6` simulations been ran by for example Vanessa, we need to check whether she is using her own output from the runscript (or additional file) or if she is using the defaults here, before merging this

configs/components/fesom/fesom.yaml
```diff
@@ -261,26 +260,45 @@ config_sources:
...
-        gf: gf
-        chloro: chloro
-
-ignore_in_work:
-        gf: goodfile
-        chloro: Chl_Sweeney.dat
-
-ignore_sources:
-        gf: goodfile
-        chloro: Chl_Sweeney.dat
```
mandresm: Why is this deleted?

configs/components/fesom/fesom.yaml
```diff
@@ -297,6 +315,18 @@ outdata_sources:
...
+        chloro: chloro
+
+ignore_in_work:
+        gf: goodfile
+        chloro: Chl_Sweeney.dat
+
+ignore_sources:
+        gf: goodfile
+        chloro: Chl_Sweeney.dat
+
```
mandresm: why is this moved?

configs/components/fesom/fesom.yaml
```diff
@@ -261,26 +260,45 @@ config_sources:
...
+                                        unit: m
+                                        rate: 1
+                                -
+                                        vars: ${daily_outputs}
+                                        unit: d
+                                        rate: 1
+                                -
+                                        vars: ${12_step_outputs}
+                                        unit: s
+                                        rate: 12
```
mandresm: why is this moved? Makes the code difficult to compare with previous versions

configs/components/fesom/fesom.yaml
```diff
@@ -321,51 +351,15 @@ choose_computer.name:
...
-                                -
-                                        vars: ${yearly_outputs}
-                                        unit: y
-                                        first: 1000
-                                        rate: 1
-
-
-
-
-
```
mandresm, line 354: why is this moved? Makes the code difficult to compare with previous versions. And there are some differences so I suggest it get's move back to where it was so the differences become clear.

configs/components/hdmodel/hdmodel.yaml
```diff
@@ -35,50 +35,30 @@ input_sources:
...
-                add_restart_out_files:
-                        hdrestart_wiso: hdrestart_wiso
-                restart_in_sources:
-                        hdrestart: restart_${parent_expid}_${parent_date!syear!smonth!sday!shour!sminute!ssecond}_hdrestart.nc
-                        hdrestart_wiso: restart_${parent_expid}_${parent_date!syear!smonth!sday!shour!sminute!ssecond}_hdrestart_wiso.nc
-                restart_out_sources:
-                        hdrestart: restart_${general.expid}_${other_date!syear!smonth!sday!shour!sminute!ssecond}_hdrestart.nc
-                        hdrestart_wiso: restart_${general.expid}_${other_date!syear!smonth!sday!shour!sminute!ssecond}_hdrestart_wiso.nc
-
-
```
mandresm: don't delete this

configs/components/hdmodel/hdmodel.yaml
```diff
@@ -35,50 +35,30 @@ input_sources:
...
 restart_in_in_work:
         hdrestart: hdrestart.nc
-        hdrestart_wiso: hdrestart_wiso.nc
```
mandresm, line 55: don't delete this

configs/components/hdmodel/hdmodel.yaml
```diff
@@ -35,50 +35,30 @@ input_sources:
...
 
 restart_in_files:
         hdrestart: hdrestart
 
-restart_out_files:
-        hdrestart: hdrestart
-
 restart_out_in_work:
         hdrestart: hdrestart.nc
-        hdrestart_wiso: hdrestart_wiso.nc
```
mandresm: don't delete this

configs/components/jsbach/jsbach.yaml
```diff
@@ -21,10 +21,8 @@ cover_fract_dir: "${input_dir}/${echam.resolution}/land_use_maps/"
 
 dynamic_vegetations: False
 
-
 version: "3.20"
 dataset: r0009
-scenario_type: ${echam.scenario}
```
mandresm: This file has the same problems as `echam.yaml` and that means too many differences and a surely break of all models using ECHAM. This needs to be solved before merging.

configs/components/recom/recom.yaml
```diff
@@ -1,40 +1,40 @@
...
-    branch: recom-update
-  2.0-recom-awicm:
-    branch: co2_coupling
+    '2.0':
+        branch: oxygen
+    2.0-r:
+        # OG: This will be the master branch for fesom-recom-2.0. We rename it later.
+        branch: recom-update
+    2.0-recom-awicm:
+        branch: oxygen
```
mandresm, line 85: Is this the correct branch for AWICM-1.4-REcoM @chrisdane?

configs/components/recom/recom.yaml
```diff
@@ -45,123 +45,119 @@ scenario: "preindustrial"
 
 time_step: 1800
 
-bgc_num: 22
-
-
```
mandresm, line 50: don't remove this, it is used by AWIESM-2.1-REcoM

configs/components/recom/recom.yaml
```diff
@@ -45,123 +45,119 @@ scenario: "preindustrial"
...
+        REcoM_PI: true
+        CO2_for_spinup: "0"
+        firstyearoffesomcycle: "${f_year_fesom_cycle}"
+        lastyearoffesomcycle: "${l_year_fesom_cycle}"
+        numofCO2cycles: "${num_CO2_cycles}"
+        currentCO2cycle: "${current_CO2_cycle}"
+
+
+#namelist_dir: "${esm_namelist_dir}/${model}"
+namelist_dir: "${general.model_dir}/recom/"
```
mandresm: please revert this, it might impact AWIESM2-REcoM

configs/components/recom/recom.yaml
```diff
@@ -45,123 +45,119 @@ scenario: "preindustrial"
...
+        pavariables:
+            REcoMDataPath: "${data_path}"
+            constant_CO2: "${constant_CO2}"
+            firstyearoffesomcycle: "${firstyearoffesomcycle}"
+            lastyearoffesomcycle: "${lastyearoffesomcycle}"
+            numofCO2cycles: "${numofCO2cycles}"
+            currentCO2cycle: "${currentCO2cycle}"
+            REcoM_PI: "${REcoM_PI}"
+        paco2_flux_param:
+            CO2_for_spinup: "${CO2_for_spinup}"
```
mandresm: bring back the `bgc_num` variable as it is used by AWIESM2-REcoM

configs/components/recom/recom.yaml
```diff
@@ -45,123 +45,119 @@ scenario: "preindustrial"
...
-        "[[monthly_outputs-->OUTPUT]]": OUTPUT
-        "[[daily_outputs-->OUTPUT]]": OUTPUT
-outdata_in_work:
-        "[[yearly_outputs-->OUTPUT]]": OUTPUT_fesom_${start_date!syear!smonth!sday}.nc
-        "[[monthly_outputs-->OUTPUT]]": OUTPUT_fesom_${start_date!syear!smonth!sday}.nc
-        "[[daily_outputs-->OUTPUT]]": OUTPUT_fesom_${start_date!syear!smonth!sday}.nc
-outdata_sources:
-        "[[yearly_outputs-->OUTPUT]]": OUTPUT_fesom_${start_date!syear!smonth!sday}.nc
-        "[[monthly_outputs-->OUTPUT]]": OUTPUT_fesom_${start_date!syear!smonth!sday}.nc
-        "[[daily_outputs-->OUTPUT]]": OUTPUT_fesom_${start_date!syear!smonth!sday}.nc
```
mandresm: This are REcoM output aren't they? Then they should be here so that they endup in the `output/recom` folder and not in the fesom one. One could easily add the `monthly_outputs` with a switch in the AWICM-REcoM yaml.

configs/machines/mistral.yaml
```diff
@@ -101,6 +101,7 @@ module_actions:
         - "load nco"
         - "load cdo"
         - "load gcc/4.8.2"
+        - "list"
```
mandresm: Is this needed? I think in some point we decided to remove it.

configs/setups/fesom-recom/fesom-recom.yaml
```diff
@@ -57,8 +57,6 @@ fesom:
         output_unit: "${restart_unit}"
         leapyear: false
 
-        yearly_outputs_start: 1     # Year in which the yearly output starts to be saved
-
```
mandresm, line 61: This file needs a thorough review

namelists/recom/fesom-recom/namelist.recom
```diff
@@ -2,32 +2,32 @@ This is the namelist recom
 
 &pavariables
 use_REcoM             =.true.
-REcoM_restart         =.false.
```
mandresm, line 5: Check that this namelist is the correct one for `fesom-recom`

src/esm_runscripts/filelists.py
```diff
@@ -592,7 +594,10 @@ def replace_year_placeholder(config):
                             # corresponding year
                             for year in all_years:
                                 new_category = file_category + "_year_" + str(year)
-
+                                # Kinne Aerosol files are handled out of this time loop
+                                if "@YEAR_1850@" in config[model][filetype + "_targets"][file_category]:
+                                    continue
+
```
mandresm: This looks dangerous. Can someone explain it?

mandresm (changes requested):
This PR, as it is, will very likely break all the models running with ECHAM, if not more. It needs a lot of work and considerations, and we need to reduce to the minimum the differences with `release`, specially there where there is duplication in functionality/configurations. I would also strongly recommend removing the `awicm_recom.yaml` and including it inside the `awicm.yaml` with switches, in a similar manner as it is done in `AWIESM-2.1-REcoM`.

configs/components/echam/echam.yaml
chrisdane: Everlasting filename issue. Then please adapt that it works for all setups which include the echam component.

configs/components/echam/echam.datasets.yaml
chrisdane: As far as I remember I did not delete those. MAC-SP and ozone files are defined for both piControl and specific scenarios. So its ok.

configs/components/echam/echam.yaml
chrisdane: Thats a typo I guess. Should be `forcing_files`.

configs/components/fesom/fesom.yaml
chrisdane: I informed Vanessa and asked for a comment. I think its important to reduce the default output and not save daily and even 12-step output to disk. Its just a waste of disk space. It is very easy to define high frequency output in the runscript via `further_reading`.

configs/components/fesom/fesom.yaml
chrisdane: This block was between the `*_outputs` and the `outdata_*` definitions. These two should be together so I moved it.

configs/components/fesom/fesom.yaml
chrisdane: Same here: I find it easier to follow if code chunks which belong together also stand together ^^

configs/components/fesom/fesom.yaml
chrisdane: I just ordered the blocks along the output frequency and replaced the hardcoded `1000`.

configs/components/hdmodel/hdmodel.yaml
chrisdane: I did not do this o_O

configs/components/hdmodel/hdmodel.yaml
chrisdane: same

configs/components/hdmodel/hdmodel.yaml
chrisdane: same

configs/components/recom/recom.yaml
chrisdane: Yes

configs/components/recom/recom.yaml
chrisdane: They define it in their awiesm.yaml so it should be save to delete?:
https://github.com/esm-tools/esm_tools/blob/0e65d3b2587eb18628845a64a9053dee622571cc/configs/setups/awiesm/awiesm.yaml#L340-L344

Explanation: Unfortunately fesom1 and recom are not coded so that an arbitrary number can be used for `bgc_num`. So by default, the values coming from the `namelist.config`s from the recom repo should be used.

configs/components/recom/recom.yaml
chrisdane: ... thats why we changed the default namelist directory to the one shipped with the recom repo. Sorry for this. Do you see a compromise?

configs/components/recom/recom.yaml
chrisdane: same

configs/components/recom/recom.yaml
chrisdane: I tried this. To make it clean I think we need the merge-lists-functionality (#672) because in the end all (fesom and recom) variables need to get exported to the environment variable `AWI_FESOM_YAML`. The drawback is that all (fesom and recom) output will be moved to `outdata/fesom` but this is ok. Currently, all recom output is saved twice! (in outdata/fesom and outdata/recom).

configs/components/fesom/fesom.yaml
vkolatschek: Yes, it won't effect my simulations. I'm already using the `further_reading` option in the runscript.

mandresm:
Branches info:

**AWICM-REcoM**
REcoM: `oxygen`
FESOM: `fesom_recom_1.4_master`
ECHAM: `co2_coupling`

**FESOM-REcoM**
REcoM: `oxygen`
FESOM: `master`

configs/machines/mistral.yaml
chrisdane: Why is it then in other machine.yamls, e.g. ollie? I think more info always helps but its not important.

configs/setups/fesom-recom/fesom-recom.yaml
chrisdane: Currently (release), the yearly variables are saved only in the 1000th model year:
https://github.com/esm-tools/esm_tools/blob/0e65d3b2587eb18628845a64a9053dee622571cc/configs/components/fesom/fesom.yaml#L346-L349

That does not make sense? I understand that variables that you do not need but must be saved according to the output scheduler are put in this block to save hard disk. However, what do you do if you actually want yearly output?

And btw, in my version this line is not deleted.

namelists/recom/fesom-recom/namelist.recom
chrisdane: This line is not deleted in my version oO

configs/components/echam/echam.yaml
chrisdane: I dont think so since Echam needs some forcing files +/- 2 years.

configs/components/echam/echam.yaml
chrisdane: same

configs/components/echam/echam.datasets.yaml
chrisdane: Found a hopefully better solution in a1518574 of (the temporary `cdanek-awicmrecom-wip` branch).

configs/components/echam/echam.datasets.yaml
chrisdane: Good to go from my side.

configs/components/jsbach/jsbach.yaml
chrisdane: In b0f75dad (the temporary `cdanek-awicmrecom-wip` branch) I removed the hard-coded personal path for the Ndepo and population density forcing files. Certainly not an all-time solution but at least for levante a workaround. 

Cannot say anything to `scenario_type`. Can't remember that I deleted this.

configs/components/fesom/fesom.yaml
chrisdane: see below

configs/components/echam/echam.yaml
mandresm: Hi Chris, I am not super sure I understand your answer, but I have a question: in ECHAM, what conditions the restart name format? Is it defined in the namelist? does it depend on the stream? it's been always very confusing to me that it changes from one couple setup to another

configs/machines/mistral.yaml
denizural: I think little verbosity does not hurt. It will help the users to see which modules are loaded.

configs/components/hdmodel/hdmodel.yaml
denizural: Resolved by 61f91684. I don't know how things got deleted. I normally only comment out the things I don't use.

configs/components/jsbach/jsbach.yaml
```diff
@@ -484,8 +484,8 @@ outdata_files:
         "[[streams-->STREAM]]_codes": STREAM_codes
 
 outdata_sources:
-        "[[streams-->STREAM]]": ${general.expid}_${start_date!syear!smonth}.${start_date!sday}_STREAM
-        "[[streams-->STREAM]]_codes": ${general.expid}_${start_date!syear!smonth}.${start_date!sday}_STREAM.codes
+        "[[streams-->STREAM]]": ${general.expid}_${start_date!syear}*.${start_date!sday}_STREAM
+        "[[streams-->STREAM]]_codes": ${general.expid}_${start_date!syear}*.${start_date!sday}_STREAM.codes
```
mandresm: I'm curious about this change @denizural, I'm totally fine with it, but can you elaborate on why it was necessary?

configs/components/echam/echam.yaml
```diff
@@ -548,6 +593,7 @@ choose_wiso_code:
                                 add_streams:
                                         - wiso
                                         - accw_wiso
+                                        # TODO: these two are missing in release
                                         - sf_wiso
                                         - rad
```
mandresm: And that is as it should be. It's one of the latest fixes. It will be solved when merging from release into this branch.

configs/components/echam/echam.yaml
chrisdane: Afaik its hardcoded in echam. It simply needs the right fileformat. Due to different esm_tools versions in the last years, things got confused. For example if someone wanted to restart from an older or non-esm_tools simulation, filename years needed to get adjusted.
So please discard this specific line of code if its not the current esm_tools standard.

mandresm: Okay, so I will revert this line to what it's in release. If you ever need to restart from a different format, adding this to your runscript will do the trick, while still keeping the most up-to-date conventions inside the new restarted experiment:
```
echam:
    choose_run.number:
        1:
            restart_in_sources:
                "[[streams-->STREAM]]": the_old_format_for_the_restart  
```
Not that this will only be done for run 1 :)

mandresm:
EDIT: **RESOLVED**

I think the following lines in the `recom.yaml` can be simplified quite a bit:
https://github.com/esm-tools/esm_tools/blob/a13326a8ef3c0096b1a26fde17f8ef15e9f58220/configs/components/recom/recom.yaml#L81-L105

What I suggest is getting rid of the `choose_lresume` block by directly adding `recom_restart: ${recom.lresume}` into the main `namelist_changes`:
```yaml
 namelist_changes: 
         namelist.recom: 
                 pavariables: 
                         REcoMDataPath: "${data_path}" 
                         constant_CO2: "${constant_CO2}" 
                         firstyearoffesomcycle: "${firstyearoffesomcycle}" 
                         lastyearoffesomcycle: "${lastyearoffesomcycle}" 
                         numofCO2cycles: "${numofCO2cycles}" 
                         currentCO2cycle: "${currentCO2cycle}" 
                         REcoM_PI: "${REcoM_PI}" 
                         recom_restart: ${recom.lresume}
                 paco2_flux_param: 
                         CO2_for_spinup: "${CO2_for_spinup}" 
```

Let me know if you agree and if I am not missing out something important

mandresm:
By the way, we are handlign the tasks for this PR in a GitHub project: https://github.com/orgs/esm-tools/projects/6/views/1

src/esm_runscripts/filelists.py
denizural: It is not dangerous but an ugly way of dealing with the crazy Echam stuff. I think I don't need that part as well as the lines 709 and below. I will keep it for now but. They will be removed. I already fixed this problem on the yaml side

configs/components/echam/echam.datasets.yaml
```diff
@@ -9,6 +9,7 @@ input_sources:
         "tslclim": "${input_dir}/${resolution}/${resolution}_TSLCLIM.nc"
         "vgratclim": "${adj_input_dir}/${resolution}${ocean_resolution}_VGRATCLIM.nc"
         "vltclim": "${adj_input_dir}/${resolution}${ocean_resolution}_VLTCLIM.nc"
+        # TODO: these (MAC files) are missing in release -> REMOVED
```
mandresm: I think we can remove this comment. The MAC are in the `forcing_sources`.

configs/components/echam/echam.datasets.yaml
```diff
@@ -51,58 +54,100 @@ forcing_sources:
...
+        
         "1950swflux": "${input_dir}/solar_irradiance/swflux_14band_1950.nc"
+        # TODO: years are different
         # ozone
-        "histozone":
-                "${forcing_dir}/ozone/${resolution}_ozone_historical_1850.nc":
-                        to: 1850
-                "${forcing_dir}/ozone/${resolution}_ozone_historical_@YEAR@.nc":
-                        from: 1851
-                        to: 2008
```
mandresm, line 97: Do we really need to remove this?

configs/components/echam/echam.yaml
```diff
@@ -789,6 +822,14 @@ coupling_fields:
         hydr_atm:
                 grid: atmo
 
+        # TODO: absent in release
```
mandresm: This should cause no trouble, I suggest we remove the comment

configs/components/fesom/fesom.yaml
```diff
@@ -260,67 +276,126 @@ config_sources:
...
-ignore_sources:
-        gf: goodfile
-        chloro: Chl_Sweeney.dat
-
+# default output of fesom component
+# 1. variables listed in `*_outputs` but not defined/not requested in the model will be ignored by the fesom-1.4 output scheduler
+# 2. variables defined/requested in the model but not listed in `*_outputs` will raise an error in the fesom-1.4 output scheduler
+# 3. variable names are case sensitive
+# 4. please keep alphabetic order
+# 5. a custom variable list can be provided in the runscript via `further_reading` in the fesom section
```
mandresm:
```suggestion
# 5. a custom variable list can be provided in the runscript via `further_reading` in the fesom section, or in your runscript
```

configs/components/fesom/fesom.yaml
```diff
@@ -260,67 +276,126 @@ config_sources:
...
+    - thetaoga
+    - tos
+    - tso
+    - uo
+    - uwind
+    - vo
+    - vwind
+    - wnet
+    - wo
+    - zos
```
mandresm, line 377: Are there any `recom` variables here?

configs/components/fesom/fesom.yaml
```diff
@@ -1,57 +1,79 @@
...
         License:
-                Please make sure you have a licence to use FESOM. In case you are unsure,
-                please contact redmine...
+                www.fesom.de
 
 restart_rate: "12"
 restart_unit: "m"
 restart_first: 12
 restart_flag: "last"
+yearly_outputs_start: 1 # Year in which the yearly output starts to be saved
```
mandresm, line 81: the default in the `AWI_FESOM_YAML` was `1000` in `release`. Why is it okay/better to keep it 1? Will this affect other people running fesom-1.4 simulations?

configs/components/hdmodel/hdmodel.yaml
```diff
@@ -16,6 +16,8 @@ license_text: |
 pool_dir: "${computer.pool_directories.pool}/JSBACH/input/${jsbach.dataset}/HD"
 input_dir: "${pool_dir}"
 
+# (deniz) TODO: this will return the string NONE. YAML supports a null type:
+# via null, ~, or !!null. Then the corresponding Python type will be None
 executable: NONE
```
mandresm:
```suggestion
executable: null
```

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -1,7 +1,10 @@
+hist_fold: hist_harvest
+hist_scenario: ssp370  # dataset to use for year > 2014
+
 input_sources:
         lctlib: "${echam.model_dir}/lctlib_nlct21.def"
         "jsbach_1850": "${adj_input_dir}/jsbach_${echam.resolution}${echam.ocean_resolution}_11tiles_5layers_1850${jsbachnc_file_ending}.nc"
-        "jsbach_1976": "${adj_input_dir}/jsbach_${echam.resolution}${echam.ocean_resolution}_11tiles_5layers_1976${jsbachnc_file_ending}.nc"
+        "jsbach_1976": "${adj_input_dir}/jsbach_${echam.resolution}${echam.ocean_resolution}_11tiles_5layers_1850${jsbachnc_file_ending}.nc"
```
mandresm: Why was this reverted back to original? The latest change is from @seb-wahl : https://github.com/esm-tools/esm_tools/commit/db283c26a0a70920b9b932425e0a608c0644901f

It kind of make sense that the 1976 file has for name 1976 and not 1850

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -41,32 +81,86 @@ choose_dataset:
                                         from: 1851
         r0010:
                 add_forcing_sources:
-                        "Ndepo_1850": "${input_dir}/${echam.resolution}/${echam.resolution}_ndepo_CMIP_NCAR_CCMI-1-0_gr_185001-185012-clim.nc"
-                        "Ndepo_1950": "${input_dir}/${echam.resolution}/ndepo/Ndepo.1950.nc"
```
mandresm, line 45: Why do we remove the `Ndepo_1950`?

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -41,32 +81,86 @@ choose_dataset:
...
-
-                        "cover_fract_hist_scenario":
-                                "${cover_fract_dir}/cover_fract_11tiles_1850.nc":
-                                        to: 1850
-                                "${cover_fract_dir}/cover_fract_11tiles_@YEAR@.nc":
-                                        from: 1851
+                        "cover_fract_ssp126":
+                                "${cover_fract_dir}/cover_fract_${echam.resolution}_11tiles_1850.nc":
                                         to: 2014
-                                "${cover_fract_dir}/${echam.scenario}/cover_fract_11tiles_@YEAR@.nc":
```
mandresm: Why are these files removed?

mandresm (changes requested):

configs/components/echam/echam.datasets.yaml
```diff
@@ -51,58 +54,100 @@ forcing_sources:
...
                 "${forcing_dir}/volcano_aerosols/strat_aerosol_ir_${resolution}_@YEAR@.nc":
                         from: 1851
-                        to: 2024
-                "${forcing_dir}/volcano_aerosols/strat_aerosol_ir_${resolution}_2024.nc":
-                        from: 2025
+                        to: 2014
+                "${forcing_dir}/volcano_aerosols/strat_aerosol_ir_${resolution}_2014.nc":
+                        from: 2015
+
+        # TODO: important: r0007 has upto 2014 but r0008 has upto 2024 (used by FOCI)
```
mandresm: We decided to use `r0008` so revert these changes

configs/components/echam/echam.datasets.yaml
```diff
@@ -51,58 +54,100 @@ forcing_sources:
...
                         to: 1850
                 "${forcing_dir}/volcano_aerosols/strat_aerosol_sw_${resolution}_@YEAR@.nc":
                         from: 1851
-                        to: 2024
-                "${forcing_dir}/volcano_aerosols/strat_aerosol_sw_${resolution}_2024.nc":
-                        from: 2025
+                        to: 2014
+                "${forcing_dir}/volcano_aerosols/strat_aerosol_sw_${resolution}_2014.nc":
+                        from: 2015
+
```
mandresm: same as above

configs/components/echam/echam.datasets.yaml
mandresm: check on the `esm_tests` if this makes a difference

configs/components/echam/echam.datasets.yaml
```diff
@@ -51,58 +54,100 @@ forcing_sources:
...
-                        to: 2008
         "ozone_hist_scenario":
                 "${forcing_dir}/ozone/${resolution}_ozone_historical_1850.nc":
                         to: 1850
                 "${forcing_dir}/ozone/${resolution}_ozone_historical_@YEAR@.nc":
                         from: 1851
                         to: 2014
+                # scenario: ssp126, ssp119, ssp245, ssp370, ssp585
                 "${forcing_dir}/ozone/${resolution}_ozone_${scenario}_@YEAR@.nc":
                         from: 2015
```
mandresm, line 120: make the years depeding on r0007/r0008?

configs/components/echam/echam.yaml
```diff
@@ -384,31 +363,79 @@ choose_scenario:
...
+            scenario_lowercase: rcp26
+            scenario_type: cmip5
+        RCP45:
+            scenario_lowercase: rcp45
+            scenario_type: cmip5
+        RCP85:
+            scenario_lowercase: rcp85
+            scenario_type: cmip5
+
+# TODO: came from release
```
mandresm: We can remove this. We can delete CMIP5

configs/components/echam/echam.yaml
```diff
@@ -384,31 +363,79 @@ choose_scenario:
...
+                aerofin_1849: piaerofin
+                aerofin_1850: piaerofin
+                aerofin_1851: piaerofin
+                ozone: ozone_cmip5_scenario
+                greenhouse: greenhouse_cmip5_scenario 
+                volcir: histvolcir
+                volcsw: histvolcsw
+                swflux: histswflux
+                MAC-SP: MAC-SP
+
```
mandresm: We can remove this. We can delete CMIP5

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -10,19 +13,56 @@ input_sources:
...
+                                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/rcp45/LUH_harvest_${echam.resolution}_rcp45_@YEAR@.nc":
+                                        from: 2015
+                        "LU_RCP85":
+                                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/${hist_fold}/LUH_harvest_${echam.resolution}_1850.nc":
+                                        to: 1850
+                                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/${hist_fold}/LUH_harvest_${echam.resolution}_@YEAR@.nc":
+                                        from: 1851
+                                        to: 2014
+                                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/rcp85/LUH_harvest_${echam.resolution}_rcp85_@YEAR@.nc":
+                                        from: 2015
```
mandresm: CMIP5 stuff, can be deleted

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -41,32 +81,86 @@ choose_dataset:
...
+                                        from: 2100
+                                        to: 2101
+                        "Ndepo_ssp585":
+                                #"${input_dir}/${echam.resolution}/${echam.resolution}_ndepo.input4MIPs_surfaceFluxes_ScenarioMIP-NCAR-CCMI-ssp585.nc": # nyear x nmonth
+                                "${input_dir_awi}/${echam.resolution}/Ndepo/ssp585/${echam.resolution}_ndepo.input4MIPs_surfaceFluxes_ScenarioMIP-NCAR-CCMI-ssp585_@YEAR@.nc": # 12 months; see README.md in Ndepo/ dir
+                                        from: 1850
+                                        to: 2099
+                                "${input_dir_awi}/${echam.resolution}/Ndepo/ssp585/${echam.resolution}_ndepo.input4MIPs_surfaceFluxes_ScenarioMIP-NCAR-CCMI-ssp585_2099.nc": # 12 months
+                                        from: 2100
+                                        to: 2101
```
mandresm: @denizural will simplify this

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -41,32 +81,86 @@ choose_dataset:
...
+                        "popdens_ssp2":
+                                #"${input_dir}/${echam.resolution}/spitfire/population_density_ssp2_${echam.resolution}.nc": # nyear
+                                "${input_dir_awi}/${echam.resolution}/spitfire/population_density_ssp2_T63.nc": # see README.md in this dir
+                                        from: 1850
+                                        to: 2100
+                        "popdens_ssp5":
+                                #"${input_dir}/${echam.resolution}/spitfire/population_density_ssp5_${echam.resolution}.nc": # nyear
+                                "${input_dir_awi}/${echam.resolution}/spitfire/population_density_ssp5_T63.nc": # see README.md in this dir
+                                        from: 1850
+                                        to: 2100
```
mandresm: @denizural will simply this. Also, we need to add `ssp3`

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -41,32 +81,86 @@ choose_dataset:
...
+                                        from: 2015
+                                        to: 2300
+                        "cover_fract_ssp534os":
+                                "${cover_fract_dir}/cover_fract_${echam.resolution}_11tiles_1850.nc":
+                                        to: 2014
+                                "${input_dir}/${echam.resolution}/land_use_maps/ssp534os/cover_fract_${echam.resolution}_11tiles_@YEAR@.nc":
+                                        from: 2015
+                                        to: 2300
+                        "cover_fract_ssp585":
+                                "${cover_fract_dir}/cover_fract_${echam.resolution}_11tiles_1850.nc":
```
mandresm: Simplify this as well

configs/components/jsbach/jsbach.yaml
```diff
@@ -165,16 +181,118 @@ choose_version:
...
+                                        LU_trans: LU_trans_none
+                                        cover_fract: cover_fract_HIST
+
+                        RCP85:
+                                add_input_files:
+                                        jsbach: jsbach_1850
+                                forcing_files:
+                                        LU: LU_RCP85
+                                        LU_trans: LU_trans_none
+                                        cover_fract: cover_fract_HIST
```
mandresm: CMIP5

configs/components/jsbach/jsbach.yaml
```diff
@@ -165,16 +181,118 @@ choose_version:
...
                         1950:
                                 add_input_files:
                                         jsbach: jsbach_1976
                                 forcing_files:
                                         LU: LU_1950
                                         LU_trans: LU_trans_none
+                                        Ndepo: Ndepo_1950
+                                        a_nd_file: a_nd_file
+                                        #lightning: lightning
+                                        #popdens: popdens
```
mandresm, line 200: Check with `release`/ask the person who did this

configs/components/jsbach/jsbach.yaml
```diff
@@ -165,16 +181,118 @@ choose_version:
...
+                                add_input_files:
+                                        jsbach: jsbach_1850
+                                forcing_files:
+                                        LU: LU_ssp585
+                                        LU_trans: LU_trans_ssp585
+                                        cover_fract: cover_fract_ssp585
+                                        Ndepo: Ndepo_ssp585
+                                        a_nd_file: a_nd_file
+                                        lightning: lightning
+                                        popdens: popdens_ssp5
```
mandresm: Can be removed

configs/components/jsbach/jsbach.yaml
```diff
@@ -249,6 +373,30 @@ choose_version:
...
+                                        cover_fract: cover_fract_HIST
+                        RCP85:
+                                add_input_files:
+                                        jsbach: jsbach_1850
+                                forcing_files:
+                                        LU: LU_RCP85
+                                        LU_trans: LU_trans_none
+                                        cover_fract: cover_fract_HIST
+
+# common options for scenarios, eg. CMIP6
```
mandresm: Can be removed as it is CMIP5

configs/components/jsbach/jsbach.yaml
```diff
@@ -349,27 +491,37 @@ outdata_sources:
         "[[streams-->STREAM]]": ${general.expid}_${start_date!syear}*.${start_date!sday}_STREAM
         "[[streams-->STREAM]]_codes": ${general.expid}_${start_date!syear}*.${start_date!sday}_STREAM.codes
 
+outdata_in_work:
+        "[[streams-->STREAM]]": ${general.expid}_${start_date!syear}*.${start_date!sday}_STREAM
+        "[[streams-->STREAM]]_codes": ${general.expid}_${start_date!syear}*.${start_date!sday}_STREAM.codes
+
 restart_in_files:
         "[[streams-->STREAM]]": STREAM
```
mandresm, line 410: Open an issue for the `.nc` stream outdata files.

mandresm: #384

configs/components/jsbach/jsbach.yaml
```diff
@@ -349,27 +491,37 @@ outdata_sources:
...
 
 restart_in_sources:
-        "[[streams-->STREAM]]": restart_${parent_expid}_STREAM_${parent_date!syear!smonth!sday}.nc
+        # deniz: removed hour, minute, second information. Now it is compatible with ECHAM
+        # INFO: for continuous restarts it is fine but branch-off restarts fail
+        # since esm-tools is looking for YYYYMMDDHHMMSS.nc data instead
+        # "[[streams-->STREAM]]": restart_${parent_expid}_${parent_date!syear!smonth!sday!shour!sminute!ssecond}_STREAM.nc
+        "[[streams-->STREAM]]": restart_${parent_expid}_${parent_date!syear!smonth!sday}_STREAM.nc
```
mandresm, line 416: #384

configs/components/jsbach/jsbach.yaml
```diff
@@ -349,27 +491,37 @@ outdata_sources:
...
         "[[streams-->STREAM]]": STREAM
 
 restart_out_sources:
-        "[[streams-->STREAM]]": restart_${general.expid}_STREAM_${other_date!syear!smonth!sday}.nc
+        # deniz: removed hour, minute, second information. Now it is compatible with ECHAM
+        # INFO: for continuous restarts it is fine but branch-off restarts fail
+        # since esm-tools is looking for YYYYMMDDHHMMSS.nc data instead
+        # Same as the problem we have in `restart_in_sources`
+        # "[[streams-->STREAM]]": restart_${general.expid}_${other_date!syear!smonth!sday!shour!sminute!ssecond}_STREAM.nc
+        "[[streams-->STREAM]]": restart_${general.expid}_${other_date!syear!smonth!sday}_STREAM.nc
```
mandresm, line 430: #384

mandresm (changes requested):

configs/components/recom/recom.yaml
```diff
@@ -1,55 +1,235 @@
...
+            - PAR
+            - pCO2s
+            - slp
+            - virtual_alk
+            - virtual_dfe
+            - virtual_dic
+            - virtual_din
+            - virtual_dsi
+            - virtual_oxy
+            - w10w
```
mandresm: see if there is any redundancy here and try to simplify things with `add_`

configs/components/recom/recom.yaml
```diff
@@ -1,55 +1,235 @@
...
```
mandresm: check for redundancy

configs/components/recom/recom.yaml
```diff
@@ -65,32 +245,51 @@ choose_scenario:
...
                 pavariables:
-                        bgc_num: "${bgc_num}"
                         REcoMDataPath: "${data_path}"
                         constant_CO2: "${constant_CO2}"
                         firstyearoffesomcycle: "${firstyearoffesomcycle}"
                         lastyearoffesomcycle: "${lastyearoffesomcycle}"
                         numofCO2cycles: "${numofCO2cycles}"
                         currentCO2cycle: "${currentCO2cycle}"
                         REcoM_PI: "${REcoM_PI}"
+                        recom_restart: "${recom.lresume}"  # this was in the feat/ssp branch
```
mandresm, line 161: perfect! we need to delete `choose_lresume` below, it makes no sense to have that choose when we can have this simple line here.

configs/setups/fesom-recom/fesom-recom.yaml
```diff
@@ -95,91 +90,11 @@ fesom:
...
-                                Kh0: 800.0
-                                brine_rejection_param: false
-                        oce_scheme:
-                                gamma_fct: 0.4
-                        age_tracer:
-                                use_age_tracer: true
-                                age_tracer_start_year: "${age_tracer_sy}"
-                        tracer_cutoff:
-                                limit_salinity: false
-                                salinity_min: 5.0
```
mandresm, line 128: This changes take place only for `fesom-1.4-recom` simulations, and where needed to reproduce exactly the control runs provided by Judith a few years ago. Do we really want to change this settings an let fesom use the default fesom namelists defined in ESM-Tools for `fesom-recom`?

configs/setups/fesom-recom/fesom-recom.yaml
```diff
@@ -95,91 +90,11 @@ fesom:
...
-                                                rate: 1
-                                        -
-                                                vars: ${recom.12_step_outputs}
-                                                unit: s
-                                                rate: 12
-                                        -
-                                                vars: ${recom.yearly_outputs}
-                                                unit: y
-                                                first: ${recom.yearly_outputs_start}
-                                                rate: 1
```
mandresm, line 181: @mandresm check this with `esm_tests`

configs/setups/fesom-recom/fesom-recom.yaml
```diff
@@ -204,14 +119,19 @@ recom:
...
                                 1:      # Restart from runscript
                                         ini_parent_exp_id: "${general.ini_parent_exp_id}"
                                         ini_restart_dir: "${general.ini_restart_dir}/recom/"
                                         ini_parent_date: "${prev_date}"
+
+        # setup-specific namelist.recom 
+        config_sources:
+                recom: "${namelist_dir}/namelist.recom"
+        add_config_in_work:
+                recom: "namelist.recom"
```
mandresm: Delete this, this is the `recom` namelist and should come from the `recom.yaml` or the `recom` section in this file

src/esm_runscripts/filelists.py
```diff
@@ -311,12 +312,13 @@ def target_subfolders(config):
                 for descr, filename in six.iteritems(
                     config[model][filetype + "_targets"]
                 ):  # * only in targets if denotes subfolder
+                    # TODO: change with user_error()
                     if not descr in config[model][filetype + "_sources"]:
                         print(
                             "no source found for target " + name + " in model " + model,
                             flush=True,
                         )
```
mandresm: as the comment says, "change with user_error"

mandresm (changes requested):

src/esm_runscripts/filelists.py
mandresm: @denizural , I'm take care of this one now if that's okay with you

configs/components/echam/echam.datasets.yaml
denizural: Should not cause an issue because `HIST` now has `ozone: ozone_hist_scenario` which basically contains
```yaml
"ozone_hist_scenario":
        "${forcing_dir}/ozone/${resolution}_ozone_historical_1850.nc":
                to: 1850
        "${forcing_dir}/ozone/${resolution}_ozone_historical_@YEAR@.nc":
                from: 1851
                to: 2014
```
plus the future scenario

configs/components/fesom/fesom.yaml
chrisdane: No

configs/components/echam/echam.datasets.yaml
denizural: It looks to me that r8 extends r7. r8 contains both the historical and the future data. I calculated the checksums of the same file names in both directories and they seem to be the same. So, let's go with r8 since it has more data. 
I have no idea they are storing the duplicate data in a different directory 🤔

configs/components/fesom/fesom.yaml
chrisdane: I assume that was decided to save disk space. I dont know who decided that. I think the default output block for yearly outputs should also start in the first year since this is what you would assume, right?
If a user wants special settings (like 'save yearly outputs only after the 1000th model year'), the user can simply edit this block.

configs/components/jsbach/jsbach.datasets.yaml
chrisdane: I didnt do that.

configs/components/jsbach/jsbach.datasets.yaml
chrisdane: I didnt do that.

configs/components/fesom/fesom.yaml
```diff
@@ -359,13 +466,8 @@ coupling_fields:
...
         sie_feom:
                 grid: feom
-        snt_feom:
+        snt_feom: # snow temp
                 grid: feom
-                #description: |
-                #       whatever snt means
-                #unit: kg m s
-                #standard_cf_name: 'shit no tuna?'
-                #cf_code: 42
```
mandresm, line 368: Oh no! The epic "shit no tuna" is gone :'( , one of my first memories of reading the yaml files for the first time

configs/setups/fesom-recom/fesom-recom.yaml
chrisdane: Yes thats fine.

configs/components/fesom/fesom.yaml
chrisdane: Yeah sorry. I thought its time to move on ;)

configs/components/recom/recom.yaml
mandresm: Yes, I do see a compromise :) We can place this line in the `awicm.yaml` so that's only used for AWICM-REcoM this way. Would that work for you?

chrisdane: Ok but its the case for both setups: `fesom-recom` and `awicm-1.0-recom`.

mandresm: Actually, we can just leave it as it is right now, so nevermind. Would you like to also load the model from the namelist for AWICM-REcoM? At the moment the default in `awicm.yaml` is `namelist_dir: "${esm_namelist_dir}/${model}/awicm_recom/"`

configs/components/fesom/fesom.yaml
mandresm: Okay, thanks for the info :)

configs/components/jsbach/jsbach.datasets.yaml
denizural: I discovered how that happened. On Feb 4 when the `origin/ssp585_awicm-recom` was merged into `feat/ssp585_awicm-recom_v6`, that line was overwritten. It basically took the previous thing at overwrote the new one. I am now reverting it to Sebastian's version

configs/components/jsbach/jsbach.datasets.yaml
mandresm: @denizural, I'm taking care of this now

configs/components/echam/echam.yaml
```diff
@@ -646,11 +643,21 @@ streamsnc:
...
+        - sp
+        - gl
+        - g3b
+        - scm
+        - ma
+        - surf
+        - cfdiag
+        - aeropt
+        - co2
+        - tdiag
```
denizural, line 650: I hope these will not break anything but what about the performance? Eg. carrying extra data can slow down Echam. What do you think @chrisdane @mandresm?

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -60,38 +60,32 @@ choose_dataset:
...
+                                        from: 1851
+                                        to: 2014
+                                    # TODO: these files don't exist on Levante
+                                    "${input_dir}/${echam.resolution}/ndepo/Ndepo_${echam.scenario}.@YEAR@.nc":
+                                        from: 2015
+                            # TODO: test this with a check run
+                                True:
+                                    # These files are 12-monthly; see README.md in Ndepo/ directory
+                                    # NOTE: ssp370 might be missing in this directory
+                                    "${input_dir_awi}/${echam.resolution}/Ndepo/${echam.scenario}/${echam.resolution}_ndepo.input4MIPs_surfaceFluxes_ScenarioMIP-NCAR-CCMI-${echam.scenario}_@YEAR@.nc":
```
mandresm: This choose won't be reached and it won't be resolved. Choose cannot be indented anywhere, except for sections (models, computer, general, ...) or other chooses.

I suggest the following:
- [x] 1. remove the `choose`
- [x] 2. leave `Ndepo_hist_scenario` as it was
- [] 3. create a new `Ndepo_ssp` label and add the changes there

mandresm (changes requested):

configs/components/jsbach/jsbach.datasets.yaml
denizural: I thought that parser was clever 😭

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -10,18 +13,44 @@ input_sources:
...
+        # yearly files where each file contain monthly data
+        Ndepo_ssp: 
+                # These files are 12-monthly; see README.md in Ndepo/ directory
+                # NOTE: ssp370 might be missing in this directory
+                "${input_dir_awi}/${echam.resolution}/Ndepo/${echam.scenario}/${echam.resolution}_ndepo.input4MIPs_surfaceFluxes_ScenarioMIP-NCAR-CCMI-${echam.scenario}_@YEAR@.nc": 
+                    from: 1850
+                    to: 2099
+                "${input_dir_awi}/${echam.resolution}/Ndepo/${echam.scenario}/${echam.resolution}_ndepo.input4MIPs_surfaceFluxes_ScenarioMIP-NCAR-CCMI-${echam.scenario}_2099.nc":
+                    from: 2100
+                    to: 2101
```
denizural, line 56: @mandresm, is this a good solution? The users can choose `Ndepo_ssp` in their runscript as `Ndepo: Ndepo_ssp`

configs/components/jsbach/jsbach.datasets.yaml
denizural: @mandresm, This has the same issue with `Ndepo`. Should we leave them like this so that the user can take this file as `popdens: popdens_ssp1`?

denizural: @seb-wahl and @chrisdane, we need your guidance here. Which popdens file should be used in CMIP6 simulations. Chris has multiple files for different scenarios whereas @seb-wahl has only one for CMIP6.

configs/components/echam/echam.yaml
mandresm: If the files are not created, they won't be copied, so I think performance wise is okay

chrisdane: Yes. Echam decides which data data created, not the esm_tools. The problem here is rather that the user will get a file-not-found warning by the esm_tools if one of those output streams is not created (but this is again this big issue #384).

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -10,18 +13,44 @@ input_sources:
...
-                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/hist_harvest/LUH_harvest_${echam.resolution}_1850.nc":
+                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/${hist_fold}/LUH_harvest_${echam.resolution}_1850.nc":
                         to: 1850
-                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/hist_harvest/LUH_harvest_${echam.resolution}_@YEAR@.nc":
+                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/${hist_fold}/LUH_harvest_${echam.resolution}_@YEAR@.nc":
                         from: 1851
+                        to: 2014
+                # HIST scenario will use this data beyond year 2014
+                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/${hist_scenario}_harvest/LUH_harvest_${echam.resolution}_@YEAR@.nc":
+                        from: 2015
```
mandresm:
```suggestion
                # HIST scenario will use this data beyond year 2014
                "${input_dir}/${echam.resolution}/New_Hampshire_LCC/${scenario}_harvest/LUH_harvest_${echam.resolution}_@YEAR@.nc":
                        from: 2015
```
@chrisdane , shouldn't this simply be the `jsbach.scenario`? The way is coded right now it will always load ssp370 one unless you specify in your runscript `jsbach.hist_scenario: ssp370` which is kind of a repetition if you already have the `general.scenario` defined already in the runscript

configs/components/jsbach/jsbach.datasets.yaml
mandresm: I agree, this looks like a good solution to me. Maybe make this be also for the `ssp` scenarios?

configs/components/jsbach/jsbach.datasets.yaml
chrisdane: I dont have `jsbach.scenario`, you mean `jsbach.scenario_type` which is only `echam.scenario`? No, please leave it as it is as here a different thing than `echam.scenario` is needed. I think the variable name `hist_scenario` is not a good choice. Its rather confusing as there is only one historical scenario. The only purpose of the usage of this variable in the `LU_hist` definition is for years larger than 2014. This, in turn, only happens due to the necessity to load forcing data of the current year + 1 to do an interpolation. Hence, as the historical simulation in CMIP6 ends in 2014, the only data that is needed by echam is data from 2015 and we agreed to choose the one from the ssp370 scenario (it does not really matter which is chosen). Thus, I suggest to rename `hist_scenario` to something like  `hist_extend_scenario` and make a comment that explains its meaning right at its definition.

mandresm: I see, thanks a lot. I'll rename the variable and copy part of your explanation as a comment on top of the variable.

mandresm (approved):
After generalization and update of the `esm_tests` `last_tested` files, all tests passed and it is ready to merge!

mandresm:
#bump

Issue #667, opened 2022-05-16
flatten merged lists


chrisdane:
**Is your feature request related to a problem? Please describe.**
I would like to merge two lists li1 and li2 to a new list li3 that contains li1 and li2 in a flattened and not nested way.

**Describe the solution you'd like**
Paste this code in https://yaml-online-parser.appspot.com:
```
li1: &li1 [a, b]
li2: &li2 [c, d]
li3: 
  - *li1
  - *li2
```
the result is
```
  "li3": [
    [
      "a", 
      "b"
    ], 
    [
      "c", 
      "d"
    ]
```
i.e. the entries in li3 are nested and not on the same level. I need:
```
a
b
c
d
```

**Describe alternatives you've considered**
I tried many versions found on SO, e.g. https://stackoverflow.com/questions/24090177/how-to-merge-yaml-arrays. It seems that feature is somewhat not implemented in base python: https://softwareengineering.stackexchange.com/questions/254279/why-doesnt-python-have-a-flatten-function-for-lists

Thanks for consideration.

chrisdane:
can we use this somehow: https://pyyaml.docsforge.com/master/api/yaml/constructor/SafeConstructor/flatten_mapping/ ?

pgierz:
Hi Chris,

this may be possible with the special "math" syntax in the ESM-Tools extension for YAML, which effectively tricks Python into do an evaluation of code disguised as a string. It is not something I would recommend leaning on heavily, as it may be error-prone.

```yaml
general:
   list_1:
     - "a"
     - "b"
     - "c"
   list_2:
     - "d"
     - "e"
     - "f"
   list_merge: $(( ${list_1} + ${general.list_2} ))
``` 
In Python, that translates to:
```python
list_1 = ["a", "b", "c"]
list_2 = ["d", "e", "f"]
list_merge = eval("list_1 + list_2")
print(list_merge)
["a", "b", "c", "d", "e", "f"]
```
Take special notice of where I have quotes, especially in the `eval` statement.

Again, this **may** work, or maybe not. I have not tested it yet.

chrisdane:
No, the second list2 is ignored in the mergelist.

denizural:
A solution is found. I willl make a PR later on.
Here: https://github.com/esm-tools/esm_tools/pull/672

mandresm:
[...]

PR #672, opened 2022-05-20
Hotfix/issue 667

denizural:
This bugifx will solve the issue: https://github.com/esm-tools/esm_tools/issues/667

The main problem is that `add_` command can't handle merging (concatenation) of referenced lists correctly. It yields nested lists. 
On the other hand, YAML specification provides a way to reference / dereference the containers similar to the C language. 
However, merging of sequences (lists) is not supported in YAML. Only, mappings can be merged.

In this bugfix, I improved the functionality in ESM-Tools to merge and flatten lists using a new tag `!esm-flatten`. Here is a quick example:
```yaml
list_1: &list_1
- 1
- 2
- 3

list_2: &list_2
- 10
- 11
- 12
- 3.1415
- 'foo'
- true
- false

list_flat: !esm-flatten
- *list_1
- *list_2
- new_item
```

@pgierz, you can find a detailed example in Ollie: `/work/ollie/dural/fesom_test/issue_667`
You can vimdiff:
`vimdiff bug_finished_config_RELEASE.yaml bug_finished_config_BUGFIX.yaml`

There is also a sample runscript that you can try.

The advantage of this solution is it does not touch the `add_` functionality but rather adds a new feature. The simple test shows that it does not break anything.

@pgierz, I also added some `TODO` tags in the code. During the clean-up and future refactoring phases, we need to touch on these points. I believe that we need to write a better loader. Eg. `esm_tools_yaml_loader`. This will eliminate the multiple instantiations of loaders. I believe that a YAML loaders must be created only once and all YAML reads must be handled by that.

If this bugfix gets approved, I will also add this to our documentation.

I also found out why your `eval` method does not work. But I need to talk to you to explain that.

@chrisdane, does this solve what you want to achieve?


src/esm_parser/yaml_to_dict.py
```diff
@@ -9,6 +9,12 @@
 
 YAML_AUTO_EXTENSIONS = ["", ".yml", ".yaml", ".YML", ".YAML"]
 
+# ===
+# Extra tags to extend the basic YAML capabilities
+# ===
+# makes the lists flat after merging
+ESM_FLATTEN_TAG = "!esm-flatten"
```
pgierz, line 16: I am not a big fan of adding new custom YAML tags. That is one more point of confusion, IMO. I was already not too happy doing this for the `!ENV` tag, as it makes the YAML files non-portable to other systems; should we ever need to use them.

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
+        raise TypeError(f"{tag} can not be used as a constructor")
+    return constructor(node.value)
+
+
+def flatten_yaml_sequence(sequence):
+    """Recursively flatten a YAML sequence and return a generator.
+
+    Parameters
+    ----------
+    sequence : yaml.SequenceNode)
```
pgierz, line 107: Extra `)` at the end is not needed

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
+        "bool": str_to_bool,
+    }
+    constructor = tag_to_constructor_map.get(tag, None)
+    if constructor is None:
+        raise TypeError(f"{tag} can not be used as a constructor")
+    return constructor(node.value)
+
+
+def flatten_yaml_sequence(sequence):
+    """Recursively flatten a YAML sequence and return a generator.
```
pgierz, line 103: In this function I would like to see a section with:
```
Examples
--------
...something...
```

And a unit test to be added to the test suite.

We always talk about doing TDD; so, why not start here?

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
+
+    Raises
+    ------
+    TypeError
+        When the items are not the instances of yaml.SequenceNode or yaml.ScalarNode
+    """
+    if isinstance(sequence, yaml.ScalarNode):
+        yield node_to_value(sequence.value)
+
+    # LBYL type checks
```
pgierz, line 123: Remove this comment; not everyone knows what LBYL means, and not everyone needs to...

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
+    if isinstance(sequence, yaml.MappingNode):
+        # I am sure that someone will try to flatten a dictionary instead
+        # of a list
+        err_type = "Type Error"
```
pgierz, line 127: Let's stay consistent with real Python and here use `"TypeError"`

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
+    ------
+    TypeError
+        When the items are not the instances of yaml.SequenceNode or yaml.ScalarNode
+    """
+    if isinstance(sequence, yaml.ScalarNode):
+        yield node_to_value(sequence.value)
+
+    # LBYL type checks
+    if isinstance(sequence, yaml.MappingNode):
+        # I am sure that someone will try to flatten a dictionary instead
```
pgierz, line 125: I am not sure this will happen, but OK; also, let's try to avoid this kind of comment in the codebase. I'd rather have comments that only explain implementation, not why a particular implementation was chosen.

src/esm_parser/yaml_to_dict.py
```diff
@@ -569,7 +682,7 @@ def __init__(self, fpath, yaml_error):
             try:
                 print("\n\n\n" + yaml_error)
             except:
-                self.message = ("\n\n A syntax error has been found in "+fpath+"\n")
+                self.message = "\n\n A syntax error has been found in " + fpath + "\n"
```
pgierz, line 685: Use an f-string here.

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
+
+        elif isinstance(item, yaml.ScalarNode):
+            yield node_to_value(item)
+
+        else:
+            raise TypeError(f"{ESM_FLATTEN_TAG} can only take scalar nodes, not {item}")
+
+
+# Handling of the ESM_FLATTEN_TAG (!esm-flatten)
+def flat_sequence_constructor(loader, node):
```
pgierz, line 147: Why not return a list from `flatten_yaml_sequence` in the first place? Then this entire function can just be removed.

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
+    # LBYL type checks
+    if isinstance(sequence, yaml.MappingNode):
+        # I am sure that someone will try to flatten a dictionary instead
+        # of a list
+        err_type = "Type Error"
+        msg = "Dictionaries can not be flattened. Only lists (aka arrays, sequences)"
+        esm_parser.user_error(err_type, msg)
+
+    if not isinstance(sequence, yaml.SequenceNode):
+        raise TypeError(f"Unable to flatten the type: {type(sequence)}")
```
pgierz, line 132: This is nice.

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
+    # Recursively loop over a SequenceNode
+    for item in sequence.value:
+        if isinstance(item, yaml.SequenceNode):
+            yield from flatten_yaml_sequence(item)
+
+        elif isinstance(item, yaml.ScalarNode):
+            yield node_to_value(item)
+
+        else:
+            raise TypeError(f"{ESM_FLATTEN_TAG} can only take scalar nodes, not {item}")
```
pgierz, line 143: Maybe provide the `type(item)` here as well to aid in debugging.

mandresm:
Wouldn't it be better to expand the `add_` capabilities, now that the users are familiarized with it, instead of extending our syntax even more?

For example:
```
list1:
    - 1
    - 2
    - 3
 list2:
    - 4
    - 5
final_list: ${list1}
add_final_list: ${list2}
```

pgierz (changes requested):
If this goes in (I don’t think it should, as Miguel’s solution is nicer), I have some clean up comments.

src/esm_parser/yaml_to_dict.py
denizural: Yes, just working on TDD 😀

src/esm_parser/yaml_to_dict.py
denizural: Well, this is more for the developers.

src/esm_parser/yaml_to_dict.py
denizural: This is from Black but I agree to opt for f-strings over string concats.

src/esm_parser/yaml_to_dict.py
denizural: Well, then brace for a long traceback. Didn't we decide on providing a good error message (via `user_error`)?

mandresm:
Any updates on this?

denizural:
I am still working on it (should not take a lot of time) but I had to postpone it for the Levante work. I will return back asap.

mandresm:
I'm taking over this one as @denizural is working on the cleanup of #665. This is needed for #665

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
 
+def str_to_bool(string):
+    """Handy function to cast strings to bools. Eg.
+    bool('false') -> True
+    str_to_bool('false') -> False
+    str_to_bool('0') -> False
+    """
+    truthy_values = ["true", "1"]
+    return string.lower() in truthy_values
+
```
mandresm, line 66: An equivalent function already exists inside `esm_master`

src/esm_parser/yaml_to_dict.py
```diff
@@ -49,6 +55,106 @@ def __init__(self, fpath, yaml_error):
...
+    tag_to_constructor_map = {
+        "int": int,
+        "float": float,
+        "str": str,
+        "bool": str_to_bool,
+    }
+    constructor = tag_to_constructor_map.get(tag, None)
+    if constructor is None:
+        raise TypeError(f"{tag} can not be used as a constructor")
+    return constructor(node.value)
```
mandresm, line 99: Is this related to the problem this PR is trying to solve? If not I'd suggest to move it to a separate PR.

mandresm (changes requested):
This PR doesn't have Levante. I'll merge `release` so that I can test it.

src/esm_parser/yaml_to_dict.py
denizural: Do you mean parser? I could not find it in `esm_master`. Do you mean `could_be_bool` function in the parser? Unfortunately, It does not do what the function above can do. Eg.
```python
>>> could_be_bool(0)
False
```

or this one:
```python
>>> print(to_boolean(0))
None

>>> print(to_boolean('0'))
None
```

denizural: By the way, a new solution is pushed so this one is obsolete now.

mandresm: https://github.com/esm-tools/esm_tools/blob/781c659c912c1bdefaecdff15e564b79c4a9e184/src/esm_parser/esm_parser.py#L2483-L2498

Also the `choose_` block have a way of evaluating 1 and 0 (any type) as booleans if the cases are True or False and the other way around.

mandresm:
This PR does not solve the following:
```
    list1:
        - "el1"
        - "el2"
    list2:
        - "el3"
        - "el4"
    list_final: ${list1}
    add_list_final: ${list2}
```

denizural:
I am closing this pull request as the PR https://github.com/esm-tools/esm_tools/pull/716/ solved the issue https://github.com/esm-tools/esm_tools/issues/667 in a better way

PR #677, opened 2022-06-01
update levante and remove double entry

JanStreffing:


pgierz:
Not bumping, as simply a documentation update

PR #684, opened 2022-06-03
Feat/esm tests

mandresm:
ESM-Tests final implementation, including check compilations and runs (comparing files with the last test) for CI.

**TODO**

Strikeout items are to be solved after the merge as the merge is urgent.

- [x] Docstring all
- [x] Finish cleanup
- [ ] <del>Documentation
- [x] Implement jobs for all computers
  - [x] ollie
  - [x] levante
  - [x] hlrn
  - [x] aleph
  - [x] juwels
- [ ] <del>ESM-Tests status (`esm_tests -t`) pushed somewhere for visualization (home at github/readthedocs?)
- [x] ESM-Tests status triggered with `esm_tools --test-state`


pgierz:
Hooray for auto-tests! I'll have some time to look over this after the lunch break

mandresm:
It's a very complicated mess with a lot of tricks to not change the ESM-Tools source code but still be able to test it for consistency... but it works :D

src/esm_tests/cli.py
```diff
@@ -142,12 +152,13 @@ def main():
         sys.exit(1)
 
     # Get user info for testing
-    info = user_config(info)
+    user_config(info)
```
pgierz, line 67: You can probably remove this; if it isn't caught in a variable it will just land in memory nirvana, or does the `user_info` function do something magical to the `info` dictionary?

src/esm_tests/cli.py
```diff
@@ -161,6 +172,12 @@ def main():
         print(e)
         raise
 
+    # Special actions for running from GitHub servers
+    if info["in_github"]:
```
pgierz, line 87: maybe with a `get`? I don't know the codebase enough, but just ideas for simple things that might otherwise explode

mandresm:
In fact, the test fail because it is finding the change in the FESOM-REcoM branch that happen yesterday in release XD

pgierz:
Do you want feedback on src for esm-tests? I can give some ideas at least on the changes in this PR; and I can clear off a day to go through the whole source code, but only if you want/feel that it is needed. If we are all happy saying "it works, don't touch it" that's fine as well :-)

mandresm:
Up to you, feedback is always good and I can learn a lot, if you feel you have time for it go for it!

denizural:
A quick question. Is there a specific reason to use `.yml` instead of `.yaml` extension? It looks like these are GitHub actions files, right?

mandresm:
I think they can have both extensions (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows), and I think it was @pgierz that wrote the first workflow, so I just kept it the same. I think it's because the default workflow provided by GitHub is named with `.yml` so we all followed

pgierz:
...yup I just followed the docu example.

src/esm_runscripts/namelists.py
```diff
@@ -491,7 +492,17 @@ def apply_iceberg_calving(config):
                     ib_num_old = 0
 
                 print(" * iceberg_dir = ", config["fesom"].get("iceberg_dir"))
-                ib_num_new = sum(1 for line in open(config["fesom"].get("iceberg_dir") + "/LON.dat"))
+                if os.path.isfile(config["fesom"].get("iceberg_dir") + "/LON.dat"):
```
denizural: Won't change the functionality but I would do it like
```python
fesom_iceberg_dir = config["fesom"].get("iceberg_dir", None)
```
since this is used multiple times after that. This would decrease the code repetition.

src/esm_tests/tests.py
```diff
@@ -997,7 +1069,10 @@ def format_results(info):
             if not state["comp"]:
                 compilation = "compilation failed"
 
-            run = "runs"
+            if info["actually_run"]:
+                run = "runs"
+            else:
+                run = "run files identical"
 
             if not state.get("submission_files_identical", True):
```
denizural: For a future code refactoring, these lines are a good candidate for using `enums`:
```python
from enum import Enum

class States(Enum):
    RUNS = 0
    RUN_FILES_IDENTICAL = 1
    RUN_FILES_DIFFER = 2
    SUBMISSION_FAILED = 3
    RUN_FILES_NOT_FOUND = 4
    RUN_FAILED = 5
``` 
You can then use these constants instead of a dictionary. It will also make it more compatible across other functions. But at this point it is not that important.

src/esm_runscripts/namelists.py
mandresm: Modified as suggested

src/esm_tests/tests.py
mandresm: Can you clarify how would you use `enum` to improve this part of the code? I am not able to understand it just with the info you provided. I understand the class you built, but how would you use it and how is it superior to the dictionary?

pgierz: Short answer: you save some square brackets and quotes, and it's more "readable", although that gets strongly into the personal opinion side of things. Downside is you introduce a construct not everyone is familiar with.

I am sure there are more reasons for enum being a nice thing (there is for instance an automatic way to create [automatic values](https://docs.python.org/3/library/enum.html#using-automatic-values) ), but unless you want to try out some new code karate it is at this level rather redundant.


- - - -

The above is an outdated comment from @denizural's outdated `enum` suggestion

pgierz: @denizural if you have suggestions like this, why not directly add them to Miguel's branch?
- - - -
The above is also an outdated comment from @denizural enum suggestion.

mandresm:
Hi everyone, something I have in mind about `esm_tests`. Appart from doing the tests (either checks and actual tests), storing all the configuration files and comparing against old ones, `esm_tests` also stores the information of the last actual tests in `esm_tools/src/resources/state.yaml`. This information can be printed with `esm_tests -t` (see image below). I think this information is very valuable for the user, as it shows what is supported and when was last tested successfully (or not). I have several questions for you about it:

1. Should we add an `esm_tools --state` command  (that basically links `esm_tests -t`)  so that the users can get that info from ESM-Tools without having to know anything about `esm_tests`?

2. Should we add this information anywhere in the documentation/github homepage? If the answer is yes, and you have some suggestions for a format (table/list/status images...) please let me konw.

Screen Shot 2022-06-13 at 11 01 45

denizural:
This idea sounds good. It is called `regression` in software testing. 

Maybe we can host this information on GitHub and and when the command is invoked, `esm_tools` can grab the info from GitHub (read only for the users, only developers can write) and display on the screen.

I would call it something like `--test-state`, btw just to reserve more generic `state` for some future work. But of course, it is up to you to choose a name.

src/esm_runscripts/namelists.py
```diff
@@ -491,7 +492,18 @@ def apply_iceberg_calving(config):
                     ib_num_old = 0
 
                 print(" * iceberg_dir = ", config["fesom"].get("iceberg_dir"))
-                ib_num_new = sum(1 for line in open(config["fesom"].get("iceberg_dir") + "/LON.dat"))
+                iceberg_dir = config["fesom"].get("iceberg_dir", None)
+                if os.path.isfile(iceberg_dir + "/LON.dat"):
+                    ib_num_new = sum(1 for line in open(config["fesom"].get("iceberg_dir") + "/LON.dat"))
+                elif config["general"]["check"]:
```
pgierz: I am not sure I like this, since the `check` mode might be on also for other reasons. Can we make it so that `esm-tests` gets its own flag for `esm-runscripts`?

src/esm_runscripts/namelists.py
```diff
@@ -491,7 +492,18 @@ def apply_iceberg_calving(config):
...
+                    # MA: Get the number of icebergs from the path for check runs where the file is not
+                    # available (e.g. CI in GitHub)
+                    ib_num_new = int([x for x in iceberg_dir.split("/") if len(x)>0][-1].split("_")[-1])
```
pgierz: Sorry but...em....what? I trust you that this works, but what wizardry is this? You are counting the number of non-empty path elements, but then the last one, and then the last one after splitting the `_`? 

Even if it never breaks, someone at some point will stumble across this code and be highly confused. Just put a comment or something that shows what number this gets. I assume it is something like:
```
/some///./crazy/path/with///a/new/file_38125
```
which yields
```
38125
```

src/esm_tests/cli.py
```diff
@@ -71,13 +71,6 @@ def main():
         help="Delete previous tests",
         action="store_true",
     )
-    # parser.add_argument(
```
pgierz, line 74: Thank you!! We should always opt rather for erasing code than commenting it out, in my view. Code can be retrieved from the git history if you really need.

src/esm_tests/cli.py
```diff
@@ -105,6 +98,21 @@ def main():
         help="bullet points for printing the results",
         action="store_true",
     )
+    parser.add_argument(
+        "-g",
+        "--github",
+        default=False,
+        help="use this flag when running in GitHub servers",
```
pgierz: OK, this already answers my comment from further above.

src/esm_tests/cli.py
```diff
@@ -142,12 +151,17 @@ def main():
...
 
     # Get user info for testing
-    info = user_config(info)
+    user_config(info)
 
     # User-specific Info to remove from the files ``last_tested`` files
+    info["mnt"] = ""
+    pwd = os.getcwd()
+    if pwd.startswith("/mnt/lustre"):
+        info["mnt"] = "/".join(pwd.split("/")[:3])
```
pgierz: As above, this works, but looks confusing. Probably it would be nice to have something in the `helpers.py` which can mutate paths. I have no problem there including computer-specific functions. As an example:
```python
def clean_computer_specific_paths(path, computers=None):
    if computers is None:
        # Or some sensible default:
        computers = ["mistral", "levante"]:
    # ....something that grabs the computer config yaml .....
    for computer_config in computer_configs:
        for remove_prefix in computer_config["allowed_removes"]:
            path = path.replace(remove_prefix, "")
    return path



```

src/esm_tests/cli.py
```diff
@@ -142,12 +151,17 @@ def main():
...
     info["rm_user_info"] = {
-        "ACCOUNT": info["user"]["account"],
         "TEST_DIR": info["user"]["test_dir"],
+        "HOME_DIR": f"{os.path.expanduser('~')}",
+        #"ACCOUNT": info["user"]["account"],
```
pgierz: Could be included as `.get("account")` and then this defaults to `None`

src/esm_tests/repos.py
```diff
@@ -10,6 +10,13 @@
 def update_resources_submodule(info, verbose=True):
     """
     Initiates and updates the module ``esm_tests_info``.
+
+    Parameters
+    ----------
+    info : dict
+        Dictionary with the general info about the tests.
```
pgierz, line 17: It may be nice to include some example of which info must be included. That can go either (or both) in the function body at the beginning of this function which throws "friendly" key errors, and/or in the doc-string.

src/esm_tests/repos.py
```diff
@@ -36,6 +43,24 @@ def update_resources_submodule(info, verbose=True):
...
+    install it sets ``info["repo_update"] = True``.
+
+    Parameters
+    ----------
+    info : dict
+        Dictionary with the general info about the tests.
+    verbose : bool
+        Verbose option.
+
+    Notes
```
pgierz, line 59: I am unsure how difficult this is, but it might be interesting to create "custom" Sphinx headings. @denizural is correct, this heading should not be `Raises` as we do not raise an error here, but it is an `Error` that might be usefully separated out in the documentation to show what sort of things break this function.

Personally I would just raise an exception, but users don't like long scary tracebacks...

pgierz (commented):
Some small comments and ideas, up to you to include them or not. To be honest, I would say: put this in, and we can separately include the clean-up, changes, and/or my ideas; as it is a valuable feature to have in CI as soon as possible to avoid future headaches.

Excellent work with the tests Miguel 👍🏻

src/esm_runscripts/namelists.py
pgierz: The above is also an outdated comment from @denizural `enum` suggestion.

pgierz:
A brief note: I am not sure why exactly, but it put comments of some ideas I had to @denizural notes in their twice. I marked those answers in my own review as redundant. Sorry for the confusion. GitHub confuses me sometimes, maybe I am slowly getting old...

src/esm_runscripts/namelists.py
denizural: Rather than using string functions we should use Pathlib or os.path methods. Eg. `parent`, ...

src/esm_tests/repos.py
denizural: We can always catch the exception with `user_error` to suppress the traceback & exit gracefully.

src/esm_tests/cli.py
mandresm: `ACCOUNT` should be removed here as it's not in used and shouldn't be used anymore.

src/esm_tests/repos.py
mandresm: `info` is a dictionary cotainning most of the information of the tests. It is build sequentially through the `cli.py`, starting with the loading of the `argpars` variables, and continuing with all the functions of esm_tests. It's similar to the `config` dictionary/object in ESM-Tools. Therefore, the description should be in a more general place, like in the `cli.py` right? Or maybe I just need to be build an `dict` subclass and put the docstrings there?

src/esm_runscripts/namelists.py
mandresm: Yeah, this is horrible... I'll find a better way to fix this.

src/esm_runscripts/namelists.py
mandresm: The minimum would have been to use regex, but I don't know what I was doing that day. In any case, I don't like this either. I will probably revert most of the `namelists.py` changes and fix it from the resources side (avoiding being so so invasive to the ESM-Tools code)

src/esm_tests/repos.py
pgierz: A dict subclass is a nice idea here. You can have it fail during initialisation if it does not have the proper parts.

src/esm_runscripts/namelists.py
pgierz: Assuming the number is always the last thing in the file name:
```suggestion
                    ib_num_new = int(iceberg_dir.split("_")[-1])
```

src/esm_tests/cli.py
mandresm: The user and computer specific parts have been generalized and included into a method.

src/esm_runscripts/namelists.py
mandresm: solved by reverting the changes in `namelist.py` and handling all throught esm_tests

mandresm:
I think all points have been tackled

src/esm_tests/info.py
```diff
@@ -0,0 +1,165 @@
...
+    def __init__(self):
+        self["script_dir"] = os.path.join(os.path.dirname(os.path.realpath(__file__)), ".")
+
+
+    def argparse(self):
+        """
+        Parses the arguments from the ``esm_tests`` call.
+
+        Returns
+        -------
```
pgierz, line 66:
```suggestion
        -------
         Returns a tuple with the following entries:
```

Check how that renders in the documentation.

pgierz (changes requested):

src/esm_tests/info.py
mandresm: The dosctring building in the docs is broken since long time, so I cannot check it. I tried a couple of times to try to fix it and failed.

mandresm:
After passing the black formatting I'll be ready to merge. I'll work on the documentation later, but at least every method/class has now detailed docstrings. I'll need an approval before merging ;)

mandresm:
#bump

mandresm:
#bump

Issue #670, opened 2022-05-19
ECHAM6-wiso in standalone mode: wrong file directories for wiso output and restart files


mwerner-awi:
When I run ECHAM6-wiso in standalone mode (and not as part of AWIESM-wiso), all *wiso* output and restart files are stored in the directory <EXP>/unknown instead of the directories <EXP>/output/[..] and <EXP>/restart/[..]. Therefore a simulation stops after the first job cycle as ECHAM6-wiso can't find the required wiso restart files in the expected directories.

To reproduce the bug, simply run echam6-wiso with namelist entries:

    add_namelist_changes:
            namelist.echam:
                wisoctl:
                    lwiso: True
                    lwiso_rerun: ${lresume}
                    nwiso: 3

(lresume is the default ECHAM6 switch for a cold start or restart.)

In the "move output" phase, the following files should be moved from <EXP>/run_xxxx/work:

to ECHAM output files:
<EXP>_XXX.01_wiso and <EXP>_XXX.01_wiso.codes
<EXP>_XXX.01_accw_wiso and <EXP>_XXX.01_accw_wiso.codes

to JSBACH output files:
<EXP>_XXX.01_js_wiso and <EXP>_XXX.01_js_wiso.codes
<EXP>_XXX.01_la_wiso and <EXP>_XXX.01_la_wiso.codes

to ECHAM restart files:
restart_<EXP>_<XXX>_wiso.nc 
restart_<EXP>_<XXX>_accw_wiso.nc 

to JSBACH restart files:
restart_<EXP>_<XXX>_js_wiso.nc 
restart_<EXP>_<XXX>_sf_wiso.nc 

This bug occurred on ollie.awi.de using esm_tools version 6.6.8.

pgierz:
Hi Martin,

thanks for reporting. It seems to be a relatively easy fix, the relevant output and restart files simply aren't known about:

https://github.com/esm-tools/esm_tools/blob/665ef93274c2efa41f27e5d5d901a585ad28ceb2/configs/components/echam/echam.yaml#L501-L505

Both the `wiso` and `accw_wiso` file tags are behind a switch which was introduced when we unified the AWIESM versions together.

Could you try to add the following in your runscript?
```yaml
echam:
   wiso_wiso: True
```

pgierz:
@mandresm, @denizural: someone should put this on the todo list or make a separate issue.

[...]

PR #690, opened 2022-06-10
Feature: add version control info for model and esm tools

### LINKED ISSUE #471: gitstatus of .venv_esmtools
Without a gitstatus or similar in an arbitrary `.venv_esmtools` directory, its hard to see which branches/tags/versions were used.

A gitstatus or similar of every bin in a `.venv_esmtools` dir.

pgierz:
This adds a new file with version control info for the models used and the esm-tools via the git commit hash

Closes #471


pgierz:
Still some minor things to come from actual testing on Ollie; therefore still a draft.

pgierz:
May be of interest to @chrisdane and @JanStreffing

mandresm:
I love this feature! I wanted to do it since so loong... I'll give my reviews later on.

mandresm:
Currently the yaml file generated looks like this:
```
esm_tools: <hash>
fesom: <hash>
```

I'd suggest that it comes with a more human readable information together with the hash:
```
fesom:
    hash:
    branch: <if it's checked out as a branch>
    tag: <if it's checked out as a tag>
    state: "dirty/clean/modified" (Modified by ESM-Tools via the SET thingy)
```

@pgierz let me know if you agreed and if you'd like me to take care of that

src/esm_runscripts/helpers.py
```diff
@@ -295,3 +296,60 @@ def def_path(self, path):
...
+
+    Raises
+    ------
+    GitDirtyError :
+       Raised if the repo is "dirty"; thus meaning that the hash is not meaningful
+    """
+    repo = git.Repo(path)
+    if repo.is_dirty():
+        raise GitDirtyError(
+            f"Your repo at {path} is dirty, thus the git hash is not meaningful!"
```
denizural, line 356: I am not sure if someone who is not very familiar with Git can understand what this `dirty` means. Maybe something like `uncommitted changes` could be better.

src/esm_runscripts/prepare.py
```diff
@@ -672,6 +674,50 @@ def set_parent_info(config):
...
+    config : dict
+        The experiment configuration
+
+    Returns
+    -------
+    config : dict
+        The experiment configuration
+    """
+    import pdb
+    pdb.set_trace()
```
denizural: Just a side not, Python on Levante supports `breakpoint()` command.

src/esm_runscripts/prepare.py
```diff
@@ -672,6 +674,50 @@ def set_parent_info(config):
...
+    experiment log directory for both the models and the esm-tools.
+
+    Parameters
+    ----------
+    config : dict
+        The experiment configuration
+
+    Returns
+    -------
+    config : dict
```
denizural, line 691: As the dicts are mutable there is no need to return them explicitly.

src/esm_runscripts/prepare.py
```diff
@@ -672,6 +674,50 @@ def set_parent_info(config):
...
+            try:
+                vcs_versions[model] = helpers.get_git_hash
+            except helpers.GitDirtyError:
+                vcs_versions[model] = "unknown (dirty repo detected)"
+
+    # NOTE(PG): There is no good way to get the repo directory from the config,
+    # this may at least be a good start:
+    esm_tools_repo = config.get("general", {}).get("esm_function_dir")
+    if esm_tools_repo is not None:
+        vcs_versions["esm_tools"] = helpers.get_git_hash(f"{esm_tools_repo}/../")
```
denizural, line 710: As I wrote to the function definition, this function raises and exception and it will terminate the flow. Maybe it needs to be placed in a try-catch block like above.

mandresm:
Also, now that we can obtain versions of the models on the fly, we can also detect automatically whether the users are running the code with an incorrect version in the yaml (that's part of the problem in issue #670, he defined the version in `general`, not in `echam` which means ESM-Tools was thinking it was running `echam-6.3.01p4`, the default). I'd love to take care of that check, as I have a runscript to test against it...

src/esm_runscripts/prepare.py
```diff
@@ -672,6 +674,50 @@ def set_parent_info(config):
...
+    pdb.set_trace()
+    with open(exp_vcs_info_file, "w") as f:
+        yaml.dump(vcs_versions, f)
+    pdb.set_trace()
+    return config
```
denizural, line 759: no need to return it explicitly.

src/esm_runscripts/prepare.py
```diff
@@ -672,6 +674,50 @@ def set_parent_info(config):
...
+
+
 def finalize_config(config):
     config.finalize()
     return config
```
denizural, line 764: no need to return it explicitly.

src/esm_runscripts/prepare.py
pgierz: Every function we have explicitly returns config. Not doing it here would break the recipe manager

denizural: Yes, this is the problem. They were all wrong from the beginning 🤣 . Well, not wrong but let's say misleading.

src/esm_runscripts/prepare.py
```diff
@@ -672,6 +674,89 @@ def set_parent_info(config):
...
+    logger.debug("Experiment information is being stored for usage under:")
+    logger.debug(f">>> {exp_vcs_info_file}")
+    vcs_versions = {}
+    all_models = config.get("general", {}).get("models", [])
+    for model in all_models:
+        model_dir = config[model]["model_dir"]
+        if helpers.is_git_repo(model_dir):
+            vcs_versions[model] = helpers.get_all_git_info(model_dir)
+        else:
+            vcs_versions[model] = "Not a git-controlled model!"
```
denizural, line 704: I hope no one is using a SVN based system. We support it but this will require extra refactoring 🥲

src/esm_runscripts/prepare.py
```diff
@@ -672,6 +674,89 @@ def set_parent_info(config):
...
+        return config  # No check needed on the very first run
+    exp_vcs_info_file = f"{config['general']['thisrun_log_dir']}/{config['general']['expid']}_vcs_info.yaml"
+    # FIXME(PG): This file might not exist if people erase every run folder...
+    last_exp_vcs_info_file = f"{config['prev_run']['general']['thisrun_log_dir']}/{config['general']['expid']}_vcs_info.yaml"
+
+    with open(exp_vcs_info_file, "r") as f:
+        current_vcs_info = yaml.safe_load(f)
+    with open(last_exp_vcs_info_file, "r") as f:
+        last_vcs_info = yaml.safe_load(f)
+    if not config["general"].get("allow_vcs_differences", False) and current_vcs_info != last_vcs_info:
```
denizural, line 751: @pgierz, I am not 100% sure but maybe library like deepdiff would be better here. It looks like `vcs_info` dicts are simple so that basic equality check should not be a big issue here.

denizural (approved):
Just added few comments there. 👏 👏

src/esm_runscripts/prepare.py
pgierz: Let's wait for someone to ask for SVN...none of the models I actively support uses anything aside from git.

src/esm_runscripts/prepare.py
pgierz: At the moment is a dictionary of dictionaries, nested only once. In any case, as soon as differences are detected at all, it should break, which this currently does. The only "valid" differences would be if someone goes in and changes the file by hand, without changing the actually relevant info. I currently don't think anyone would be that stupid...but, you never know. The question then becomes, how should I check for that?

pgierz:
#bump

mandresm:
there are conflicts in this merge that need to be resolved before merging

pgierz:
I just fixed them, so it should be ready now.

mandresm:
there you go, you were faster XD

pgierz:
Why did it not #bump ...? Git robots, do your thing...

mandresm:
because it could not merge `release` into the branch due to the conflict, so it won't bump unless all conflicts are resolved :)

PR #693, opened 2022-06-12
fix compilation flags in rnfmap (only affecting FOCI-OpenIFS)

mandresm:
The file `configs/components/rnfmap/rnfmap.env.yaml` contained the compilation flags for `levante` (`-march` and `-mtune`) and was applying it to all computers (including `blogin`). After this fix, the default will be the standard, and `-march` and `-mtune` will only be used in `levante`. This only affects to FOCI-OpenIFS, because AWICM3 uses a common environment for all components based on `oifs.env.yaml`, and there, the levante flags are already behind a choose block for levante.


mandresm:
Tested without errors

denizural (approved):
👍

mandresm:
#bump

PR #703, opened 2022-06-25
Fix/esm tests group output GitHub

mandresm:
It should make the output for `esm_tests` in CI more readable by adding github actions grouping labels.


pgierz (approved):
This looks like just print formatting, yes? At some point I will want to dig into `src` of `esm_tests` so I can give good reviews...

But for now, thumbs up 👍🏻

mandresm:
Look at the ollie tests [here](https://github.com/esm-tools/esm_tools/runs/7053770082?check_suite_focus=true#step:6:50) there you'll see that the differences are collapsed so that is way easier to look at one at a time.

pgierz:
Ah, so these clickable arrow things are the new feature? then thumbs up!!! excellent!

pgierz:
Bump and put in, once we decide how to bump CI things. I would say quietly in the background without new numbers

mandresm:
#bump

PR #704, opened 2022-06-27
fixes wrong oasis branch name

denizural:
Oasis branch names are wrong. Some of the branch names does not exist in the remote repositories. Eg. 
```shell
$ git ls-remote --exit-code --heads https://gitlab.dkrz.de/modular_esm/oasis3-mct.git 2.8
$ echo $?
2

$ git ls-remote --exit-code --heads https://gitlab.dkrz.de/modular_esm/oasis3-mct.git v2.8
$ echo $?
0
```

This will lead to **buggy** behavior. Eg.
```shell
$ esm_master get-oasis3mct-2.8
...
    You are in 'detached HEAD' state. You can look around, make experimental
    changes and commit them, and you can discard any commits you make in this
    state without impacting any branches by switching back to a branch.
...

$ git status
Not currently on any branch.
```

After the bugfix:
```shell
$ esm_master get-oasis3mct-2.8
# no warning messages

$ git status
On branch v2.8
Your branch is up to date with 'origin/v2.8'.
```

Here are what I found:
- Branch name 2.8 should be **2.8**. This is what I changed in this commit.
- **3.0** is missing in both:
    - https://gitlab.awi.de/paleodyn/Models/oasis3-mct/-/branches/all
    - https://gitlab.dkrz.de/modular_esm/oasis3-mct/-/branches/all
-  **4.0** is synced across DKRZ and AWI
- **4.0-awicm-3.0** is missing but we have **awicm-3-frontiers** which are synced in DKRZ and AWI
- Branch **4.0-awicm-3.1** is missing
- Branch **4.0-awicm-frontiers** is missing
- **ec-earth** version is OK.
- Did not check the GEOMAR versions since I don't have any access.


configs/components/oasis3mct/oasis3mct.yaml
```diff
@@ -5,6 +5,7 @@ time_step: 900 # changed by coupled setup anyway
 norestart: F
 use_lucia: False
 mct_version: "2.8" # TODO: set mct_version = version (can't do at the moment as version is a string)
+git-repository: https://gitlab.dkrz.de/modular_esm/oasis3-mct.git
```
denizural, line 8: Moved it here because it is more clear and friendlier to keep the default version and the repository together as they are closely related.

configs/components/oasis3mct/oasis3mct.yaml
```diff
@@ -34,7 +35,7 @@ choose_version:
     branch: '2.8mct-awiesm-2.1'
     git-repository: https://gitlab.awi.de/paleodyn/Models/oasis3-mct.git
   '2.8':
-    branch: '2.8'
+    branch: 'v2.8'
```
denizural, line 38: As explained in the PR message, we need to remove the obsolete branches and fix the missing ones.

mandresm (approved):
`2.8` exist as a tag, so it indeed exist, and it does what it suppose to do. The detached message is what you would expect from git when you pull a tag. However, I agree it can add confusion, and in the end it changes nothing so I am okay with it.

mandresm:
By the way, `3.0` is not missing in the repos, is just another tag

denizural:
Ok, I also verified that -b takes tags too. Corresponding yaml variable is branch so it can be confusing. Git doc also says that in the case of tag, it detaches the head. I though that -b was only cloning the branch.

mandresm:
can you go ahead an use the bump command and merge

denizural:
#bump

PR #712, opened 2022-06-30
feat/yaml warning and error

### LINKED ISSUE #696: Trigger warnings from yamls
Useful to report warnings to users under certain configurations. For example in the case of #695, if in the future, a new `LU_trans_1850` is set as default for PI simulations, the user needs to be notified about it, so that she/he can decide whether that's okay, or wants to continue an older run with the old file.

One could use the following syntax (or something similar) in yaml files to trigger a user warning:
```yaml
warning:
    <warning_name>: # Name for the specific warning
        message: "the message of the warning"
        esm_tools_version: ">/</=/!=/version_number" # trigger it under certain ESM-Tools version conditions
        ask_user_to_continue: True/False # Ask user about continuing or stopping the process
```

This could be included inside `choose_` blocks, which means it can be triggered under certain conditions/configurations.

The warning would be displayed when submitting the simulation (and the conditions are met) and would look something like this:
```
Warning: <warning_name>:
---------------------------
<message>

[optional] Do you want to continue (set general.ignore_yaml_warnings: False to avoid quesitoning)? (y/n): 
```

**Additional context**
I'd like to take into account the feedback from the advance users for this feature before jumping into coding, so please reply to this issue with your ideas and feel free to tag users I forgot about (@pgierz, @denizural, @JanStreffing, @joakimkjellsson, @christian-stepanek, @chrisdane, @fernandadialzira, @tsemmler05)

mandresm:
Closes #696 

- One can use the following syntax yaml files to trigger a user warning or an error:
  ```yaml
  warning/error:
      <name>: # Name for the specific warning or error
          message: "the message of the warning/error"
          esm_tools_version: ">/</=/!=/version_number" # trigger it under certain ESM-Tools version conditions
          ask_user_to_continue: True/False # Ask user about continuing or stopping the process, only for warnings, errors always kill the process
  ```

- This syntax can be included in `choose_` blocks, which means it can be triggered under certain conditions/configurations.

- The warning/error are displayed when submitting the simulation (and the conditions are met) and look something like this:
  ```
  Warning: <warning_name>:
  ---------------------------
  <message>

  [optional] Do you want to continue (set general.ignore_yaml_warnings: False to avoid quesitoning)? (y/n): 
  ```

- One can include multiple warnings/errors by expanding the `warning`/`error` dictionary with `add_`.

- Defining `general.ignore_config_warnings: True` in the runscript or adding the ``--ignore-config-warnings`` flag to the `esm_runscripts` call still prints the warnings but don't stop at those where `ask_user_to_continue` is set to `True`.

Note: for this to work it was necessary to bring back from the death the `--last-jobtype` flag, from which it is possible to understand whether the `esm_runcstips` is called from the command line and therefore. If it's called from the command line it's considered `isinteractive` meaning that we can ask users they want to continue or not. When `isinteractive==False` then that's in the middle of a compute job, and we don't want to ask for input to the users because that will break the run.

**TODO**

- [x] Test
- [x] Documentation
- [x] Docstrings
- [x] Commenting
- [x] Take the version checker function out of `esm_motd` and put it in a more general package.


mandresm:
A case example:
```yaml
recom:
    choose_scenario:
        HIST:
            [ ... ]
        PI-CTRL:
            [ ... ]
        "*":
            add_warning:
                "wrong scenario type":
                    message: "The scenario you specified (``${recom.scenario}``) is not supported!"
                    ask_user_to_continue: True
```

If you define `recom.scenario: hist` instead of `HIST` then you'll get the following:

Screen Shot 2022-06-30 at 20 46 43

src/esm_utilities/utils.py
```diff
@@ -2,6 +2,9 @@
 import io
 import datetime
 import numpy as np
+import semver
+
+import esm_tools
```
pgierz, line 7: At some point I want to have a unified `_version.py` file so that we don't need to keep the version numbers in each sub-project separately. But that is for later...

src/esm_utilities/utils.py
```diff
@@ -99,3 +102,62 @@ def logfile_stats(logfile_to_read):
         f"Theoretical Throughput assuming no queueing time: {np.round(throughput, 2)} runs per day"
     )
     return df
+
+
+def check_valid_version(versionrange, version=False):
```
pgierz: `version=""` to be in line with the docstring. An empty string also evaluates to `False` in `if` statements.

src/esm_motd/esm_motd.py
```diff
@@ -41,62 +41,6 @@ def __init__(self):
         self.database_connected = True
         self.message_dict = yaml.load(self.motdfile, Loader=yaml.FullLoader)
 
-    def check_valid_version(self, version, versionrange):
```
pgierz, line 44: Did I originally write this one...?

src/esm_utilities/utils.py
```diff
@@ -99,3 +102,62 @@ def logfile_stats(logfile_to_read):
...
+    """
+    Returns ``True`` if the ``version`` provided matches the condition of
+    ``versionrange``.
+
+    Parameters
+    ----------
+    version : str
```
pgierz, line 114: I think this is still the old docstring from before, and might not be accurate as it is...?

src/esm_runscripts/prev_run.py
```diff
@@ -197,9 +197,7 @@ def prev_run_config(self):
         )
         # Check for interactive, or submitted from a computing node, to avoid
         # using ``input()`` or ``questionaries`` in the second case
-        self.isinteractive = (
-            self._config["general"].get("last_jobtype", "") == "command_line"
-        )
+        self.isinteractive = self._config["general"].get("isinteractive", False)
```
pgierz, line 200: I don't understand the old implementation, this is much better 👍🏻

src/esm_runscripts/prepexp.py
```diff
@@ -168,9 +168,12 @@ def copy_tools_to_thisrun(config):
...
         restart_command = f"cd {scriptsdir}; esm_runscripts {new_command}"
 
-        # prevent continuous addition of --no-motd
-        if not "--no-motd" in restart_command:
-            restart_command += " --no-motd "
+        # Add non-interaction flags
+        non_interaction_flags = ["--no-motd", f"--last-jobtype {config['general']['jobtype']}"]
+        for ni_flag in non_interaction_flags:
+            # prevent continuous addition of ``ni_flag``
+            if not ni_flag in restart_command:
```
pgierz:
```suggestion
            if ni_flag not in restart_command:
```

src/esm_runscripts/prepare.py
```diff
@@ -729,3 +730,60 @@ def set_logfile(config):
...
+        triggers["warning"] = {"note_function": esm_parser.user_note}
+
+    for trigger, trigger_info in triggers.items():
+        warn_error(config, trigger, trigger_info["note_function"])
+
+    return config
+
+def warn_error(config, trigger, note_function):
+
+    if trigger=="warning":
```
pgierz, line 834:
```suggestion
    if trigger is "warning":
```

src/esm_runscripts/prepare.py
```diff
@@ -729,3 +730,60 @@ def set_logfile(config):
...
+
+                    if (
+                        trigger=="warning"
+                        and config["general"].get("isinteractive")
+                        and action_info.get("ask_user_to_continue", False)
+                        and not config["general"].get("ignore_config_warnings", False)
+                    ):
+                        user_reply = ""
+                        while user_reply != "y" or user_reply != "n":
+                            user_reply = input(
```
pgierz: You could use questionary here if you want

src/esm_runscripts/prepare.py
```diff
@@ -729,3 +730,60 @@ def set_logfile(config):
...
+            for action_name, action_info in actions.items():
+                version_condition = action_info.get("esm_tools_version", ">0.0.0")
+                if esm_utilities.check_valid_version(version_condition):
+                    note_function(
+                        f"{action_name}{sufix_name}",
+                        f'Section: ``{section}``\n\n{action_info.get("message", "")}'
+                    )
+
+                    if (
+                        trigger=="warning"
```
pgierz, line 861: replace `==` with `is`

src/esm_runscripts/prepare.py
```diff
@@ -729,3 +730,60 @@ def set_logfile(config):
...
+    # Only warn if it is an interactive session or while submitted
+    if not isresubmitted or isinteractive:
+        triggers["warning"] = {"note_function": esm_parser.user_note}
+
+    for trigger, trigger_info in triggers.items():
+        warn_error(config, trigger, trigger_info["note_function"])
+
+    return config
+
+def warn_error(config, trigger, note_function):
```
pgierz, line 776: It would be nice to include in the docstring a minimal example of what would be written in the config, and what the terminal output would look like.

src/esm_runscripts/prepare.py
```diff
@@ -729,3 +730,60 @@ def set_logfile(config):
...
+
+    if trigger=="warning":
+        sufix_name = f" WARNING"
+    else:
+        sufix_name = f""
+
+    for section, value in config.items():
```
pgierz, line 840: Does this go far enough down into nested sections of the config?

pgierz (commented):
Looks good for the first draft! 👍🏻 I have a few cleanup comments you might consider.

pgierz:
> A case example:
> 
> ```yaml
> recom:
>     choose_scenario:
>         HIST:
>             [ ... ]
>         PI-CTRL:
>             [ ... ]
>         "*":
>             add_warning:
>                 "wrong scenario type":
>                     message: "The scenario you specified (``${recom.scenario}``) is not supported!"
>                     ask_user_to_continue: True
> ```
> 
> If you define `recom.scenario: hist` instead of `HIST` then you'll get the following:
> 
> Screen Shot 2022-06-30 at 20 46 43

This is exactly what I wanted in the docstring while reading the code 😉

src/esm_runscripts/prepare.py
denizural: I disagree with that one. Strings are immutable and it is better to use `==` for comparison. `is` operator is used for checking if two objects are identical.

src/esm_runscripts/prepare.py
pgierz: Alright, disregard this suggestion:
```
Python 3.9.5 (default, May 18 2021, 12:31:01)
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> s = "mystring"
>>> s == "mystring"
True
>>> s is "mystring"
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
True
```

src/esm_runscripts/prepare.py
pgierz: This one can be disregarded as well:
```
Python 3.9.5 (default, May 18 2021, 12:31:01)
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> s = "mystring"
>>> s == "mystring"
True
>>> s is "mystring"
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
True
```

I find it easier to ready, but that `SyntaxWarning` will probably pop up in the end user view, and it might be that they at some point make this style of comparisons for strings illegal.

denizural: Yes, exactly. This is a new feature in Python. When you try to compare immutable types with `is` it tells you that you are doing it wrong. I already fixed the previous ones in the parser.

src/esm_motd/esm_motd.py
mandresm: I think so, it's a pretty tough function, never gave a problem ;)

src/esm_runscripts/prepare.py
mandresm: Yeah, I remember changing my habits a year ago or so and starting to use `==` instead of `is` for this reason.

src/esm_runscripts/prepare.py
mandresm: No, this goes only into choose_blocks, because when they are resolved their content get's into the first indentation. I've done it this way, because it's easier and most importantly, because all our syntax works this way.

src/esm_runscripts/prepare.py
denizural: Well, `is` is still needed since it compares the `id()` of the objects. Whereas, `==` calls the `.__eq__()` method of that object.

denizural (approved):
👏 👏

src/esm_utilities/utils.py
mandresm: It is still accurate :)

denizural:
Hi everyone,
I also coded some consistency checking functionality (WIP) but did not push 🤦‍♂️ . Upon discussing with Miguel, I pushed it and maybe we can combine them together.
https://github.com/esm-tools/esm_tools/tree/hotfix/report_missing_scenario
The main idea was having a thin wrapper around multiple consistency checks (eg. versions, scenarios, ...) before the job is submitted.

pgierz:
Deniz, push more often!! Even if it is WIP, you can mark it as a draft PR, and then we avoid double work in the future 😉

mandresm:
I think is ready to go, I think the "wish-list" of @pgierz is covered ;)

src/esm_runscripts/prepare.py
```diff
@@ -733,7 +734,28 @@ def set_logfile(config):
 
 
 def check_config_for_warnings_errors(config):
+    """
+    Wrapper method to call ``warn_error`` from the `ESM-Tools` recipes. Loops through
+    the trigger variables to call ``warn_error``, only including ``warning`` in the
+    ``triggers`` if it is an interactive session (i.e. user just submitted a job), or
+    if it's not a resubmission (to avoid displaying the warning twice).
+
+    .. note:: use this in the recipe after the ``choose_`` blocks are resolved,
```
pgierz, line 743: I have a silly idea....can we check for this? The `recipe` in in `config` after all. I will make a separate issue with some ideas about sanity checks for recipe order.

pgierz (approved):
Hooray! 👍🏻 

Reading through the docstrings gave me an idea, I will maybe put together some ideas for that in a separate issue.

mandresm:
All `esm_tests` passed so... #bump!

PR #714, opened 2022-07-01
[WIP] adds features to check the consistency of the config before sub…

denizural:
This is a work in process. Not meant to be merged right now but to be combined with 
https://github.com/esm-tools/esm_tools/pull/712

### DIFF
--- a/src/esm_runscripts/config_control.py
+++ b/src/esm_runscripts/config_control.py
@@ -1,0 +1,112 @@
+import yaml
+import esm_parser
+
+
+def check_configuration_consistency(config):
+    """ TODO: fill docstring
+    """
+    runscript_abspath = config["general"]["command_line_config"]["runscript_abspath"]
+    # TODO: can this be None or KeyError?
+    models = config['general']['valid_model_names']
+
+    with open(runscript_abspath, "rt") as f:
+        runscript = yaml.load(f, Loader=yaml.FullLoader)
+
+    for model in models:
+        model_config_path = config[model]["debug_info"]["loaded_from_file"]
+        # some models have multiple files. Take the one with <model>.yaml
+        if isinstance(model_config_path, list):
+            model_config_path = model_config_path[0]
+            # deniz: remaining lines was a good idea but don't always work
+            # base_name = f"{model}.yaml"
+            # model_config_path = next((item for item in model_config_path if base_name in item), None)
+        with open(model_config_path, "rt") as f:
+            model_config = yaml.load(f, Loader=yaml.FullLoader)
+
+        check_version(config, model_config, runscript, model)
+        check_scenario(config, model_config, runscript, model)
+
+
+def check_version(config, model_config, runscript, model):
+    """TODO: fill docstring
+    """
+    # model is not specified in the runscript. Using the default
+    if model not in runscript:
+        return True
+    #TODO: what about in coupled setups?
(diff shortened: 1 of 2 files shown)


src/esm_runscripts/config_control.py
```diff
@@ -0,0 +1,112 @@
...
+        return True
+
+    # exit if requested scenario is not found
+    if scenario_from_runscript not in available_scenarios:
+        error_type = "Scenario not available"
+        error_msg = f"Scenario '{scenario_from_runscript}' is not available for model {model}"
+        esm_parser.user_error(error_type, error_msg)
+
+    # scenario requested in the runscript is found in the model configuration
+    return True
```
mandresm, line 108: As I mentioned in our talk, `scenario` is not a ESM-Tools featured variable, meaning that it means nothing to ESM-Tools and should not be handled here, but from the YAML files, with the "yaml warnings/errors". In turn, I agree with the checks of the `version` as it is a featured variable in ESM-Tools.

src/esm_runscripts/filelists.py
```diff
@@ -1317,6 +1318,7 @@ def get_movement(config, model, categ, filetype, source, target):
 
 
 def assemble(config):
+    config_control.check_configuration_consistency(config)
```
mandresm, line 1321: I wouldn't hook this here (in the `filelists.py`) but in the recipe instead.

src/esm_runscripts/config_control.py
```diff
@@ -0,0 +1,112 @@
...
+    models = config['general']['valid_model_names']
+
+    with open(runscript_abspath, "rt") as f:
+        runscript = yaml.load(f, Loader=yaml.FullLoader)
+
+    for model in models:
+        model_config_path = config[model]["debug_info"]["loaded_from_file"]
+        # some models have multiple files. Take the one with <model>.yaml
+        if isinstance(model_config_path, list):
+            model_config_path = model_config_path[0]
```
pgierz, line 19: Maybe not the best check against index of 0. Models normally implement further reading, so those will be behind the actual model yaml in this list, yet this might not always be the case

src/esm_runscripts/config_control.py
```diff
@@ -0,0 +1,112 @@
...
+            # deniz: remaining lines was a good idea but don't always work
+            # base_name = f"{model}.yaml"
+            # model_config_path = next((item for item in model_config_path if base_name in item), None)
+        with open(model_config_path, "rt") as f:
+            model_config = yaml.load(f, Loader=yaml.FullLoader)
+
+        check_version(config, model_config, runscript, model)
```
pgierz, line 26: What happens with the return values? In your functions, you give back true but that is never used.

src/esm_runscripts/config_control.py
```diff
@@ -0,0 +1,112 @@
...
+        check_scenario(config, model_config, runscript, model)
+
+
+def check_version(config, model_config, runscript, model):
```
pgierz, line 30: These can both be simplified by only checking for the incorrect case and then failing, or, if you to break out of "good" cases, have the fail be the default ending behaviour of the function, and return none in a good case. Also be careful; the order of your checks might end up being important.

pgierz (commented):
I'll give some more comments on my train back to Bremen tomorrow

pgierz:
As #712 went in, can this one be closed? If not, please re-open!

PR #716, opened 2022-07-04
Feat/merging lists

denizural:
This bugifx will solve the issue: https://github.com/esm-tools/esm_tools/issues/667 and it is an updated version of the PR https://github.com/esm-tools/esm_tools/pull/672 based on the reviews from @mandresm and @pgierz and upon discussion with Christopher.

In this updated version, usage of YAML anchor and reference operators are dropped and the solution is implemented solely on ESM-Tools YAML syntax.

An inline example usage of the new parser function is also added to make @pgierz happy.

Note that this solution assumes that we will not support nested list additions as we discussed before. So, all lists will be flattened before merging. Please look at the examples in the docstring for further explanation.

It supports full ESM-Tools features like resolving variables via `${}` syntax, referencing to other sections with the dot syntax and referencing external files (`further_reading`). Here is an example. Add the following lines on eg. `general` section of your favorite runscript:
```yaml
foo:
    bar: 99
general:
    ...
    mylist1: [1,2,3]
    mylist2:
    - 50
    - 51
    - "${foo.bar}"                # reference to a variable in another section

    add_mylist1:
    - [20, 21, 22]                # simple list
    - [[31, 32, 33], [41,42,43]]  # nested list
    - "${fesom.dummy_list}"       # comes from further_reading file
    - "${mylist2}"                # comes from another section

fesom:
    ...
    further_reading: "dummy.yaml"
```

Contents of the external file `dummy.yaml`:
```yaml
dummy_info: "I come from dummy.yaml"
dummy_list:
- 60
- 61
- I came from mylist3 in dummy.yaml
``` 

The result in the `finished_config` will be:
```yaml
mylist1:
- 1
- 2
- 3   # these three are the initial mylist1
- 20  # first three add_mylist1 additions
- 21
- 22
- 31  # next six elements are the nested list added. Note that they are flattened before merging
- 32
- 33
- 41
- 42
- 43
- 60   # next three items come from an external file
- 61
- I came from mylist3 in dummy.yaml
- 50   # mylist2 is added
- 51
- 99   # note that this variable in mylist2 comes from another section (foo.bar)
```

### DIFF
--- a/src/esm_parser/esm_parser.py
+++ b/src/esm_parser/esm_parser.py
@@ -139,6 +139,7 @@ early_choose_vars = ["include_models", "version", "omp_num_threads"]
 if six.PY2:  # pragma: no cover
     FileNotFoundError = IOError
 
+
 def flatten_nested_lists(lst):
     """Recursively flattens an arbitrarily nested list and yields a generator
 
@@ -146,13 +147,13 @@ def flatten_nested_lists(lst):
     --------
     >>> list(flatten_nested_lists( [[1,2,3]] ))
     [1, 2, 3]
-    
+
     >>> list(flatten_nested_lists( [1,2,3, [4,5,6], "foo"] ))
     [1, 2, 3, 4, 5, 6, 'foo']
-    
+
     >>> list(flatten_nested_lists( [[1,2,3], [4,5,6], [7,8,9]] ))
     [1, 2, 3, 4, 5, 6, 7, 8, 9]
-    
+
     >>> list(flatten_nested_lists( [[1,2,3], {"foo": "bar"}] ))
     [1, 2, 3, {'foo': 'bar'}]
     """
@@ -1471,7 +1472,7 @@ def resolve_basic_choose(config, config_to_replace_in, choose_key, blackdict={})
         user_error(
             "choose_ block",
             "``choose_`` blocks need to be defined as ``dictionaries``. Currently, "
-            f"``{choose_key}`` is of type ``{type(config_to_replace_in[choose_key])}``"
+            f"``{choose_key}`` is of type ``{type(config_to_replace_in[choose_key])}``",
         )
     for ckey, cval in config_to_replace_in.get(choose_key, {}).items():
         if (
@@ -1496,9 +1497,7 @@ def resolve_basic_choose(config, config_to_replace_in, choose_key, blackdict={})
 
     # Resolve the choose variables
     if choice in choices_available:
-        for update_key, update_value in six.iteritems(
-            choices_available[choice]
-        ):
+        for update_key, update_value in six.iteritems(choices_available[choice]):
             deep_update(update_key, update_value, config_to_replace_in, blackdict)
 
     elif "*" in config_to_replace_in.get(choose_key):
@@ -1968,9 +1967,8 @@ def find_variable(tree, rhs, full_config, white_or_black_list, isblacklist):
             # preceding (``prefix``) or following (``suffix``) string, then add up
             # the parts, making sure that other variables (``${}``) are also
             # substitute. The "NONE_YET" part is to handle PrevRunInfo class correctly
-            if (
-                (not isinstance(var_result, list) and (prefix or suffix)) or
-                (isinstance(var_result, dict) and "NONE_YET" in var_result)
+            if (not isinstance(var_result, list) and (prefix or suffix)) or (
+                isinstance(var_result, dict) and "NONE_YET" in var_result
             ):
                 prefix, var_result, more_rest = (
                     str(prefix),
@@ -3085,7 +3083,7 @@ class ConfigSetup(GeneralConfig):  # pragma: no cover
             user_config = setup_config = self
         if (
             setup_config["general"].get("standalone")
-            and user_config["general"].get("run_or_compile", "runtime")=="runtime"
+            and user_config["general"].get("run_or_compile", "runtime") == "runtime"
         ):
             version_in_runscript_general = user_config["general"].get("version")
             model_name = user_config["general"]["setup_name"]
@@ -3096,7 +3094,7 @@ class ConfigSetup(GeneralConfig):  # pragma: no cover
                     "You have defined the ``version`` variable both in the "
                     f"``general`` and ``{model_name}`` sections of your runscript. "
                     "This is not supported for ``standalone`` simulations. Please "
-                    "define ``only one version`` in one of the two sections."
+                    "define ``only one version`` in one of the two sections.",
                 )
             elif version_in_runscript_general and not version_in_runscript_model:
                 if model_name in user_config:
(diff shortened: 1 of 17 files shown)


src/esm_parser/esm_parser.py
```diff
@@ -71,11 +71,11 @@
 import subprocess
 import sys
 import warnings
-import numpy
```
pgierz, line 74: In conjunction with `black`, we could run `isort` which will automatically sort out imports.

src/esm_parser/esm_parser.py
```diff
@@ -1038,7 +1065,10 @@ def add_entry_to_chapter(
                     add_chapter.split(".")[-1].replace("add_", "")
                 ]
                 # Add the entries
-                mod_list += add_entries
+                if isinstance(add_entries, list):
+                    mod_list.extend(list(flatten_nested_lists(add_entries)))
+                else:
+                    mod_list.append(add_entries)
```
pgierz: Not sure this is needed...? Isn't `mod_list` already defined behind a `isinstance` check against lists? From my point of view, this only needs to be:
```suggestion
                    mod_list.extend(list(flatten_nested_lists(add_entries)))
```

src/esm_parser/esm_parser.py
```diff
@@ -1467,9 +1497,7 @@ def resolve_basic_choose(config, config_to_replace_in, choose_key, blackdict={})
 
     # Resolve the choose variables
     if choice in choices_available:
-        for update_key, update_value in six.iteritems(
-            choices_available[choice]
-        ):
+        for update_key, update_value in six.iteritems(choices_available[choice]):
```
pgierz: We should start cleaning up and remove the `six` stuff. Nothing uses Python 2 anymore.

```suggestion
        for update_key, update_value in choices_available[choice].items():
```

pgierz (changes requested):
Looks good! 👍🏻  A few minor suggestions.

I'd also be curious if someone could check if the examples properly show up in `pytest` with the `doctest`

mandresm (changes requested):
Looks great! does one single thing, cleanly, keeping the syntax the users are used to, and it's robust. My only comment is that we might not want to flatten recursively the lists, because there are instances, like the FESOM-1 outputs, where we need nested lists and adding nested lists might be useful. But I'd like to get the feedback from all of you.

Can you also add a brief note in the documentation of the `add_` feature? https://esm-tools.readthedocs.io/en/latest/yaml.html#append-to-an-existing-list-add

src/esm_parser/esm_parser.py
denizural: Yes, I am also implementing pytest.

pgierz: You don't need to implement `pytest`, it is already there. See here: 
https://github.com/esm-tools/esm_tools/blob/781c659c912c1bdefaecdff15e564b79c4a9e184/.github/workflows/esm_tools_actions.yml#L35-L36

pgierz:
For doctest:
```
$ pytest --doctest-modules --verbose src/esm_parser
============================================================ test session starts ============================================================
platform darwin -- Python 3.9.5, pytest-7.1.2, pluggy-1.0.0 -- /private/tmp/esm_tools/.direnv/python-3.9.5/bin/python
cachedir: .pytest_cache
rootdir: /private/tmp/esm_tools
collected 2 items

src/esm_parser/esm_parser.py::esm_parser.esm_parser.attach_to_config_and_reduce_keyword PASSED                                        [ 50%]
src/esm_parser/esm_parser.py::esm_parser.esm_parser.flatten_nested_lists PASSED                                                       [100%]

============================================================= 2 passed in 0.45s =============================================================
```

Excellent work @denizural with the examples! Examples now are automatically tests and we can use them to make sure our code does what we think, without having to deal with all the actual unit-test stuff if you do not want to.

pgierz:
And importantly, stuff like this results in failure. I broke one of the tests to ensure that it was doing it correctly.

```diff
>>> list(flatten_nested_lists( [[1,2,3], {"foo": "bar"}] ))  
-[1, 2, 3, {'foo': 'bar'}]  
+[1, 2, 3, {'foo': 'bar'}, 4, 5, 6]
```

So, the written down "expected answer" is not what the code produces. That results in:
```
pytest --doctest-modules --verbose src/esm_parser
=================================================================================================================================== test session starts ===================================================================================================================================
platform darwin -- Python 3.9.5, pytest-7.1.2, pluggy-1.0.0 -- /private/tmp/esm_tools/.direnv/python-3.9.5/bin/python
cachedir: .pytest_cache
rootdir: /private/tmp/esm_tools
collected 2 items

src/esm_parser/esm_parser.py::esm_parser.esm_parser.attach_to_config_and_reduce_keyword PASSED                                                                                                                                                                                      [ 50%]
src/esm_parser/esm_parser.py::esm_parser.esm_parser.flatten_nested_lists FAILED                                                                                                                                                                                                     [100%]

======================================================================================================================================== FAILURES =========================================================================================================================================
__________________________________________________________________________________________________________________ [doctest] esm_parser.esm_parser.flatten_nested_lists ___________________________________________________________________________________________________________________
148     >>> list(flatten_nested_lists( [[1,2,3]] ))
149     [1, 2, 3]
150
151     >>> list(flatten_nested_lists( [1,2,3, [4,5,6], "foo"] ))
152     [1, 2, 3, 4, 5, 6, 'foo']
153
154     >>> list(flatten_nested_lists( [[1,2,3], [4,5,6], [7,8,9]] ))
155     [1, 2, 3, 4, 5, 6, 7, 8, 9]
156
157     >>> list(flatten_nested_lists( [[1,2,3], {"foo": "bar"}] ))
Expected:
    [1, 2, 3, {'foo': 'bar'}, 4, 5, 6]
Got:
    [1, 2, 3, {'foo': 'bar'}]

/private/tmp/esm_tools/src/esm_parser/esm_parser.py:157: DocTestFailure
================================================================================================================================= short test summary info =================================================================================================================================
FAILED src/esm_parser/esm_parser.py::esm_parser.esm_parser.flatten_nested_lists
=============================================================================================================================== 1 failed, 1 passed in 0.61s ===============================================================================================================================

```

src/esm_parser/esm_parser.py
mandresm: I think `mod_list` is the list into which you add, which is already under a `isinstance`, but `add_entries` is what you want to add to that first list, and that could be a string or a list to be flattened, so I think the whole if and else is needed here

denizural: Yes, I controlled that and you are right. I removed the redundant checks and only kept the extend. Moreover, I added better error handling by `user_error`. If an uncompatible type is tried to merge now we will have:
```
ERROR: Type mismatch
--------------------
Can not add a variable of incompatible type int to mylist3
```
instead of a long traceback with `TypeError: Something is wrong` which is not very explanatory.

mandresm (approved):

denizural:
#bump

pgierz (approved):

PR #718, opened 2022-07-04
fix(jsbach.yaml): correct naming convention for harvest file

pgierz:
Corrects the land-use harvest file name in work folder. Checked for ECHAM 6.3.05p2 in AWI-ESM and ECHAM used in ReCOM.

`esm-tests` will need an update of "the truth" 😉

### DIFF
--- a/configs/components/jsbach/jsbach.yaml
+++ b/configs/components/jsbach/jsbach.yaml
@@ -289,7 +289,8 @@ forcing_in_work:
         "a_nd_file": "a_nd_file.nc"
         "lightning": "lightning.nc"
         "popdens": "population_density.nc"
-        "LU" : "landuse_harvest.@YEAR@.nc"
+        # Naming based upon: https://gitlab.awi.de/paleodyn/Models/echam6/-/blob/6.3.05p2-awiesm-2.1/src/src_jsbach3/mo_cbal_landcover_change.f90#L3050 
+        "LU" : "landuseHarvest.@YEAR@.nc"
         "Ndepo": "Ndepo.@YEAR@.nc"
         "LU_trans": "landuseTransitions.@YEAR@.nc"
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.11.1
+current_version = 6.11.2
 commit = True
 tag = True
 
--- a/setup.py
+++ b/setup.py
@@ -92,6 +92,6 @@ setup(
     test_suite="tests",
     tests_require=test_requirements,
     url="https://github.com/esm-tools/esm_tools",
-    version="6.11.1",
+    version="6.11.2",
     zip_safe=False,
 )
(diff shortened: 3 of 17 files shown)


configs/components/jsbach/jsbach.yaml
```diff
@@ -289,7 +289,8 @@ forcing_in_work:
         "a_nd_file": "a_nd_file.nc"
         "lightning": "lightning.nc"
         "popdens": "population_density.nc"
-        "LU" : "landuse_harvest.@YEAR@.nc"
+        # Naming based upon: https://gitlab.awi.de/paleodyn/Models/echam6/-/blob/6.3.05p2-awiesm-2.1/src/src_jsbach3/mo_cbal_landcover_change.f90#L3050 
+        "LU" : "landuseHarvest.@YEAR@.nc"
```
pgierz, line 293: To be cleaned, maybe: we have `LU_trans` below, maybe we should change here to `LU_harvest` here. In principle, I would also be happy with changing some of the other left-hand-side names to be more human understandable, e.g. `population_density` instead of `popdens`, and `a_nd_file`...whatever that is? It will help people also understand what JSBACH is doing.

mandresm:
Does this PR needs to come with a warning to the users? (as we talk about it in the AWICM meeting). So if your version is higher than the current one then it displays something like:
```
From version <version> ESM-Tools uses by default the correct LU files...  (a bit more explanation here). If you want to continue an old simulation and still run with the "wrong" files for consistency pleas change <this variable>
```

Something on those lines, @chrisdane maybe you can help us with a better text for the message as you are now an expert on the topic :)

pgierz:
👍🏻 +1 for @chrisdane writing a detailed warning, as the now un-willing expert on JSBACH :-)

chrisdane:
Would be happy to provide meaningful warning messages. 

But I feel two things are mixed here: this PR fixes a bug (and Paul suggests to give better variable names, I agree on that). Miguel asks for a warning message if different forcing files are used across experiments. 

These are two different things for me. Miguel, the community did not decided yet which landuseTransition file should be the default (discussion #695). So I dont see a point in providing a warning message for this specific PR.

pgierz:
We can add more human readable names in this PR.

@chrisdane, you can make a new PR with the warning text you have in mind, and then we figure out where to put it once there is a consensus. I am afraid that will need to be a meeting though, since we might need to poke people a bit to make up their minds...

@mandresm can help you with the syntax if you get stuck, but there is already some documentation for it here: https://esm-tools.readthedocs.io/en/latest/yaml.html#error-handling-and-warning-syntax

mandresm:
> Would be happy to provide meaningful warning messages.
> 
> But I feel two things are mixed here: this PR fixes a bug (and Paul suggests to give better variable names, I agree on that). Miguel asks for a warning message if different forcing files are used across experiments.
> 
> These are two different things for me. Miguel, the community did not decided yet which landuseTransition file should be the default (discussion #695). So I dont see a point in providing a warning message for this specific PR.

Yeap, I was confused, sorry.

denizural:
I forgot about this discussion and made the same PR: https://github.com/esm-tools/esm_tools/pull/720
We can take this and merge is so that the latter can be dismissed.

denizural (approved):
👍

pgierz:
#bump

PR #720, opened 2022-07-05
fixes landuse_harvest to landuseHarvest

denizural:
How the hell awicm-1 run without that?


mandresm:
I might be crazy, but isn't this exactly the same as #718?

denizural:
Yes, it is. I just cherry picked from my commit on ssp branch. Because this was the reason for the awicm-recom failure. The Paul told me to make a pr. Anyways, we can ignore this PR if the previous one already solves it.

pgierz:
Closed, see #718 instead

PR #722, opened 2022-07-05
fix: change the -U flag to use questionary

pgierz:
This PR changes the `-U` flag to also use questionary. I do not know why it is defined in two places...

### DIFF
--- a/src/esm_runscripts/compute.py
+++ b/src/esm_runscripts/compute.py
@@ -5,6 +5,7 @@ import copy
 import pathlib
 
 import f90nml
+import questionary
 import six
 import yaml
 from colorama import Fore, Back, Style, init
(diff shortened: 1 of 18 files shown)


denizural:
👍

denizural (approved):

pgierz:
#bump

src/esm_runscripts/compute.py
```diff
@@ -462,28 +463,16 @@ def update_runscript(fromdir, scriptsdir, tfile, gconfig, file_type):
...
-                        esm_parser.user_error(
-                            f"Original {file_type} different from target",
-                            differences
-                            + "\n"
-                            + "You can choose to -U flag in the esm_runscripts call "
-                            + "to update the runscript without asking (WARNING: This "
-                            + f"will overwrite your {file_type} in the experiment folder!)\n\n",
-                        )
-                    else:
-                        print(f"'{update_choice}' is not a valid answer.")
```
mandresm, line 486: My brain might be burnt by now... but what happens if the user selects `n`?

pgierz: it's already merged....but if you push `n`, you just get asked the question again until you push C-c. That might be better to just exit instead. I'll fix that in a bit if we want that behavior

mandresm: I think it is better to let the user decide what to do, if you keep asking the same question again and again after they set `n` many users will eventually go for `y`.

mandresm: by the way, using `questionary` we can get rid of the while loop as it automatically detects wrong answers and asks again right?

mandresm: I mean substituting the following
https://github.com/esm-tools/esm_tools/blob/ae1ce2be3a622bf3bfd3b4f67beac3d48f485d7e/src/esm_runscripts/prepexp.py#L366-L376
by something like this:
```
                update_choice = questionary.confirm(
                    f"Do you want that {scriptsdir}/{tfile} is "
                    + "updated with the above changes?"
                ).ask()
                if update_choice:
                    oldscript = fromdir + "/" + tfile
                    print(oldscript)
                    shutil.copy2(oldscript, scriptsdir)
                    print(f"{scriptsdir + '/' + tfile} updated!")
                elif not update_choice:
                    print("Submission stopped")
                    sys.exit(0)
```
Note that we are getting rid of the `while` loop, and at the same time it will accept `y` (or `return`) and `n`, but no other input. I also like that `n` stops the submission.

PR #715, opened 2022-07-01
Feat/awiesm 2.1 levante

khimstedt:
The changes are mainly related to the tracer loop parallelization of FESOM REcoM for awiesm 2.1 on Levante. Some of the changes may also be of more general relevance: E.g. workaround for hetjob blocks generation in runscript header, workaround for hetjob blocks generation for `srun` command, setting of OpenMP environment variables for appropriate pinning of threads, setting `--hint=nomultithread` option for `srun` command.

PR #723, opened 2022-07-06
Fix/levante by kai

mandresm:
Fixes for Levante made by @khimstedt (cherry-picked here), taken from #715 for earlier merging and to minimize the PR content.

@khimstedt wrote on #715:
> Some of the changes may also be of more general relevance: E.g. workaround for hetjob blocks generation in runscript header, workaround for hetjob blocks generation for `srun` command, setting of OpenMP environment variables for appropriate pinning of threads, setting `--hint=nomultithread` option for `srun` command.


src/esm_runscripts/batch_system.py
```diff
@@ -829,6 +846,12 @@ def calc_launcher_flags(config, model, cluster):
             nproc = config[model]["nproca"] * config[model]["nprocb"]
             cpus_per_proc = 1
             omp_num_threads = 1
+        else:
+
+# kh 22.06.22 defensive (user_error/user_note could also be added here)
+            nproc = 1
+            cpus_per_proc = 1
+#           omp_num_threads = 1
```
mandresm:
```suggestion
            nproc = 0
            cpus_per_proc = 0
#           omp_num_threads = 0
```
By making all these 0 if no `command`, the number of `nodes` calculated right below will be 0. This means that for `hdmodel` the number of nodes to be requested is 0 and therefore we won't need the `hdmodel` exception in one of the other changes below in the PR.

src/esm_runscripts/slurm.py
```diff
@@ -200,7 +200,17 @@ def het_par_headers(config, cluster, headers):
...
+# kh 24.06.22 observed behavior was:
+#model:  echam, nodes:  4
+#model:  fesom, nodes:  32
+#model:  oasis3mct, nodes:  None
+#model:  recom, nodes:  None
+#model:  jsbach, nodes:  None
+#model:  hdmodel, nodes:  1
+
+# kh 24.06.22 workaround: filter hdmodel
+                if nodes and (model != "hdmodel"):
```
mandresm:
```suggestion
                if nodes:
```
This should work if the previous change is taken

configs/machines/levante.yaml
```diff
@@ -275,7 +291,12 @@ mpifc: "$MPIFC"
 mpicc: "$MPICC"
 cxx: "$CXX"
 
-launcher_flags: "-l --distribution=plane=128 --cpu-bind=${cpu_bind}"
+# kh 23.06.22 do not use --distribution=plane=128 for hybrid (MPI + OpenMP) jobs
+# launcher_flags: "-l --distribution=plane=128 --cpu-bind=${cpu_bind}"
+
+# kh 23.06.22 should be set generally in accordance with 
+# the tips from DKRZ-TechTalk "From Mistral to Levante" June 22, 2022
+launcher_flags: "-l ${mt_launcher_flag}" # --distribution=block:cyclic"
```
mandresm: @khimstedt , I moved the `--hint=nomultithread` flag to be added here, so that it only affects `Levante` and not all the other systems we are currently using. Before, you added it to the `launcher_flags_per_component` in the `slurm.yaml` which means that for an heterogeneous parallel `srun` call the flag will be called for each component:
```
srun ... --hint=nomultithreading <component1> :
... --hint=nomultithreading <component2>
```

After my changes --hint=nomultithreading is used only once. Would that be okay for heterogeneous parallelization runs with hetjobs?

JanStreffing (approved):

JanStreffing:
I found no problems with these change. Also no significant impact on my AWI-CM3 simulations on levante. Probably needs to be checked in fesom-recom of course.

configs/machines/levante.yaml
khimstedt: On Levante
```
time srun -l \
--nodes=4 --ntasks=512 --ntasks-per-node=128 --cpus-per-task=1 --export=ALL,OMP_NUM_THREADS=1 ./echam6 \
: --hint=nomultithread --nodes=8 --ntasks=512 --ntasks-per-node=64 --cpus-per-task=2 --export=ALL,OMP_NUM_THREADS=2 ./fesom  2>&1
```
would work whereas

```
time srun -l \
--hint=nomultithread --nodes=4 --ntasks=512 --ntasks-per-node=128 --cpus-per-task=1 --export=ALL,OMP_NUM_THREADS=1 ./echam6 \
: --nodes=8 --ntasks=512 --ntasks-per-node=64 --cpus-per-task=2 --export=ALL,OMP_NUM_THREADS=2 ./fesom  2>&1
```
would not work.

Nevertheless I would prefer to specify on Levante for the sake of clarity `--hint=nomultihread` (even if the name of this option is a bit misleading - it is essentially setting "hyper-threading=off") for each `srun` component:
```
time srun -l \
--hint=nomultithread --nodes=4 --ntasks=512 --ntasks-per-node=128 --cpus-per-task=1 --export=ALL,OMP_NUM_THREADS=1 ./echam6 \
: --hint=nomultithread --nodes=8 --ntasks=512 --ntasks-per-node=64 --cpus-per-task=2 --export=ALL,OMP_NUM_THREADS=2 ./fesom  2>&1
```
It would also be defensive unless hyper-threading is explicitly desired. A positive side effect is that then also for ECHAM in this example exactly the cores 1-128 of a node are used, which simplifies plausibility checks with `htop`: Are all 128 phys. cores used and is hyper-threading not active – i.e. cores 129-256 of the `htop` columns 3 and 4 should be idle.

mandresm: Hi Kai, I'll work on a solution so that each srun component can get the flag without affecting the other HPCs srun commands.

mandresm:
Hi @khimstedt, I have added the `--hint=nomultithread` flag back into each srun component, but this time is activated exclusively in levante (through the `mt_launcher_flag` which default is empty for the other computers), and it is included in heterogeneous parallelization with `hetjobs`, in the `taskset` trick and in non heterogeneous parallelization:

**Het par with hetjobs**
```
time srun -l \
--hint=nomultithread --nodes=1 --ntasks=128 --ntasks-per-node=128 --cpus-per-task=1 --export=ALL,OMP_NUM_THREADS=1 ./fesom \
: --hint=nomultithread --nodes=2 --ntasks=128 --ntasks-per-node=64 --cpus-per-task=2 --export=ALL,OMP_NUM_THREADS=2 ./oifs -v ecmwf -e awi3 \
: --hint=nomultithread --nodes=1 --ntasks=1 --ntasks-per-node=1 --cpus-per-task=128 --export=ALL,OMP_NUM_THREADS=128 ./rnfma  2>&1 &
```

**Het par with taskset**
```
 time srun -l --hint=nomultithread --multi-prog hostfile_srun 2>&1 &
```

**non-heterogeneous parallelization**
```
time srun -l --kill-on-bad-exit=1 --cpu_bind=cores --hint=nomultithread --multi-prog hostfile_srun 2>&1 &
```

mandresm:
#bump

PR #725, opened 2022-07-07
trying to fix the inclusion of the oasis_rmp_rst_to_input.py into the…

JanStreffing:
… recipe of awicm3

This works again now. ty!


mandresm:
That's great news! Thanks for testing it!

#bump

mandresm (approved):

PR #729, opened 2022-07-10
modifications to the update runscript interface

mandresm:
Changes to #722
- getting rid of an unnecessary while loop
- if you answer that you don't want to update the runscript submission is halted

### DIFF
--- a/src/esm_runscripts/compute.py
+++ b/src/esm_runscripts/compute.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import subprocess
+import sys
 import copy
 import pathlib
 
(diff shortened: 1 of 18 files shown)


src/esm_runscripts/compute.py
```diff
@@ -457,22 +458,23 @@ def update_runscript(fromdir, scriptsdir, tfile, gconfig, file_type):
...
-                        print(f"{scriptsdir + '/' + tfile} updated!")
+                update_choice = questionary.confirm(
+                    f"Do you want that {scriptsdir}/{tfile} is "
+                    + "updated with the above changes?"
+                ).ask()
+                if update_choice:
+                    oldscript = fromdir + "/" + tfile
+                    print(oldscript)
+                    shutil.copy2(oldscript, scriptsdir)
+                    print(f"{scriptsdir + '/' + tfile} updated!")
```
denizural, line 474: Why not simply an `else`?

src/esm_runscripts/prepexp.py
```diff
@@ -358,22 +359,23 @@ def update_runscript(fromdir, scriptsdir, tfile, gconfig, file_type):
...
+                update_choice = questionary.confirm(
+                    f"Do you want that {scriptsdir}/{tfile} is "
+                    + "updated with the above changes?"
+                ).ask()
+                if update_choice:
+                    oldscript = fromdir + "/" + tfile
+                    print(oldscript)
+                    shutil.copy2(oldscript, scriptsdir)
+                    print(f"{scriptsdir + '/' + tfile} updated!")
+                elif not update_choice:
```
denizural: Same here.  Or are we expecting something more to ask?

denizural (changes requested):

src/esm_runscripts/compute.py
pgierz: It's "Kleinkram", but @denizural is right: you could just use an `else` here. `questionary.confirm` will always give you back a boolean, so it'll always be `if True, else`.

src/esm_runscripts/prepexp.py
```diff
@@ -358,22 +359,23 @@ def update_runscript(fromdir, scriptsdir, tfile, gconfig, file_type):
...
+                    print("Submission stopped")
+                    sys.exit(0)
```
pgierz: I would exit with 1 here instead of 0. exit code 1 is the standard "something went wrong" message, and I would say that this counts at "something went wrong"

pgierz (changes requested):
Everything looks good, I would just exit with 1 instead of 0.

src/esm_runscripts/prepexp.py
denizural: Do you really think that is should be 1. Isn't that the user's choice?

pgierz: that turns more and more into a philosophical question, one in which I need more coffee to address. In the end, I would say `esm_runscripts` should return `0` if the run ends up on the batch system, and `1` (or something more specific) if that is not the case

mandresm:
All your suggestions were added, can someone give me an approval?

denizural (approved):
👏 👏👏

pgierz (approved):

pgierz:
#bump

Issue #734, opened 2022-07-12
git feature #690 broken


JanStreffing:
It seems to me that the git version feature #690 is not quite working.
After merging release into my current test branch in oder to get the latest levante flags, I recompiled awicm3-v3.1 on levante and tried to run a basic runscript (`/runscripts/awicm3/v3.1/awicm3-v3.1-levante-TCO95L91-CORE2_initial.yaml`)

I get:

```
Traceback (most recent call last):
  File "/home/a/a270092/.local/bin/esm_runscripts", line 33, in <module>
    sys.exit(load_entry_point('esm-tools', 'console_scripts', 'esm_runscripts')())
  File "/home/a/a270092/esm_tools/src/esm_runscripts/cli.py", line 278, in main
    Setup()
  File "/home/a/a270092/esm_tools/src/esm_runscripts/sim_objects.py", line 69, in __call__
    self.prepcompute()
  File "/home/a/a270092/esm_tools/src/esm_runscripts/sim_objects.py", line 161, in prepcompute
    self.config = prepcompute.run_job(self.config)
  File "/home/a/a270092/esm_tools/src/esm_runscripts/prepcompute.py", line 36, in run_job
    config = evaluate(config, "prepcompute", "prepcompute_recipe")
  File "/home/a/a270092/esm_tools/src/esm_runscripts/helpers.py", line 61, in evaluate
    config = esm_plugin_manager.work_through_recipe(
  File "/home/a/a270092/esm_tools/src/esm_plugin_manager/esm_plugin_manager.py", line 141, in work_through_recipe
    config = getattr(submodule, workitem)(config)
  File "/home/a/a270092/esm_tools/src/esm_runscripts/prepare.py", line 702, in add_vcs_info
    vcs_versions[model] = helpers.get_all_git_info(model_dir)
  File "/home/a/a270092/esm_tools/src/esm_runscripts/helpers.py", line 422, in get_all_git_info
    "branch_name": get_git_branch(path),
  File "/home/a/a270092/esm_tools/src/esm_runscripts/helpers.py", line 375, in get_git_branch
    return repo.head.reference.name
  File "/home/a/a270092/.local/lib/python3.9/site-packages/git/refs/symbolic.py", line 294, in _get_reference
    raise TypeError("%s is a detached symbolic reference as it points to %r" % (self, sha))
TypeError: HEAD is a detached symbolic reference as it points to 'e48c9f09ff85aaf7ccb4e20d84e06887eb99d824'
```

[...]

PR #735, opened 2022-07-12
fix/provisional solution to 734

mandresm:
reverts the version bug in #690 (reported in #734) by removing the steps affected in the recipe (provisional solution)


mandresm:
Files are different from the "truth" because the truth has not been updated for a while, but at least all the checks run.

mandresm:
#bump

mandresm:
#bump the previous did not work because a bumped branch was not merged and tags where conflicting

PR #738, opened 2022-07-13
fix(doctest): allow doctest to pass

pgierz:
...and removes the broken database functionality of `esm_archiving`, which no one was using anyway.


pgierz:
All failing tests are now only from `esm-tests`, and only as "Run Files Differ"

mandresm (approved):

mandresm:
#bump

PR #745, opened 2022-07-22
fixes the venv for the monorepo

mandresm:
On the past we had to install the multiple packages of ESM-Tools in the virtual environment. Now we only need the `esm_tools` package because that contain all the rest. However, when running a simulation with `venv` it was still installing all the independent deprecated packages on top of the `esm_tools` one and of course that was leading to conflicts on the `imports`. This PR solves the problem.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.13.0
+current_version = 6.13.1
 commit = True
 tag = True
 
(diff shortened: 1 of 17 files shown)


pgierz (approved):
Sorry I forgot about that. Too often we have duplicated info

mandresm:
I thought you were on holidays, I won't tag you in anything else today, I'll wait until Monday :p
#bump

PR #748, opened 2022-07-26
fix/awiesm_general_scenario

mandresm:
Before, the AWIESM's `general.scenario` was never transmitted to ECHAM. One can see by just running a runscript without any scenario variable, and then looking at the `finished_config.yaml`: the `general.scenario` there is `PALEO`, the default for AWIESM, however, in ECHAM the scenario is `PI-CTRL`, the default for ECHAM. 

`general.scenario` did really nothing on the past, since it was not fed into any other variable, but `echam.scenario` does a bunch. That's what I decided to change the default of `general.scenario` for AWIESM to `PI-CTRL`, so when people update ESM-Tools the still get the same behavior (`echam.scenario: PI-CTRL`)

### DIFF
--- a/configs/setups/awiesm/awiesm.yaml
+++ b/configs/setups/awiesm/awiesm.yaml
@@ -85,7 +85,7 @@ general:
         with_icebergs: false
         with_bootstrap: false
 
-        scenario: "PALEO"
+        scenario: "PI-CTRL"
         resolution: ${echam.resolution}_${fesom.resolution}
         postprocessing: false
         post_time: "00:05:00"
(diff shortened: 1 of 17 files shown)


mandresm:
Not all the test passed because... well, the ones that use PALEO scenarios now think they are PI-CTRL. I'll fix that in the runscripts of the tests tomorrow.

pgierz:
The only thing that is important to check here is the dynamic vegetation settings. AWI-ESM should be default of dynamic vegetation turned on.

pgierz (approved):
JSBACH seems to have that in it's own section, so ignore my previous comment.

pgierz:
#bump

PR #749, opened 2022-07-26
echam can now do daily restarts

mandresm:
Fixes the bug for daily restarts in ECHAM.

In echam.yaml , namelist_changes there was this:
```
                         dt_stop:
                                 - ${pseudo_end_date!syear}
                                 - ${pseudo_end_date!smonth}
                                 - ${pseudo_start_date!sday}
```
where `pseudo_end_date`  and `pseudo_start_date` are the dates of the end and start of this run, respectively. The bug was on the extraction of the day for `pseudo_start_date` instead of `pseudo_end_date`. This worked for monthly and yearly restarts as they all start on day 1 of the month, but it doesn’t work for daily restarts of fractions of months because it would be providing the incorrect day for stopping.

### DIFF
--- a/configs/components/echam/echam.yaml
+++ b/configs/components/echam/echam.yaml
@@ -730,7 +730,7 @@ namelist_changes:
                         dt_stop:
                                 - ${pseudo_end_date!syear}
                                 - ${pseudo_end_date!smonth}
-                                - ${pseudo_start_date!sday}
+                                - ${pseudo_end_date!sday}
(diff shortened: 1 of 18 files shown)


seb-wahl (approved):
Wow that's a mean one. good work @mandresm

mandresm:
Thanks!

This passes all the tests, which means monthly and yearly restarts have no problem with this fix :)

#bump

PR #756, opened 2022-08-01
feat/pool_utils

pgierz:
This PR creates a small utility function for checking dead links in an HPC pool
which may have been `rsync`-ed across from another HPC system, and provides
ways of fixing these links. Usage examples can be seen in the docstrings for `cli.py`


pgierz:
Not all functionality is there yet, but I would include extra bonus features over time. For right now, the things in this PR should be stable and ready to be put into the main branch.

pgierz:
As @seb-wahl is in charge of pools for GEOMAR, I would also be happy to get his feedback here. Importantly: how to handle links with different prefixes: the easiest solution is to use relative links, but maybe there is something more clever

setup.py
```diff
@@ -13,6 +14,7 @@
 
 requirements = [
     "Click>=7.0",
+    "click-loguru==1.3.6",
```
seb-wahl, line 17: Do you really want to specify an exact version. Are you sure this will still work in a year from now. I see the risk that we can't resolve this hardcoded dependency anymore at some point in the future.

seb-wahl (commented):
As I only use relative links in our pool, and control everything via git lfs I'll probably never need this script. But I do see the need with your large mesh files which may have absolute links. Instead of replacing with a new absolute link, you could try to replace the broken link with a relative link instead. This would allow to move the pool to a different location on the same machine without any issues.

setup.py
pgierz: Unless PyPI goes offline, this should always be available. I would even prefer exact versions to non-specific ones: this way I know exactly what I get and (while unlikely) it might be that getting whatever is "newest" has slightly different behaviors. Finding that out after the fact sounds painful. For a small dependency like this it isn't a very big problem, but imagine for instance if the YAML loading library suddenly behaves differently.

seb-wahl: Ok, thanks for clarifying. I had problems in another project (old python 2.x project) where hardcoded versions started to make problems when python 2.7 started to die...

seb-wahl (approved):

pgierz:
#bump

mandresm (approved):

denizural:
Merge is probably the other way around. The feat must be merged on to release.

denizural (changes requested):
Merge is in the wrong direction.
Bump must be reverted.

pgierz:
I'll just delete the tag. I also want to see about the relative link Sebastian suggested, so maybe this is still a draft.

PR #765, opened 2022-08-06
Update awicm3 master

JanStreffing:
This branch enables AWI-ESM developers to use the awicm3-master version as a starting point for model developments. It also removes the deprecated awicm3-frontiers, which was without XIOS. Finally it contains a setups that runs a few days for TCO639-SO3 AWI-CM3 with v3.1 on levante.

- awicm3-master is using an updated oifs master branch which includes all changes made for v3.1. 
- awicm3-master is using the latest fesom2 refactoring branch which includes the changes made for v3.1, as well as fixes to the water mass conservation.


JanStreffing:
This is a lot of files, but almost all runscripts. The only changes in files that others use is the fesom2 switching from master to refactoring. This should be fine, as only awicm3 uses the fesom-2.0 yaml. 

I wont be around to provide feedback for a week or so. Would be nice to have this merged unless I made a clear mistake. I think some tests might need an update. Sorry for that.

mandresm:
I don't have time right now to have a look at this, but I'll do it through out the week. Just a question, isn't Frontiers what Helge's group is using for their experiments?

JanStreffing:
@mathanase Are you using the awicm3-frontiers version currently?

mathanase:
@JanStreffing Currently I am working with v3.1 on Levante (so the version with XIOS), and I should still have v3.0 on Juwels (without XIOS). I don't think we used awicm3-frontiers in a while.

mandresm:
Okay, great, I'll be reviewing this by the end of the day so that it can be merged tomorrow.

configs/components/oifs/oifs.env.yaml
```diff
@@ -326,6 +326,7 @@ compiletime_environment_changes:
         # compilers and compile switches
         OIFS_FC: '$FC'
         OIFS_FFLAGS: '"-r8 -fp-model precise -align array32byte -O3 -qopenmp -g -traceback -convert big_endian -march=core-avx2 -mtune=core-avx2"' # -qoverride-limits -fast-transcendentals -m64  -fma -pc64"'
+        #OIFS_FFLAGS: '"-r8 -fp-model precise -align array32byte -O0 -check all,noarg_temp_created,bounds,uninit -ftrapuv -qopenmp -g -traceback -convert big_endian -march=core-avx2 -mtune=core-avx2"'
```
mandresm, line 329: Is this added for a reason or can be just removed?

configs/setups/awicm3/awicm3.yaml
```diff
@@ -299,15 +317,52 @@ fesom:
         choose_general.version:
                 "v3.0":
                         namelist_dir: "${esm_namelist_dir}/fesom2/2.0/awicm3/v3.0/"
+                        add_namelist_changes:
+                                namelist.oce:
+                                        oce_tra:
+                                                surf_relax_s: "${surf_relax_s}"
```
mandresm, line 323:
```suggestion
```
I think this is not needed as it is part of the general `namelist_changes` and it seems you wanted for all the 4 different versions.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -299,15 +317,52 @@ fesom:
...
                 "v3.1":
                         namelist_dir: "${esm_namelist_dir}/fesom2/2.0/awicm3/v3.1/"
+                        add_namelist_changes:
+                                namelist.oce:
+                                        oce_tra:
+                                                surf_relax_s: "${surf_relax_s}"
```
mandresm, line 329:
```suggestion
```
Same as before.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -299,15 +317,52 @@ fesom:
...
+                        add_namelist_changes:
+                                namelist.config:
+                                        ale_def:
+                                                which_ALE: 'zstar'
+                                namelist.oce:
+                                        oce_dyn:
+                                                SPP: true
+                                namelist.tra:
+                                        tracer_phys:
+                                                surf_relax_s: "${surf_relax_s}"
```
mandresm, line 353:
```suggestion
```
Same as the previous one

configs/setups/awicm3/awicm3.yaml
```diff
@@ -299,15 +317,52 @@ fesom:
...
+                                        ice_therm:
+                                                albi: 0.6
+                                                albim: 0.43
                 "frontiers-xios":
                         model_tag: "frontiers"
                         namelist_dir: "${esm_namelist_dir}/fesom2/2.0/awicm3/frontiers/"
+                        add_namelist_changes:
+                                namelist.oce:
+                                        oce_tra:
+                                                surf_relax_s: "${surf_relax_s}"
```
mandresm, line 365:
```suggestion
```
Same as the previous one

configs/setups/awicm3/awicm3.yaml
```diff
@@ -299,15 +317,52 @@ fesom:
...
-                        namelist_dir: "${esm_namelist_dir}/fesom2/2.0/awicm3/frontiers/"
+                        add_namelists:
+                                - namelist.tra
+                                - namelist.cvmix
+                        add_config_sources:
+                                tra:     "${namelist_dir}/namelist.tra"
+                                dyn:     "${namelist_dir}/namelist.dyn"
+                        add_config_in_work:
+                                tra:     "namelist.tra"
+                                dyn:     "namelist.dyn"
```
mandresm, line 340: I would recommend to add these outside of the choose block, in fact the best would be to add them to the `fesom-2.1.yaml` if we would be using that file at all, but given that that's not the case, I say we can keep them at least outside of the choose. `sources` and `in_work` are not used unless `files` points at their tags.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -323,11 +378,11 @@ fesom:
...
 
-        add_config_files:
-                cvmix: cvmix
+        #add_config_files:
+                #cvmix: cvmix
 
-        add_config_sources:
-                cvmix: "${namelist_dir}/namelist.cvmix"
+        #add_config_sources:
+                #cvmix: "${namelist_dir}/namelist.cvmix"
```
mandresm, line 385: Doesn't other versions of AWICM3 need the `cvmix` namelist?

mandresm (changes requested):

configs/components/oifs/oifs.env.yaml
JanStreffing: These are debug flags that i used. I think its good to have them as a comment.

configs/setups/awicm3/awicm3.yaml
JanStreffing: The problem is that this option used to be in namelist.oce, but is now in master in namelist.tra. Thus a general option is no longer possible. We have to state per version where we want that change made.

configs/setups/awicm3/awicm3.yaml
JanStreffing: wont that send warnings about missing namelist templates to users of v3.0 and v3.1?

configs/setups/awicm3/awicm3.yaml
JanStreffing: not by default

configs/components/oifs/oifs.env.yaml
pgierz: Agreed. While it does "pollute" the file and one could go back into the git history to recover it; that often is harder for people than just swapping out a comment.

configs/setups/awicm3/awicm3.yaml
pgierz: That's an annoying design decision, but so be it.

configs/setups/awicm3/awicm3.yaml
pgierz: It would, you could consider adding them to the "allowed to be missing" list. 

Have a look here for the code: https://github.com/esm-tools/esm_tools/blob/f94c5dfc7b70c884e39526927ab6d33eaf95a408/src/esm_runscripts/filelists.py#L1006

And here for example tests: https://github.com/esm-tools/esm_tools/blob/f94c5dfc7b70c884e39526927ab6d33eaf95a408/tests/test_esm_runscripts/test_filelists.py#L6

configs/setups/awicm3/awicm3.yaml
pgierz: Wouldn't it be cleaner to add this on the `fesom.yaml` instead?

pgierz:
Hey @JanStreffing, I had two comments you might still consider and resolved the other conversations. They are only "soft" suggestions, so if you think it's not needed I have no problem with that. 

I'd still like to use this PR to also update Miguel's "Truth" (TM) for the automatic tests, but I am not 100% sure how to do that. A checklist-style guide of how one could do that certainly belongs in the handbook at some point.

If it is urgent, we can skip the truth update and just put this in as is, so long as someone remembers to update the truth at some point down the road.

configs/setups/awicm3/awicm3.yaml
JanStreffing: I'm keeping my modifications to the awicm3.yaml mostly. Since I'm basing off of fesom2.0.yaml there is no cvmix in there. I think it works ok like this.

JanStreffing:
> Hey @JanStreffing, I had two comments you might still consider and resolved the other conversations. They are only "soft" suggestions, so if you think it's not needed I have no problem with that.
> 
> I'd still like to use this PR to also update Miguel's "Truth" (TM) for the automatic tests, but I am not 100% sure how to do that. A checklist-style guide of how one could do that certainly belongs in the handbook at some point.
> 
> If it is urgent, we can skip the truth update and just put this in as is, so long as someone remembers to update the truth at some point down the road.

I think it would be good to have this merged. This allows colleages like Fernanda a clean start for model development. Otherwise we only have tags for AWI-CM3.

I've done tests on juwels and levante with v3.1 v3.0 and master on this branch. They look ok to me.

pgierz:
👍 ok

pgierz (approved):

pgierz:
#bump minor

JanStreffing:
ty for taking over on this one Paul!

PR #766, opened 2022-08-08
Sprint/filedicts/base class

pgierz:
Example for the base class which we shall expand upon.

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -2,10 +2,8 @@
 The file-dictionary implementation
 """
 
-from collections import UserDict
 
-
-class SimulationFile(UserDict):
+class SimulationFile(dict):
     """
     Desribes a file used within a ESM Simulation.
 
@@ -32,16 +30,16 @@ class SimulationFile(UserDict):
         >>> sim_file.cp_to_exp_tree()
     """
 
-    # Do NOT implment an __init__ unless you really think you need to. UserDict
+    # Do NOT implment an __init__ unless you really think you need to. dict
     # takes care of this in a way is that is considerably smarter than any of us are.
     #
     # Please delete my annoying comments before merging into actual release ;-)
 
-    def cp_to_exp_tree(self) -> None:
+    def cp(self) -> None:
         pass
 
-    def ln_to_exp_tree(self) -> None:
+    def ln(self) -> None:
         pass
 
-    def mv_to_exp_tree(self) -> None:
+    def mv(self) -> None:
         pass
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -1,0 +1,47 @@
+"""
+The file-dictionary implementation
+"""
+
+from collections import UserDict
+
+
+class SimulationFile(UserDict):
+    """
+    Desribes a file used within a ESM Simulation.
+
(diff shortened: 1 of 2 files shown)


src/esm_runscripts/filedicts.py
```diff
@@ -0,0 +1,47 @@
+"""
+The file-dictionary implementation
+"""
+
+from collections import UserDict
+
+
+class SimulationFile(UserDict):
```
mandresm: Do we really need to use `UserDict` here? I subclassed `dict` for now quite some classes in ESM-Tools (`prev_run`, `SmartSink`, ...) and I never really missed any of the extra functions UserDict provides. I've also been reading that it was mostly valuable in Python versions <2.2, but that now a days, subclassing the `dict` is practically the same. Am I missing something fundamental here?

src/esm_runscripts/filedicts.py
```diff
@@ -0,0 +1,47 @@
...
+    # Please delete my annoying comments before merging into actual release ;-)
+
+    def cp_to_exp_tree(self) -> None:
+        pass
+
+    def ln_to_exp_tree(self) -> None:
+        pass
+
+    def mv_to_exp_tree(self) -> None:
+        pass
```
```suggestion
    def cp(self, source, target) -> None:
        pass

    def ln(self, source, target) -> None:
        pass

    def mv(self, source, target) -> None:
        pass
```

mandresm: we need simple `cp`/`mv`/'ln' methods, because we might copy in multiple directions right? Where the direction will be given from the `source` and `target` right?

mandresm (changes requested):

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -0,0 +1,24 @@
+import yaml
+
+from esm_runscripts.filedicts import SimulationFile
+
+
+def test_filedicts_basics(fs):
```
denizural, line 43: We can move the test cases as fixtures. I can do it later on.

src/esm_runscripts/filedicts.py
pgierz: OK, instead from `dict`. Here is a good article: https://realpython.com/inherit-python-dict

src/esm_runscripts/filedicts.py
pgierz: I put this in without `source` and `target` for right now; some things might be extracted from the object's properties. For right now, everything is anyway just a "stub" of a method, actual implementation starts tomorrow.

src/esm_runscripts/filedicts.py
mandresm: thanks, reading through it (I did not find this article during my google search :S )

tests/test_esm_runscripts/test_filedicts.py
pgierz: Not sure what you mean by this...?

denizural: I mean the test cases and their expected results can be parameterized by a decorator.

pgierz: OK, that would be a separate PR though. For right now, I would keep the tests in a way that they can be written down as simply as possible, just make sure `test_blahblahblah` is in the name so `pytest` recognizes it.

mandresm (approved):

PR #767, opened 2022-08-08
copy function

mandresm:
`cp` method to be developed in this branch.

**TODO**

- [x] capable of copying files
- [x] capable of copying folders
- [x] basic unit test for copied file
- [x] basic unit test for copied folder


mandresm:
Very simple unit test added. I want to add at least one unit test more that checks for errors reported from the function (non-existing source file/folder or target folder)

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -59,3 +61,37 @@ def test_filedicts_basics(fs):
...
+                name_in_pool: T63CORE2_jan_surf.nc
+                name_in_work: unit.24
+                path_in_pool: /work/ollie/pool/ECHAM/T63/
+        thisrun_work_dir: /work/ollie/mandresm/awiesm/run_20010101-20010101/work/
+    """
+    config = yaml.safe_load(dummy_config)
+
+    # Set source and targets
+    target_folder = config["echam"]["thisrun_work_dir"]
+    source = Path(
```
pgierz: It might be nicer to ensure certain parts always become Path during the dict init.

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -59,3 +61,37 @@ def test_filedicts_basics(fs):
...
+        target_folder,
+        config["echam"]["files"]["jan_surf"]["name_in_work"],
+    )
+
+    # Create files and folders
+    fs.create_file(source)
+    fs.create_dir(target_folder)
+
+    # Test the method
+    esm_runscripts.filedicts.SimulationFile.cp(source, target)
```
pgierz: Do you want this as a class method, or rather as a method on the object? Right now it's written  down as a class method.

mandresm: shit sorry... I wanted as a class method and I forgot to initialize the object (facepalm). I'll fix that right away.

mandresm: I just fixed it, I think it now makes much more sense. No use of real paths for `source` and `target`, but instead used to specified directions (`source: "pool"`, `target: "work"`) because the actual paths are gonna be computed by the function itself using the inner properties of the simulation file (just an idea, I will bring this up tomorrow to our meeting)

pgierz: One problem I still foresee: we are at some point going to need to give this thing the config so it knows how to look up such information. I am not sure I really like that..😵‍💫

src/esm_runscripts/filedicts.py
```diff
@@ -45,30 +47,93 @@ def __init__(self, full_config, attrs_address):
...
-    # PG: No. We do not want this kind of general function. This is just to
-    # demonstrate how the test would work
-    return config
+
+    def _path_type(self, path):
+        if path.is_file():
+            return "file"
+        elif path.is_dir:
+            return "dir"
+        elif path.is_link:
```
denizural: This should be `islink()`. Ie. `os.path.islink()`

denizural (changes requested):

src/esm_runscripts/filedicts.py
mandresm: fixed

mandresm:
I say this is ready for review. I will create a new ticket for producing doctstrings for the auxiliary methods and preparing unit tests for those, so that this can be merge asap, to don't block tomorrow's work.

pgierz (commented):
Looks good except one thing: do we want the class `SimulationFile` to also describe a folder worth of files? To me that feels like a different thing, but I am not sure.

src/esm_runscripts/filedicts.py
```diff
@@ -45,30 +47,112 @@ def __init__(self, full_config, attrs_address):
...
+
+        # Actual copy
+        if source_path_type == "dir":
+            copy_func = shutil.copytree
+        else:
+            copy_func = shutil.copy2
+        try:
+            copy_func(source_path, target_path)
+        except Exception:
+            user_error("Filedict Error", f"Unable to copy {source_path} to {target_path}")
```
denizural: I would not use `user_error` here. It is a very strong function because it aborts the whole executable. I think it is better to raise an exception and let the driver routine to give the final decision.

mandresm:
> Looks good except one thing: do we want the class `SimulationFile` to also describe a folder worth of files? To me that feels like a different thing, but I am not sure.

I don't understand what you mean by a "folder worth of files"

mandresm:
I understand it now, you mean that a file in our `files` can be a folder.

This comes from OpenIFS needing to copy/linking a lot of input that it's always the same folder, and where OpenIFS expects this folder in the work directory. Sebastian and Jan wanted to do it all at once, so what they ended up doing was to indicate in the files `in_sources: path/o3_data/*` and `in_work: o3_data/*`.

It's the same for the memory-dump-restarts of fesom for fast restarting in high resolution models: they use globing, and I was always under the impression that using globbing for so many files (even if it was just moving or linking) was slower than simply moving/linking the whole folder (because it has to iterate at a higher level than just the shutil.copytree)

denizural (changes requested):

src/esm_runscripts/filedicts.py
mandresm: Agreed

pgierz (approved):

PR #770, opened 2022-08-09
move function

pgierz:
Draft of move functionality.

I am unsure about the `get`, since it might return `None`. What should the default be? The `.` for "here"? Or change from get to direct access and instead throw a key error if you don't have it set?

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -4,6 +4,8 @@ The file-dictionary implementation
 import pathlib
 
 import dpath.util
+from esm_parser import user_error
+from loguru import logger
 
 
 class SimulationFile(dict):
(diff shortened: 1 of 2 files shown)


src/esm_runscripts/filedicts.py
```diff
@@ -63,10 +65,40 @@ def cp(self) -> None:
...
+        # will use it in copy, move, and link:
+
+        # Figure out names in source and target:
+        sname = self.get(f"name_in_{source}")
+        tname = self.get(f"name_in_{target}")
+        # Relative path in source and target
+        srpath = self.get(f"path_in_{source}")
+        trpath = self.get(f"path_in_{target}")
+        # Build target and source paths:
+        spath = self.locations[source].joinpath(srpath, sname)
```
denizural: we can use little bit more verbosity like `source_path`, `target_path` and let vim do the autocomplete with `ctrl-n`

denizural (changes requested):

denizural (approved):

src/esm_runscripts/filedicts.py
```diff
@@ -63,10 +67,68 @@ def cp(self) -> None:
...
         pass
 
-    def mv(self) -> None:
-        pass
-        
-        
+    def mv(self, source: str, target: str) -> None:
+        """
+        Moves (renames) the SimulationFile from it's location in `source` to
+        it's location in `target`.
```
```suggestion
        Moves (renames) the SimulationFile from it's location in ``source`` to
        it's location in ``target``.
```

mandresm (approved):
Great, I'm excited about starting to unify both cp and mv tomorrow :)

denizural:
I would keep the `ValueError` because `user_error` lowers the testability of the function. `user_error` shoud be called from the driver routine instead.

src/esm_runscripts/filedicts.py
```diff
@@ -63,10 +67,68 @@ def cp(self) -> None:
...
+        source : str
+            One of ``"pool"``, ``"work"``, ``"exp_tree"``, "``run_tree``"
+        target : str
+            One of ``"pool"``, ``"work"``, ``"exp_tree"``, "``run_tree``"
+        """
+        if source not in self.locations:
+            user_error(
+                "Filedict Error",
+                "source is incorrectly defined, and needs to be in {self.locations}",
+            )
```
denizural, line 84: I would keep the ValueError because user_error lowers the testability of the function. user_error shoud be called from the driver routine instead. What about the builtin `FileNotFoundError`?

src/esm_runscripts/filedicts.py
```diff
@@ -63,10 +67,68 @@ def cp(self) -> None:
...
+        source_path, target_path = self._determine_names(source, target)
+        # Perform the movement:
+        try:
+            source_path.rename(target_path)
+            logger.success(f"Moved {source_path} --> {target_path}")
+        except Exception:  # Probably better to look for specific "breaking" things here
+            user_error(
```
denizural: Same here. I think `IOError` would be better.

denizural (changes requested):

src/esm_runscripts/filedicts.py
pgierz: I would say `ValueError`, as the value of the variable `source` is incorrect. That is fixed though now.

src/esm_runscripts/filedicts.py
pgierz: OK, it has `IOError` now, but with a customized message.

PR #771, opened 2022-08-09
Sprint/filedicts/resolve file movements

denizural:
New functionality `resolve_file_movements` that may replace `assemble` in the future

### DIFF
--- a/tests/test_esm_runscripts/test_filedicts.py
+++ b/tests/test_esm_runscripts/test_filedicts.py
@@ -12,12 +12,15 @@ Some considerations
   unit test, which you don't really want.  
 """
 import os
-from pathlib import Path
 
 import yaml
 import pytest
 
 import esm_runscripts.filedicts as filedicts
+
+from pathlib import Path
+
+import esm_runscripts.filedicts
 
 
 @pytest.fixture()
--- a/tests/test_esm_runscripts/test_filedicts.py
+++ b/tests/test_esm_runscripts/test_filedicts.py
@@ -12,12 +12,20 @@ Some considerations
   unit test, which you don't really want.  
 """
 import os
-
-import yaml
-
 from pathlib import Path
 
+import yaml
+import pytest
+
 import esm_runscripts.filedicts
+from esm_runscripts.filelists import resolve_file_movements
+
+
+@pytest.fixture()
+def config():
+    """Generates fake config to be used before each test"""
+    fake_config = dict()
+    yield fake_config
 
 
 def test_example(fs):
@@ -41,7 +49,8 @@ def test_example(fs):
     assert os.path.exists(
         "/some/dummy/location/expid/run_18500101-18501231/work/unit.24"
     )
-    
+
+
 def test_filedicts_basics(fs):
     """Tests basic attribute behavior of filedicts"""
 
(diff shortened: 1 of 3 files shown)


tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -12,12 +12,20 @@
...
 
 import esm_runscripts.filedicts
+from esm_runscripts.filedicts import resolve_file_movements
+
+
+@pytest.fixture()
+def config():
+    """Generates fake config to be used before each test"""
+    fake_config = dict()
+    yield fake_config
```
pgierz, line 29: Why does this need to be a yield? (Just out of curiosity)

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -12,12 +12,20 @@
   unit test, which you don't really want.  
 """
 import os
+from pathlib import Path
 
 import yaml
-
-from pathlib import Path
+import pytest
```
pgierz, line 18: The import order might change once in a while. I typically run `isort` from my vim whenever a Python file is saved, but that is a trivial change that might happen now and then.

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -12,12 +12,20 @@
...
 
 import esm_runscripts.filedicts
+from esm_runscripts.filedicts import resolve_file_movements
```
pgierz: Could be removed, you already have the entire module imported above.

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -95,3 +103,11 @@ def test_cp(fs):
     esm_runscripts.filedicts.SimulationFile.cp(source, target)
 
     assert os.path.exists(target)
+
+
+def test_resolve_file_movements(config):
+    # act
+    config = resolve_file_movements(config)
```
pgierz: I would rather call the function via `module.func_name`, as it is done in other tests.

pgierz (changes requested):
Looks good. I would remove the `filelists.py` from this PR to keep it focused on only new functionality.

tests/test_esm_runscripts/test_filedicts.py
mandresm: agreed

mandresm (commented):
Everything looks good to me too after resolving the requests from @pgierz. I would also remove the changes of `filelists.py` from this PR and use them for another one, maybe even for `release`.

tests/test_esm_runscripts/test_filedicts.py
denizural: `yield` is recommended by pytest: https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#yield-fixtures-recommended

tests/test_esm_runscripts/test_filedicts.py
denizural: What about this? `import esm_runscripts.filedicts as filedicts`

mandresm (approved):

mandresm:
Thanks!

PR #773, opened 2022-08-09
Allowed Missing Files Decorator

pgierz:
Allowed missing files functionality draft. Here, you can mark a method as:

```
@allowed_missing
def my_method(self, *args, **kwargs)
```

and it will just return None instead of running *if* the SimulationFile has an attribute `allowed_to_be_missing: True`. 

For now, still a draft so we can all look and decide if this style of programming works for us or if we all just sit there and scratch our heads.

Notes:
+ [x] Maybe name the decorator the same as the file property, e.g. `@allowed_to_be_missing` and in the file properties the same.


pgierz:
The new functionality here is ready to review. Note that are still breaking tests, and the `path_in_computer` branch is not yet in this one. So, you can look at the new code and comment on it, but there will still be changes when I update with newest parts tomorrow.

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -83,6 +82,67 @@ def test_filedicts_basics(fs):
...
+                path_in_work: .
+                movement_type: move
+    """
+    config = yaml.safe_load(dummy_config)
+    fs.create_dir("/work/data/pool")
+    fs.create_file("/work/data/pool/not_foo_at_all")
+    sim_file = esm_runscripts.filedicts.SimulationFile(
+        config, "echam.files.human_readable_tag_001"
+    )
+    sim_file.mv("pool", "work")
```
mandresm:
```suggestion
    sim_file.mv("computer", "work")
```

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -83,6 +82,67 @@ def test_filedicts_basics(fs):
...
+        exp_dir: "/work/ollie/pgierz/some_exp"
+        thisrun_dir: "/work/ollie/pgierz/some_exp/run_20010101-20010101"
+    computer:
+        pool_dir: "/work/ollie/pool"
+    echam:
+        files:
+            human_readable_tag_001:
+                allowed_to_be_missing: True
+                name_in_pool: foo
+                path_in_pool: /work/data/pool
```
mandresm:
```suggestion
                name_in_computer: foo
                path_in_computer: /work/data/pool
```

mandresm (approved):

PR #776, opened 2022-08-10
enforce absolute path for path_in_computer 

mandresm:
- [x] Changes `pool` for `computer` in our moving-direction logic
- [x] Fixes the broken unit tests
- [x] Checks that the `path_in_computer` is an absolute path and if not reports an user error
- [x] Unit test for the absolute path check

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -209,7 +209,7 @@ class SimulationFile(dict):
                 "File Dictionaries",
                 "The path defined for "
                 f"``{self.component}.files.{self.name}.path_in_computer`` is not "
-                "abosulte. Please, always define an absolute path for the "
+                "absolute. Please, always define an absolute path for the "
                 "``path_in_computer`` variable."
             )
 
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -21,7 +21,7 @@ class SimulationFile(dict):
         echam:
             files:
                 jan_surf:
-                    name_in_pool: T63CORE2_jan_surf.nc
+                    name_in_computer: T63CORE2_jan_surf.nc
                     name_in_work: unit.24
                     filetype: NetCDF
                     description: >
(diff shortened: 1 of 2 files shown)


mandresm:
I'm sure there is a better way for capturing a sys.exit() than the `Capture` class I added on the testing file... if you have ideas on how to do that in a better way please let me know

pgierz (approved):
Looks great to me. 

The `Capturing` class looks very similar to something I stole for `test_namelist` (a test routine which I apparently wrote at some point?!). There are many things I do not understand about it though and it looks very magical.

@denizural, you have read more about TDD than I have, maybe there is a good reason to instead raise something like `FileDictError` here (which we don't have yet but should probably make at some point?), but for me the check @mandresm has looks entirely reasonable.

Overall 👍🏻

mandresm:
> The `Capturing` class looks very similar to something I stole for `test_namelist` (a test routine which I apparently wrote at some point?!). There are many things I do not understand about it though and it looks very magical.

I wrote the last namelist unit test one a month ago or so, I also found the `Capturing` class online, that's why I decided to recycle it here :) First hands-on with unit tests for me

mandresm:
The reason for me using `user_error` here is because here we are going to be communicating with the user, that is the whole point of that check.

pgierz:
> > The `Capturing` class looks very similar to something I stole for `test_namelist` (a test routine which I apparently wrote at some point?!). There are many things I do not understand about it though and it looks very magical.
> 
> I wrote the last namelist unit test one a month ago or so, I also found the `Capturing` class online, that's why I decided to recycle it here :) First hands-on with unit tests for me

Apparently I reviewed it though! I thought that looked familiar somehow...

mandresm:
> 

Indeed, you did not only reviewed it but you gave me a nice intro course to the unit tests :)

pgierz:
I somehow have no memory of this. Perhaps I need to go see the head doctor after the sprint is done.

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -164,28 +181,58 @@ def test_mv(fs):
...
+                path_in_computer: pool/ECHAM/T63
+        thisrun_work_dir: "/work/ollie/pgierz/some_exp/run_20010101-20010101/work"
+    """
+    config = yaml.safe_load(dummy_config)
+
+    # Captures output (i.e. the user-friendly error)
+    with Capturing() as output:
+        try:
+            sim_file = esm_runscripts.filedicts.SimulationFile(config, "echam.files.jan_surf")
+        except SystemExit as e:
```
denizural: You can use `pytest.raises` context manager there. For example usage you can have a look at my test. Apart from that looks good.
```python
# or ValueError
with pytest.raises(SystemExit):
    # I would also parametrize the hard coded file name
    sim_file = esm_runscripts.filedicts.SimulationFile(config, "echam.files.jan_surf")
```

denizural (changes requested):

src/esm_runscripts/filedicts.py
```diff
@@ -196,6 +203,16 @@ def path_type(self, path: pathlib.Path) -> str or bool:
         else:
             raise Exception(f"Cannot identify the path's type of {path}")
 
+    def _check_path_in_computer_is_abs(self):
+        if not self.path_in_computer.is_absolute():
+            user_error(
+                "File Dictionaries",
+                "The path defined for "
+                f"``{self.component}.files.{self.name}.path_in_computer`` is not "
+                "abosulte. Please, always define an absolute path for the "
```
denizural: typo: absolute

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -164,28 +181,58 @@ def test_mv(fs):
...
+            error = e
+
+    # error needs to occur as the path is not absolute
+    assert isinstance(error, SystemExit)
+    assert any(["ERROR: File Dictionaries" in line for line in output])
```
denizural, line 235: I thinks a single assert is enough. Probably, nothing will get printed after the first assert. This also drops the need for the `Capturing` class

denizural (changes requested):

tests/test_esm_runscripts/test_filedicts.py
mandresm: Thanks! I knew you say that on a meeting, I just could not understand the exact syntax, so I recycled the Capturing class. I will change this to your solution.

PR #779, opened 2022-08-10
Sprint/filedicts/linking

denizural:
This PR implements symbolic linking functionality. All tests are passing. 
I also tried it with manual tests.
Docstring is also provided.


src/esm_runscripts/filedicts.py
```diff
@@ -65,8 +67,70 @@ def __init__(self, full_config, attrs_address):
     def cp(self) -> None:
         pass
 
-    def ln(self) -> None:
-        pass
+    def ln(
+        self, source_path: Union[AnyStr, os.PathLike], destination_path: str
```
mandresm: Again, do we want real paths here or are we just gonna give `"work"`, `"in_exp_tree"`, ... strings to point at the keys in the files dictionary to be used as paths. I don't see the need of using paths here, specially since this is not a general function but a method inside a class that defines the paths a properties in the initialization.

mandresm (commented):
My only comment is that I would not pass paths to the method. I would just pass the keys that specify the directions of the linking, because all the path information is given by the user in the `files` properties, and we are always going to be using those, not our own built paths.

This approach will change a lot of code here. 

There is also a lot of checks in the `ln` method, many of which are already included in the `check_source_and_target` method, but this is something we can unify later on today.

pgierz:
I'm happy to put this in once the fixes for `source_path` being either a `str` *or* a `pathlib.Path` are finished to be changed to only accept `str`. I believe we agreed on keeping strings so we can use them to look up canonical locations in one of the object attributes. There are both tests and source code that needs to change for that to happen.

And there are also still unfortunately merge conflicts. I also some specific change questions, please see the review.

src/esm_runscripts/__init__.py
```diff
@@ -14,6 +14,7 @@
 from .last_minute import *
 from .dataprocess import *
 from .filelists import *
+from .filedicts import *
```
pgierz, line 12: I do not understand why this is needed here at the moment. In any case, wildcard "import everything" from a module is not a concept I would promote in the new functionality; I would prefer explicit and verbose imports, even if it means more typing.

src/esm_runscripts/filedicts.py
pgierz: Why is `AnyStr` needed here as opposed to just regular `str`? Is that a Python version difficulty?

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -155,3 +182,49 @@ def test_mv(fs):
...
+# + check if destination is written
+# + check if destination already exists or a symlink
+# + check for self pointing
+# + check if source path does not exist
+# + check if destination path does not exist
+# ===
+def test_ln_iscallable(simulation_file):
+    assert isinstance(simulation_file.ln, Callable)
+
+def test_ln_links_files_correctly(simulation_file, fs):
```
pgierz: Question: does the order of the fixtures you use as function arguments matter? Or is that arbitrary?

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -155,3 +182,49 @@ def test_mv(fs):
...
+# ===
+# Tests for SimulationFile.ln() method
+# + check if source exists
+# + check if destination is written
+# + check if destination already exists or a symlink
+# + check for self pointing
+# + check if source path does not exist
+# + check if destination path does not exist
+# ===
+def test_ln_iscallable(simulation_file):
```
pgierz, line 281: I do not understand why this test is necessary.

src/esm_runscripts/__init__.py
denizural: Actually, we don't even need this I guess. `from esm_runscripts import <module>` should do the trick. Maybe we can have a blank `__init__` file like many modern projects. Currently, `__init__.py` is no longer needed to make a directory module.

src/esm_runscripts/filedicts.py
denizural: This is resolved now

denizural: I think it is better use this one since it comes from the typing moduel.

tests/test_esm_runscripts/test_filedicts.py
denizural: We can remove that actually. Since I write the test first and then the code as a part of the red->green->refactor cycle, I test first if the function is callable.

PR #780, opened 2022-08-10
Fix/echam default namelists ssp

mandresm:
Adds default namelists for ECHAM and JSBACH for `cmip6` type simulations without `no_dynveg` for AWICM1 (ECHAM version `6.3.04p1`)

Solves some of the issues on #730


antoniofis10000 (approved):

mandresm:
@pgierz, can you review this namelists for ssp scenarios?

namelists/echam/6.3.04p1/cmip6/namelist.echam
```diff
@@ -0,0 +1,139 @@
+&parctl
+	nproca=24
+	nprocb=48
+/
+&runctl
+	NPROMA=72
```
pgierz, line 6: Careful here: this might have influence on the comparability of two runs if the number is different. From the handbook, this controls: "vector length of calculations in grid point space"

 https://www.cen.uni-hamburg.de/en/icdc/data/atmosphere/docs-atmo/echam6-userguide.pdf

namelists/echam/6.3.04p1/cmip6/namelist.echam
```diff
@@ -0,0 +1,139 @@
...
+/
+&radctl
+	iaero = 9
+	io3 = 4
+	isolrad = 1
+	ico2 = 4
+	ighg = 1
+	icfc = 4
+	ich4 = 4
+	in2o = 4
```
pgierz, line 134: I am not sure about these settings here. If they are what is recommended by MPI-Met, then it's good... One could double check using the MPI Infrastructure toolkit `mkexp` and see values are used there.

pgierz (commented):
Looks good; I won't explicitly give a green check mark since there are some things you may want to double check to ensure that runs using these namelists are correctly configured. See the specific comments for that.

namelists/echam/6.3.04p1/cmip6/namelist.echam
mandresm: I agree with this, in all the namelist I've seen, also the ones of @chrisdane, `nproma: 8`. @antoniofis10000, can you check that for Tido's simulation NPROMA is 72? 

I would suggest to set NPROMA to 8 by default. Then you can add to your runscript if Tido's simulations had that:
```yaml
echam:
     add_namelist_changes:
         namelist.echam:
            runctl:
                nproma: 72
```

PR #781, opened 2022-08-11
added support for parallel I/O in ECHAM6

seb-wahl:
A feature I added locally a while ago and never bothered to push it. With ECHAM6-HAMMOZ in our old environment we heavily used this feature to speed up simulations by allowing parallel I/O. Speedup (if any) depends on the machine, the file system and a lot of other things I don't understand. And on some machines it just doesn't work. To use it add
```yaml
general:
  echam:
      nprocio: 6
      namelist_changes:
         namelist.echam:
           parctl:
              iomode: 2
              nprocio: ${nprocio}
```
`iomode = 1` is another way of writing parallel I/O. You need to test on your machine whether `1` or `2` works. `0` is the default and means no parallel I/O. I don't know whether it works if you use netCDF output with ECHAM6.


mandresm:
For some reason #643 was never merged but we made some suggestions to get rid of the additional namelist changes in the runscript. Something was still failing in the tests, but I say we cherry pick b97e0a0 and try again with that, if you agree with those changes.

seb-wahl:
> For some reason #643 was never merged but we made some suggestions to get rid of the additional namelist changes in the runscript. Something was still failing in the tests, but I say we cherry pick [b97e0a0](https://github.com/esm-tools/esm_tools/commit/b97e0a007f23192a3b79eaaeee0fb2bb68d9a1ca) and try again with that, if you agree with those changes.

Perfect. In the end you just need to add the those few lines in `batch_system.py`, and that's it.

mandresm (approved):

mandresm:
#bump #patch

PR #788, opened 2022-08-15
Datestamps

pgierz:
Allows inclusion of datestamps in the `target` of a movement, implemented as a decorator.

A `sim_file` object now has two attributes:
1. `datestamp_method`: controls how the date stamp is (possibly) added. Either `never`, `always`, or `check_overwrite`
2. `datestamp_format`: controls how the date stamp is formatted. If `check_extract`, it will not add a date stamp if it is unique in the target. If it is `append`, it will always append a date stamp.

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -172,7 +172,7 @@ class SimulationFile(dict):
         return super().__setattr__(name, value)
 
     def __setitem__(self, key: Any, value: Any) -> None:
-        # Checks for changing with sim_file["my_key"] = "new_value"
+        """Checks for changing with sim_file["my_key"] = "new_value""""
         if key == "datestamp_format":
             self._check_datestamp_format_is_allowed(value)
         if key == "datestamp_method":
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -15,8 +15,9 @@ import shutil
 from typing import Any, AnyStr, Tuple, Type, Union
 
 import dpath.util
-from esm_parser import ConfigSetup, user_error
 from loguru import logger
+
+from esm_parser import ConfigSetup, user_error
 
 
 # NOTE(PG): Comment can be removed later. Here I prefix with an underscore as
@@ -70,6 +71,23 @@ def _allowed_to_be_missing(method):
     return inner_method
 
 
+def _fname_has_date_stamp_info(fname, date, reqs=["%Y", "%m", "%d"]):
+    date_attrs = {
+        "%Y": "syear",
+        "%m": "smonth",
+        "%d": "sday",
+        "%H": "shour",
+        "%M": "sminute",
+        "%S": "ssecond",
+    }
+    required_attrs = [getattr(date, v) for k, v in date_attrs.items() if k in reqs]
+    # all(attr in fname for attr in required_attrs)
+    for attr in required_attrs:
+        if attr in fname:
+            fname = fname.replace(attr, "checked", 1)
+    return fname.count("checked") == len(reqs)
+
+
 class SimulationFile(dict):
     """
     Desribes a file used within a ESM Simulation.
@@ -115,6 +133,7 @@ class SimulationFile(dict):
         )
         super().__init__(attrs_dict)
         self._config = full_config
+        self._sim_date = full_config["general"]["current_date"]
         self.name = attrs_address.split(".")[-1]
         self.component = component = attrs_address.split(".")[0]
         self.path_in_computer = pathlib.Path(self["path_in_computer"])
@@ -224,6 +243,12 @@ class SimulationFile(dict):
         source_path = self[f"absolute_path_in_{source}"]
         target_path = self[f"absolute_path_in_{target}"]
 
+        # Datestamps
+        if self.datestamp_method == "always":
+            target_path = self._always_datestamp(target_path)
+        if self.datestamp_method == "avoid_overwrite":
+            target_path = self._avoid_override_datestamp(target_path)
+
         # Checks
         self._check_source_and_target(source_path, target_path)
 
@@ -273,6 +298,15 @@ class SimulationFile(dict):
         source_path = self[f"absolute_path_in_{source}"]
         target_path = self[f"absolute_path_in_{target}"]
 
+        # This will need to be moved once Deniz implements this function to be
+        # in the same style as the others.
+        #
+        # Datestamps
+        if self.datestamp_method == "always":
+            target_path = self._always_datestamp(target_path)
+        if self.datestamp_method == "avoid_overwrite":
+            target_path = self._avoid_override_datestamp(target_path)
+
         points_to_itself = source_path == target_path
         if points_to_itself:
             err_msg = (
@@ -321,9 +355,18 @@ class SimulationFile(dict):
             raise ValueError(
                 f"Source is incorrectly defined, and needs to be in {self.locations}"
             )
+        if target not in self.locations:
+            raise ValueError(
+                f"Target is incorrectly defined, and needs to be in {self.locations}"
+            )
         source_path = self[f"absolute_path_in_{source}"]
         target_path = self[f"absolute_path_in_{target}"]
 
+        # Datestamps
+        if self.datestamp_method == "always":
+            target_path = self._always_datestamp(target_path)
+        if self.datestamp_method == "avoid_overwrite":
+            target_path = self._avoid_override_datestamp(target_path)
         # Checks
         self._check_source_and_target(source_path, target_path)
 
@@ -361,7 +404,7 @@ class SimulationFile(dict):
 
     Notes on possible datestamp formats
     -----------------------------------
-    from_filename : str
+    check_from_filename : str
         This option will add a datestamp to a file, if the year, month, and day
         cannot be extracted from the standard declared filename.
     append : str
@@ -440,6 +483,38 @@ class SimulationFile(dict):
             return False
         else:
             raise Exception(f"Cannot identify the path's type of {path}")
+
+    def _always_datestamp(self, fname):
+        if self.datestamp_format == "append":
+            return pathlib.Path(f"{fname}_{self._sim_date}")
+        if self.datestamp_format == "check_from_filename":
+            if _fname_has_date_stamp_info(fname, self._sim_date):
+                return fname
+            else:
+                return pathlib.Path(f"{fname}_{self._sim_date}")
+
+    def _avoid_override_datestamp(self, target: pathlib.Path) -> pathlib.Path:
+        """
+        If source and target are identical, adds the date stamp to the target.
+
+        This method is used in the case that the object's attribute
+        ``datestamp_method`` is set to ``avoid_overwrite``, and is called
+        before the checks of each of ln, cp, and mv.
+
+        Parameters
+        ----------
+        target : pathlib.Path
+
+        Returns
+        -------
+        pathlib.Path :
+            The new target that can be used
+        """
+        if target.exists():
+            if self.datestamp_format == "append":
+                target = pathlib.Path(f"{target}_{self._sim_date}")
+            # The other case ("check_from_filename") is meaningless?
+        return target
 
     def _check_path_in_computer_is_abs(self):
         if not self.path_in_computer.is_absolute():
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -83,7 +83,9 @@ def _fname_has_date_stamp_info(fname, date, reqs=["%Y", "%m", "%d"]):
     date : esm_calendar.Date
         The date to be checked against
     reqs : list of str
-        A list of ``strftime`` compliant strings to determine which elements of the date to check. Compatible with %Y %m %d %H %M %S (year, month, day, hour, minute, second)
+        A list of ``strftime`` compliant strings to determine which elements of
+        the date to check. Compatible with %Y %m %d %H %M %S (year, month, day,
+        hour, minute, second)
         
     Returns
     -------
@@ -181,6 +183,10 @@ class SimulationFile(dict):
         # Checks
         self._check_path_in_computer_is_abs()
 
+    ##############################################################################################
+    # Overrides of standard dict methods
+    ##############################################################################################
+
     def __setattr__(self, name: str, value: Any) -> None:
         """Checks when changing dot attributes for disallowed values"""
         if name == "datestamp_format":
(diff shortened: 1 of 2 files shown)


denizural:
Looks good so far

pgierz:
Regarding the broken tests:

What follows is a copy from the CI. 

Functionality from @denizural:
```
FAILED tests/test_esm_runscripts/test_filedicts.py::test_ln_raises_exception_when_pointing_to_itself
[343](https://github.com/esm-tools/esm_tools/runs/7878295110?check_suite_focus=true#step:7:344)
FAILED tests/test_esm_runscripts/test_filedicts.py::test_ln_raises_exception_when_target_is_a_directory_and_not_a_file
[344](https://github.com/esm-tools/esm_tools/runs/7878295110?check_suite_focus=true#step:7:345)
FAILED tests/test_esm_runscripts/test_filedicts.py::test_ln_raises_exception_when_target_path_exists
```
This relates to the fact that I had to move the stamping above your checks. If you have the same style for the `ln` function as we do for `cp` and `mv` that might work.

These ones from @mandresm I need to look at in detail:
```
FAILED tests/test_esm_runscripts/test_filedicts.py::test_check_file_syntax_type_missing
[345](https://github.com/esm-tools/esm_tools/runs/7878294985?check_suite_focus=true#step:7:346)
FAILED tests/test_esm_runscripts/test_filedicts.py::test_check_file_syntax_type_incorrect
[346](https://github.com/esm-tools/esm_tools/runs/7878294985?check_suite_focus=true#step:7:347)
FAILED tests/test_esm_runscripts/test_filedicts.py::test_check_file_syntax_input
[347](https://github.com/esm-tools/esm_tools/runs/7878294985?check_suite_focus=true#step:7:348)
FAILED tests/test_esm_runscripts/test_filedicts.py::test_check_file_syntax_output
```

pgierz:
The Miguel breakages are fixed.

**NOTE** I needed to add a new "hard-requirement" to get the date stamps to work. The objects obviously need to know what the date is for that to work 😉

src/esm_runscripts/filedicts.py
```diff
@@ -73,6 +72,23 @@ def inner_method(self, *args, **kwargs):
     return inner_method
 
 
+def _fname_has_date_stamp_info(fname, date, reqs=["%Y", "%m", "%d"]):
```
mandresm, line 91: Missing docstrings

src/esm_runscripts/filedicts.py
```diff
@@ -73,6 +72,23 @@ def inner_method(self, *args, **kwargs):
...
+        "%Y": "syear",
+        "%m": "smonth",
+        "%d": "sday",
+        "%H": "shour",
+        "%M": "sminute",
+        "%S": "ssecond",
+    }
+    required_attrs = [getattr(date, v) for k, v in date_attrs.items() if k in reqs]
+    # all(attr in fname for attr in required_attrs)
+    for attr in required_attrs:
```
mandresm, line 122: I think this attributes will be resolved by the time we call filedicts in the recipe. Not resolving variables before that will likely generate all sorts of problems. So I guess this function will not do anything in a real word scenario?

src/esm_runscripts/filedicts.py
```diff
@@ -146,6 +163,22 @@ def __init__(self, full_config: dict, attrs_address: dict):
         # Checks
         self._check_path_in_computer_is_abs()
 
+    def __setattr__(self, name: str, value: Any) -> None:
+        # Checks when changing dot attributes for disallowed values:
```
```suggestion
        """Checks when changing dot attributes for disallowed values"""
```

src/esm_runscripts/filedicts.py
```diff
@@ -146,6 +163,22 @@ def __init__(self, full_config: dict, attrs_address: dict):
...
+        if name == "datestamp_format":
+            self._check_datestamp_format_is_allowed(value)
+        if name == "datestamp_method":
+            self._check_datestamp_method_is_allowed(value)
+        return super().__setattr__(name, value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        # Checks for changing with sim_file["my_key"] = "new_value"
```
```suggestion
        """Checks for changing with sim_file["my_key"] = "new_value""""
```

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -537,27 +560,56 @@ def test_datestamp_format_attr(simulation_file):
...
 def test_datestamp_not_added_if_in_filename(simulation_file):
-    assert False
+    pass  # Still to be implemented
+
+
+def test_fname_has_date_stamp_info():
+    fname = "blah_2000-01-01.nc"
+    fname2 = "blah_2000-01.nc"
+    date = esm_calendar.Date("2000-01-01T00:00:00")
+    assert filedicts._fname_has_date_stamp_info(fname, date) is True
```
denizural: Be careful. not `is bool` but `== bool`

denizural (changes requested):

mandresm (changes requested):
I see one major problem:
For many files our target in a given moment will be the source in the next step (restarts in work folder are moved to the experiment folder, and from there to the next run work folder). Here, it might be the case that the file is time stamped, but then when it changes direction no file exists, because the source name is not time-stamped.

This was not happening with the old `filelists` due to the linking approach (you not only timestamp but you also make a link to that timestamped file with the original name). I find this approach pretty horrible and mega error prone, so I'm not suggesting to use it.

Maybe the solution is to check sources existence in the case of not timestamped and also timestamped?

src/esm_runscripts/filedicts.py
pgierz: ...and type hints. I'll take care of that.

src/esm_runscripts/filedicts.py
pgierz: Probably not; I pulled this one out because putting it into the other function was making it longer than I wanted.

tests/test_esm_runscripts/test_filedicts.py
pgierz: Your suggestion is an anti-pattern. See here: https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html#compare-values-to-true-using-the-pattern-if-cond-is-true

src/esm_runscripts/filedicts.py
pgierz: I should note: the overall goal of the functionality was: don't add a "redundant" date stamp if you can already uniquely identify the date of the file from the filename. Probably this never happened in the old version, so probably this one was overkill.

mandresm: I'm fine with including it anyway. We might need it later on.

tests/test_esm_runscripts/test_filedicts.py
denizural: Sorry my bad. But the document says that `is True` was also as bad according to PEP8.

pgierz:
> I see one major problem: For many files our target in a given moment will be the source in the next step (restarts in work folder are moved to the experiment folder, and from there to the next run work folder). Here, it might be the case that the file is time stamped, but then when it changes direction no file exists, because the source name is not time-stamped.
> 
> This was not happening with the old `filelists` due to the linking approach (you not only timestamp but you also make a link to that timestamped file with the original name). I find this approach pretty horrible and mega error prone, so I'm not suggesting to use it.
> 
> Maybe the solution is to check sources existence in the case of not timestamped and also timestamped?

@mandresm, isn't that a problem for the next phase though? This would happen if the files are "double" time-stamped. I am not quite sure when that would happen. Restart files anyway always get a date somehow in their names, or are you talking about forcings are restarts?

tests/test_esm_runscripts/test_filedicts.py
denizural: I think `is` opeator is for comparing the identities (eg. `id()`) of an object and `==` for the logical comparison.

pgierz: From the document:
> The code below uses the PEP 8 preferred pattern of `if condition:`. If the type of the condition is Boolean, it is obvious that comparing to True is redundant.

So, try fixing it with the suggestion feature! 😃

pgierz: > I think is opeator `is` for comparing the identities (eg. `id()`) of an object and `==` for the logical comparison.

Correct! 👍🏻

denizural:
```suggestion
    assert filedicts._fname_has_date_stamp_info(fname, date)
```

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -633,3 +671,75 @@ def test_resolve_paths_old_config():
...
+def test_datestamp_not_added_if_in_filename(simulation_file):
+    pass  # Still to be implemented
+
+
+def test_fname_has_date_stamp_info():
+    fname = "blah_2000-01-01.nc"
+    fname2 = "blah_2000-01.nc"
+    date = esm_calendar.Date("2000-01-01T00:00:00")
+    assert filedicts._fname_has_date_stamp_info(fname, date) is True
+    assert filedicts._fname_has_date_stamp_info(fname2, date) is False
```
denizural:
```suggestion
    assert not filedicts._fname_has_date_stamp_info(fname2, date)
```

mandresm:
> @mandresm, isn't that a problem for the next phase though? This would happen if the files are "double" time-stamped. I am not quite sure when that would happen. Restart files anyway always get a date somehow in their names, or are you talking about forcings are restarts?

I think this will happen for every restart that has to be time stamped by us, `OpenIFS` generates non-time stamped restart files, for example.

mandresm:
FESOM also does something similar for monthly restarts. They are time stamped yearly. The restart will go from the work to the experiment, be changed the name and then the source for the next run won't be found.

pgierz:
@mandresm I think we need to put both of the cases you just mentioned as ~unit~ behavior tests. Not being familiar with the AWICM3 way of moving files around, I might need some help to design those tests correctly.

EDIT: those are actually real world "behavior" tests (for lack of a better word)

mandresm:
agreed, maybe we can simply open a ticket for that, and leave it for later?

src/esm_runscripts/filedicts.py
```diff
@@ -119,6 +135,7 @@ def __init__(self, full_config: dict, attrs_address: dict):
         super().__init__(attrs_dict)
         self._original_filedict = copy.deepcopy(attrs_dict)
         self._config = full_config
+        self._sim_date = full_config["general"]["current_date"]
```
```suggestion
        self._sim_date = full_config["general"]["current_date"] # NOTE: we might have to change this in the future, depending on whether SimulationFile is access through tidy ("end_date") or prepcompute ("start_date")
```

src/esm_runscripts/filedicts.py
```diff
@@ -146,6 +163,22 @@ def __init__(self, full_config: dict, attrs_address: dict):
...
+            self._check_datestamp_method_is_allowed(value)
+        return super().__setattr__(name, value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        """Checks for changing with sim_file['my_key'] = 'new_value'"""
+        if key == "datestamp_format":
+            self._check_datestamp_format_is_allowed(value)
+        if key == "datestamp_method":
+            self._check_datestamp_method_is_allowed(value)
+        return super().__setitem__(key, value)
```
mandresm, line 312: This is a really good idea :D, I'll keep these methods in mind for when we cleanup, there might be some variables that we want to check here, instead of in the __init__

src/esm_runscripts/filedicts.py
```diff
@@ -158,6 +191,42 @@ def allowed_to_be_missing(self):
...
+        Defines which datestamp_format shall be used when possibly including
+        date stamps to the file. Valid choices are "check_from_filename" and
+        "append".
+        """
+        datestamp_format = self.get(
+            "datestamp_format", "append"
+        )  # This is the old default behaviour
+        return datestamp_format
+
+    def update(self, *args, **kwargs):
```
mandresm: Can we move this up, close the `__setitem__` method? Then we have all the modified `dict` methods together.

src/esm_runscripts/filedicts.py
```diff
@@ -158,6 +191,42 @@ def allowed_to_be_missing(self):
...
+        """
+        Standard dictionary update method, enhanced by additional safe-guards
+        for particular values.
+        """
+        for k, v in dict(*args, **kwargs).items():
+            if k == "datestamp_format":
+                self._check_datestamp_format_is_allowed(v)
+            if k == "datestamp_method":
+                self._check_datestamp_method_is_allowed(v)
+            self[k] = v
```
mandresm: Is it okay to not use `super().update` here? I ask because I am lacking the knowledge on whether `self[k] = v` is identical to `self.update({k: v})`

mandresm (approved):
Just a few comments and changes, but otherwise ready to merge

src/esm_runscripts/filedicts.py
pgierz: Short answer: it is. What update basically does is: if a key exists, replace it with this new value. If a key does not exist, add it with this new value. So it's basically the same.

src/esm_runscripts/filedicts.py
pgierz: Sure. My organisation was double underscore methods, properties, regular methods without underscore, and then private methods with one underscore. But I don't really care.

src/esm_runscripts/filedicts.py
pgierz: Thanks! That was one of the unit tests I wrote down and thought "oh if we don't check this it will probably bite us in the butt somehow"

src/esm_runscripts/filedicts.py
pgierz, line 91: Missing docstring
```suggestion
def _fname_has_date_stamp_info(fname, date, reqs=["%Y", "%m", "%d"]):
    """
    Checks if a particular file has all elements of a particular date in its name.
    
    Parameters
    ----------
    fname : str
        The name of the file to check
    date : esm_calendar.Date
        The date to be checked against
    reqs : list of str
        A list of ``strftime`` compliant strings to determine which elements of the date to check. Compatible with %Y %m %d %H %M %S (year, month, day, hour, minute, second)
        
    Returns
    -------
    bool : 
        True if all elements appear in the filename, False otherwise.
    
    """
```

src/esm_runscripts/filedicts.py
pgierz: There is still a docstring missing though, which I will add. Also I am sure this function will misbehave with files that happen to have `01` or something in their name anyway. Or if someone uses a crazy `expid`...but let's just be optimistic and assume that will not happen.

PR #789, opened 2022-08-15
syntax checker

mandresm:
- [x] Checks:
   - [x] Did the user specify `type`?
   - [x] Is the `type` correct?
   - [x] and `path_in_computer`?
   - [x] and `name_in_computer`?
   - [x] and `name_in_work`?
- [x] Docstrings 
- [x] Unit test

The error reported back to the user looks like this:
Screen Shot 2022-08-15 at 14 54 44

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -349,9 +349,22 @@ class SimulationFile(dict):
             raise Exception(f"Cannot identify the path's type of {path}")
 
     def _check_file_syntax(self):
+        """
+        Checks for missing variables:
+        - ``type``
+        - ``path_in_computer`` if the file it an input for the experiment
+        - ``name_in_computer`` if the file it an input for the experiment
+        - ``name_in_work`` if the file it an output of the experiment
+
+        It also checks whether ``type``'s value is correct.
+
+        It notifies the user about this errors in the syntacm using
+        ``esm_parser.error``.
+        """
         error_text = ""
         missing_vars = ""
         types_text = ", ".join(self.all_model_filetypes)
+        this_filedict = copy.deepcopy(self._original_filedict)
         input_file_types = ["config", "forcing", "input"]
         output_file_types = [
             "analysis", "couple", "log", "mon", "outdata", "restart", "viz", "ignore"
@@ -374,13 +387,8 @@ class SimulationFile(dict):
                     f"(``files.{self.name}.type``), please choose one of the following "
                     f"types: {types_text}\n"
                 )
-                self._original_filedict["type"] = f"``{self._original_filedict['type']}``"
+                this_filedict["type"] = f"``{this_filedict['type']}``"
 
-        missing_vars = (
-            f"Please, complete/correct the following vars for your file:\n\n"
-            f"{self.pretty_original_filedict()}"
-            f"{missing_vars}"
-        )
         if "path_in_computer" not in self.keys() and self.get("type") in input_file_types:
             error_text = (
                 f"{error_text}"
@@ -415,15 +423,18 @@ class SimulationFile(dict):
                 f"{missing_vars}    ``name_in_work``: <name_of_file_in_work_dir>\n"
             )
 
+        missing_vars = (
+            f"Please, complete/correct the following vars for your file:\n\n"
+            f"{self.pretty_filedict(this_filedict)}"
+            f"{missing_vars}"
+        )
+
         if error_text:
             error_text = (
                 f"The file dictionary ``{self.name}`` is missing relevant information "
                 f"or is incorrect:\n{error_text}"
             )
             user_error("File Dictionaries", f"{error_text}\n{missing_vars}")
-
-    def pretty_original_filedict(self):
-        return yaml.dump({"files": {self.name: self._original_filedict}})
 
     def _check_path_in_computer_is_abs(self):
         if not self.path_in_computer.is_absolute():
@@ -467,6 +478,22 @@ class SimulationFile(dict):
             raise Exception(
                 f"Target directory ``{target_path_parent_type}`` does not exist!"
             )
+
+    def pretty_filedict(self, filedict):
+        """
+        Returns a string in yaml format of the given file dictionary.
+
+        Parameters
+        ----------
+        dict
+            A file dictionary
+
+        Returns
+        -------
+        str
+            A string in yaml format of the given file dictionary
+        """
+        return yaml.dump({"files": {self.name: filedict}})
 
 
 def copy_files(config):
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -8,10 +8,12 @@ Developer Notes
 * Decorators should have names that map to an attribute of the object. See the
   example in ``_allowed_to_be_missing``.
 """
+import copy
 import functools
 import os
 import pathlib
 import shutil
+import yaml
 from typing import AnyStr, Tuple, Type, Union
 
 import dpath.util
@@ -73,7 +75,7 @@ def _allowed_to_be_missing(method):
 
 class SimulationFile(dict):
     """
-    Desribes a file used within a ESM Simulation.
+    Describes a file used within a ESM Simulation.
 
     Given a config, you should be able to use this in YAML::
 
@@ -115,10 +117,24 @@ class SimulationFile(dict):
             full_config, attrs_address, separator=".", default={}
         )
         super().__init__(attrs_dict)
+        self._original_filedict = copy.deepcopy(attrs_dict)
         self._config = full_config
         self.name = attrs_address.split(".")[-1]
         self.component = component = attrs_address.split(".")[0]
         self.path_in_computer = pathlib.Path(self["path_in_computer"])
+        self.all_model_filetypes = full_config["general"].get("all_model_filetypes", [
+            "bin",
+            "config",
+            "forcing",
+            "input",
+            "couple",
+            "log",
+            "outdata",
+            "restart",
+            "ignore",
+        ])
+
+        self._check_file_info_is_complete()
 
         # Complete tree names if not defined by the user
         self["name_in_run_tree"] = self.get(
@@ -340,6 +356,35 @@ class SimulationFile(dict):
             return False
         else:
             raise Exception(f"Cannot identify the path's type of {path}")
+
+    def _check_file_info_is_complete(self):
+        error_text = ""
+        missing_var = (
+            f"Please, complete the following vars for your file:\n\n"
+            f"{self.pretty_original_filedict}"
+        )
+        types_text = ", ".join(self.all_model_filetypes)
+
+        if "type" not in self.keys():
+            error_text = (
+                f"- the ``type`` variable is missing. Please define a ``type`` "
+                f"({types_text})\n"
+            )
+            missing_var = (
+                f"{missing_var}    ``type``: forcing/input/restart/outdata/...\n"
+            )
+
+        if error_text:
+            error_text = (
+                f"The file ``{self.name}`` is missing relevant information:\n{error_text}"
+            )
+            user_error("File Dictionaries", f"{error_text}\n{missing_var}")
+
+    @property
+    def pretty_original_filedict(self):
+        original_filedict_str = yaml.dump({"files": {self.name: self._original_filedict}})
+
+        return original_filedict_str
 
     def _check_path_in_computer_is_abs(self):
         if not self.path_in_computer.is_absolute():
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -379,15 +379,14 @@ class SimulationFile(dict):
             missing_vars = (
                 f"{missing_vars}    ``type``: forcing/input/restart/outdata/...\n"
             )
-        else:
-            if self["type"] not in self.all_model_filetypes:
-                error_text = (
-                    f"{error_text}"
-                    f"- ``{self['type']}`` is not a supported ``type`` "
-                    f"(``files.{self.name}.type``), please choose one of the following "
-                    f"types: {types_text}\n"
-                )
-                this_filedict["type"] = f"``{this_filedict['type']}``"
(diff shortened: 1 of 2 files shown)


tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -412,6 +412,32 @@ def test_ln_raises_exception_when_target_path_does_not_exist(simulation_file, fs
...
+                name_in_computer: T63CORE2_jan_surf.nc
+                name_in_work: unit.24
+                path_in_computer: /work/ollie/pool/ECHAM/T63/
+        experiment_input_dir: /work/ollie/pgierz/some_exp/input/echam
+        thisrun_input_dir: /work/ollie/pgierz/some_exp/run_20010101-20010101/input/echam
+    """
+    config = yaml.safe_load(dummy_config)
+
+    # Captures output (i.e. the user-friendly error)
+    with Capturing() as output:
```
denizural, line 440: We don't need `Capturing`. Simply `pytest.raises(SystemExit)` does the job. You can have a look at https://github.com/esm-tools/esm_tools/pull/790

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -412,6 +412,32 @@ def test_ln_raises_exception_when_target_path_does_not_exist(simulation_file, fs
 
 # ========== end of ln() tests ==========
 
+def test_check_file_info_is_complete():
```
denizural: To keep the test case short, the fixture can be used. I already provided in the first commits.

src/esm_runscripts/filedicts.py
```diff
@@ -341,6 +357,46 @@ def _path_type(self, path: pathlib.Path) -> Union[str, bool]:
...
+            f"{missing_vars}"
+        )
+        if error_text:
+            error_text = (
+                f"The file dictionary ``{self.name}`` is missing relevant information "
+                f"or is incorrect:\n{error_text}"
+            )
+            user_error("File Dictionaries", f"{error_text}\n{missing_vars}")
+
+    @property
```
denizural: Do we really need a `@property` decorator here? It simply prints the output. Not really a getter/setter. Are you planning to add new things to this function?

src/esm_runscripts/filedicts.py
```diff
@@ -115,10 +117,24 @@ def __init__(self, full_config: dict, attrs_address: dict):
...
+            "bin",
+            "config",
+            "forcing",
+            "input",
+            "couple",
+            "log",
+            "outdata",
+            "restart",
+            "ignore",
+        ])
```
denizural: If we are going to use this list again, maybe it would be a good idea to put them into another list like `default_filetypes = ["bin", "forcing", ...]` and use it instead. That would remove the duplications in the future.

denizural (changes requested):

tests/test_esm_runscripts/test_filedicts.py
pgierz: That's incorrect Deniz, we do need it. Look at the assert statement below. Miguel is also checking the heading of the user error.

src/esm_runscripts/filedicts.py
pgierz: That depends on what we want. This provides you a string representation of the Yaml as a attribute of object. 

Making it a callable would require you capturing that in a variable to use it. 

Also in principle the property decorator does not necessitate additionally defining getter/setter behaviour. It _can_ but doesn't need to. Basically, it is used to have a dot-accessible attribute when the attribute needs some kind of logical computation. 

Maybe this is a global style choice we need to agree upon.

tests/test_esm_runscripts/test_filedicts.py
pgierz: This also is a global style choice we need to agree upon.

src/esm_runscripts/filedicts.py
```diff
@@ -115,10 +117,24 @@ def __init__(self, full_config: dict, attrs_address: dict):
...
+
+        self._check_file_info_is_complete()
```
pgierz: I like the function to ensure everything needed to build the file works. My only concern is that we are slowly making the unit routine longer and longer.

tests/test_esm_runscripts/test_filedicts.py
denizural: I think SystemExit is strong enough. Is there a further need for string parsing?

src/esm_runscripts/filedicts.py
```diff
@@ -341,6 +357,46 @@ def _path_type(self, path: pathlib.Path) -> Union[str, bool]:
...
+            f"Please, complete/correct the following vars for your file:\n\n"
+            f"{self.pretty_original_filedict}"
+            f"{missing_vars}"
+        )
+        if error_text:
+            error_text = (
+                f"The file dictionary ``{self.name}`` is missing relevant information "
+                f"or is incorrect:\n{error_text}"
+            )
+            user_error("File Dictionaries", f"{error_text}\n{missing_vars}")
```
pgierz, line 436: This is a requested change: is there a way to check **something** and either use user_error or raise an exception here?

The reason: if you ever want to use this class in your own script, maybe outside of the esm-tools context, the way it is now closely connects and couples it to esm-tools. There are (many) other parts of the code that do this too, that's a generic thing for the entire project, not just this sprint. It's often impossible to just import a class out of esm tools and use it without the whole, emm I guess "bloat"

pgierz (changes requested):
Looks good! I have one specific change but we might need to discuss that as a general style choice to make together.

tests/test_esm_runscripts/test_filedicts.py
mandresm: Yes, the new function will report to the user all the issues of a given file at once, so that the user can correct them, and does not need to discover the syntax errors one by one. Therefore, checking precisely which error needs to be reported it should be part of the test.

src/esm_runscripts/filedicts.py
mandresm: I was just trying out a property for the sake of trying. I don't have an opinion on whether it should be a property or not.

src/esm_runscripts/filedicts.py
mandresm: I don't get this, this should we always be provided by `config["general"]["all_model_filetypes"]`, the main reason for me to use a `get` here, was to avoid problems with the current unit tests, but I will update those once I have done further developments on this feature.

src/esm_runscripts/filedicts.py
mandresm: Do you mean the unit test routine? We could divide it into different routines if that's the case. I see a conflict in between having unit tests for functions + keeping methods minimal on what they do + keeping the unit test routines short

src/esm_runscripts/filedicts.py
mandresm: Maybe we can create a `file_error` routine that wraps `user_error` and if not available raises an exception? I think it is a bit of an over-complication, after all this is built to be used in the context of ESM-Tools file structure and it already needs a lot of variables from the config.

src/esm_runscripts/filedicts.py
denizural: What I mean is that, whenever I see a long list passed to a function, I would give it a name. So that it is both more readable and reusable.

tests/test_esm_runscripts/test_filedicts.py
denizural: I still don't get this. I commented out the lines and only left `SystemExit`
```python
# Captures output (i.e. the user-friendly error)
# with Capturing() as output:
with pytest.raises(SystemExit) as error:
    sim_file = esm_runscripts.filedicts.SimulationFile(
        config, "echam.files.jan_surf"
    )

# error needs to occur as the path is not absolute
# assert any(["ERROR: File Dictionaries" in line for line in output])
```
This still gives Green output. Maybe we can talk about that during the meeting.

pgierz (approved):

tests/test_esm_runscripts/test_filedicts.py
pgierz: That's because you commented out the test. Comment out the function!

pgierz:
@mandresm, you can re-request a review from me once your checklist is ready :-)

tests/test_esm_runscripts/test_filedicts.py
denizural: `pytest.raises` is the **test**. Have a look at this: When I correct the path from `pool/ECHAM/T63` to `/pool/ECHAM/T63`, it yields:
```python
        with pytest.raises(SystemExit) as error:
            sim_file = esm_runscripts.filedicts.SimulationFile(
>               config, "echam.files.jan_surf"
            )
E           Failed: DID NOT RAISE <class 'SystemExit'>

/home/ollie/dural/esm_tools/tests/test_esm_runscripts/test_filedicts.py:469: Failed
======================================================================== short test summary info ========
FAILED test_filedicts.py::test_check_path_in_computer_is_abs - Failed: DID NOT RAISE <class 'SystemExit'>
```

But I will bring back the `Capturing` stuff back so that we are all on the same line.

mandresm: Deniz, check now the new unit tests, I think you'll understand now why we need capturing.

tests/test_esm_runscripts/test_filedicts.py
mandresm: This will be unify on a later PR after we agree on a common strategy

src/esm_runscripts/filedicts.py
mandresm: `@property` removed

src/esm_runscripts/filedicts.py
```diff
@@ -341,6 +358,83 @@ def _path_type(self, path: pathlib.Path) -> Union[str, bool]:
...
+            error_text = (
+                f"{error_text}"
+                f"- the ``type`` variable is missing. Please define a ``type`` "
+                f"({types_text})\n"
+            )
+            missing_vars = (
+                f"{missing_vars}    ``type``: forcing/input/restart/outdata/...\n"
+            )
+        else:
+            if self["type"] not in self.all_model_filetypes:
```
denizural: `elif`?

src/esm_runscripts/filedicts.py
```diff
@@ -341,6 +358,83 @@ def _path_type(self, path: pathlib.Path) -> Union[str, bool]:
...
+            missing_vars = (
+                f"{missing_vars}    ``name_in_work``: <name_of_file_in_work_dir>\n"
+            )
+
+        if error_text:
+            error_text = (
+                f"The file dictionary ``{self.name}`` is missing relevant information "
+                f"or is incorrect:\n{error_text}"
+            )
+            user_error("File Dictionaries", f"{error_text}\n{missing_vars}")
```
denizural, line 436: Is this the correct indentation?

denizural (changes requested):

src/esm_runscripts/filedicts.py
denizural: Sorry my bad. It looks more clear when you view it in the editor instead of GitHub

tests/test_esm_runscripts/test_filedicts.py
pgierz: So, I had thought this was clear last week on Monday when we started, but to definitively write down once how we should be doing this: describe all expected input and all expected output or state changes in the test. Use one single function in between these two to modify the input into the output. Ideally only test one behaviour or state change per test.  

Commenting out part of what the test asserts against (and by the way, at the end of the test) and then wondering why your test does not break doesn't surprise me. You are making the test less comprehensive, so of course it tests less behaviour and is more likely to pass. 

If you wanted to induce a breaking test to discover a bug, you comment out in the function (singular) that modifies the state or expected output, but not in the test. 

If we all as a group decide we need a different broad scale design, then and only then do we redesign and fix the tests. 

But don't let that make you afraid to try stuff out. If you need a new global behaviour and that breaks other peoples stuff, then include in your PR ways to make those previous tests work again. 

With that, enough sprint from me for the day

denizural: Yes, I check that. One thing I still don't understand is this: unit tests are for the developers and not for the users. If the method is raising an exception and my unit test catches that it is fine. I don't need to parse the error string as an extra effort. The end user will see the error message anyways since it is handled by the `user_error` method. This is why I 
1) load the correct yaml file, 
2) instantiate the `simulation_file` object (no exception is raise)
3) inject buggy line into the `simulation_file` object
4) call the method that is expected to fail / raise exception.

But anyways, this is a difference between the approaches. If it works, it is good.

denizural: @pgierz, this is why I adopted the test first approach 😃

pgierz: OK then make this into two checks.  Once that it raises an exception. And a second check that ensures the user is correctly informed. 

What Miguel is trying to do here is guard against our most common user complaint. People hate looking at python trackbacks and instead want a beautiful message. The bottom part of this test will ensure that all part of those beautiful messages are included. 

So. Split it. Two tests.

pgierz: And: a system exit with a useful user message is certainly a state change

denizural: This is why `branch coverage` is important as much as unit testing. Pytest can measure it, as far as I can remember.

mandresm: To adds to this:
> What Miguel is trying to do here is guard against our most common user complaint. People hate looking at python trackbacks and instead want a beautiful message

The function to be tested here tries to get as much information as possible about wrong syntax, and reports all that info at once, so the users don't have to discover one error at a time, one `esm_runscripts --check` at a time. That's what makes this function more sensitive to errors: it needs to assemble a set of user-readable errors and then report the full thing as a user error. I find crucial that the users are reported the correct errors and for that I put together 4 unit tests, testing for each set of errors possible.

denizural (approved):

PR #790, opened 2022-08-15
Sprint/filedicts/unify functions

denizural:
- [x] adds some inline documentation for advanced users
- [x] unifies path checks
- [x] adds unit tests for new and some other older functions. Increased code coverage. Found bugs in the older code
- [x] multiple failure cases & type checks for function arguments 
- [x] cleans some of the older redundant code.
- [x] `_resolve_paths` --> `_resolve_abs_paths` : better method name
- [ ] TODO: WIP. Convert all bare exceptions to more specific ones

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -341,7 +341,8 @@ class SimulationFile(dict):
         if isinstance(path, str):
             path = pathlib.Path(path)
         elif not isinstance(path, (str, pathlib.Path)):
-            raise TypeError(f"{path} has an incompatible type {type(path)}")
+            datatype = type(path).__name__
+            raise TypeError(f"Path ``{path}`` has an incompatible datatype ``{datatype}``. str or pathlib.Path is expected")
 
         # NOTE: is_symlink() needs to come first because it is also a is_file()
         if not path.exists():
@@ -382,8 +383,12 @@ class SimulationFile(dict):
         # TODO: add typing
 
         # Types. Eg. file, dir, link, or None
-        source_path_type = self._path_type(source_path)
-        target_path_type = self._path_type(target_path)
+        try:
+            source_path_type = self._path_type(source_path)
+            target_path_type = self._path_type(target_path)
+        except TypeError as err:
+            print(err)
+            raise
 
         # Checks
         # ------
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -438,7 +438,7 @@ class SimulationFile(dict):
 
         # if source is a broken link. Ie. pointing to a non-existing file
         if source_path_type == FileTypes.BROKEN_LINK:
-            err_msg = f"Unable to create symbolic link: `{source_path}` points to a broken path: {source_path.resolve()}"
+            err_msg = f"Unable to create symbolic link: ``{source_path}`` points to a broken path: {source_path.resolve()}"
             raise FileNotFoundError(err_msg)
 
         return True
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -317,35 +317,6 @@ class SimulationFile(dict):
         for key, path in self.locations.items():
             self[f"absolute_path_in_{key}"] = path.joinpath(self[f"name_in_{key}"])
 
-    def convert_to_path(self, path: Union[str, pathlib.Path]) -> pathlib.Path:
-        """
-        Converts the ``path`` to pathlib.Path object since ESM-Tools uses pathlib instead of str. This is a low level function to be called before higher level file I/O functions
-
-        Parameters
-        ----------
-        path : str, pathlib.Path
-            path to be checked
-
-        Returns
-        -------
-        path : pathlib.Path
-            converted path
-
-        Raises
-        ------
-        TypeError
-            - When input is not string or pathlike object. Eg. dict, integer, ...
-        """
-        if isinstance(path, pathlib.Path):
-            return path
-        if isinstance(path, str):
-            return pathlib.Path(path)
-        if not isinstance(path, (str, pathlib.Path)):
-            datatype = type(path).__name__
-            raise TypeError(
-                f"Path ``{path}`` has an incompatible datatype ``{datatype}``. str or pathlib.Path is expected"
-            )
-
     def _path_type(self, path: pathlib.Path) -> Union[str, None]:
         """
         Checks if the given ``path`` exists. If it does returns it's type, if it
@@ -364,9 +335,6 @@ class SimulationFile(dict):
         """
 
         # TODO: check docstring again
-
-        # type check and casts
-        path = self.convert_to_path(path)
 
         # NOTE: is_symlink() needs to come first because it is also a is_file()
         if not path.exists():
@@ -405,10 +373,6 @@ class SimulationFile(dict):
 
         # TODO: return True
         # TODO: add typing
-
-        # type check and casts
-        source_path = self.convert_to_path(source_path)
-        target_path = self.convert_to_path(target_path)
 
         # Types. Eg. file, dir, link, or None
         try:
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -250,27 +250,9 @@ class SimulationFile(dict):
             )
             raise OSError(err_msg)
 
-        if not os.path.exists(source_path):
-            err_msg = f"Unable to create symbolic link: source file `{source_path}` does not exist"
-            raise FileNotFoundError(err_msg)
-
         if os.path.isdir(target_path):
             err_msg = f"Unable to create symbolic link: `{target_path}` is a directory"
             raise OSError(err_msg)
-
-        target_exists = os.path.exists(target_path) or os.path.islink(target_path)
-        if target_exists:
-            err_msg = (
-                f"Unable to create symbolic link: `{target_path}`. File already exists"
-            )
-            raise FileExistsError(err_msg)
-
-        target_parent = target_path.parent
-        if not target_parent.exists():
-            err_msg = (
-                f"Unable to create symbolic link: `{target_parent}` does not exist"
-            )
-            raise FileNotFoundError(err_msg)
 
         os.symlink(source_path, target_path)
 
@@ -335,6 +317,10 @@ class SimulationFile(dict):
         for key, path in self.locations.items():
             self[f"absolute_path_in_{key}"] = path.joinpath(self[f"name_in_{key}"])
 
+
+    # TODO: def to_path(): class method olabilir. Asagidaki type check leri buraya al
+
+
     def _path_type(self, path: pathlib.Path) -> Union[str, None]:
         """
         Checks if the given ``path`` exists. If it does returns it's type, if it
@@ -392,27 +378,34 @@ class SimulationFile(dict):
             - If the parent dir of the ``target_path`` does not exist
         """
 
-        # Types. Eg. file, dir, link, or not set
+        # TODO: return True
+        # TODO: add typing
+
+        # Types. Eg. file, dir, link, or None
         source_path_type = self._path_type(source_path)
         target_path_type = self._path_type(target_path)
-        target_path_parent_type = self._path_type(target_path.parent)
-            
 
         # Checks
         # ------
-        # Source exists
-        if not source_path_type:
-            raise Exception(f"Source file ``{source_path}`` does not exist!")
-        # Target exist
+        # Source does not exist
+        if source_path_type is None:
+            err_msg = f"Unable to perform file operation. Source ``{source_path}`` does not exist!"
+            raise FileNotFoundError(err_msg)
+
+        # Target already exists
+        target_exists = os.path.exists(target_path) or os.path.islink(target_path)
         if target_path_type:
-            # TODO: Change this behavior
-            raise Exception(f"File ``{target_path_type}`` already exists!")
-        # Target dir exists
-        if not target_path_parent_type:
+            err_msg = f"Unable to perform file operation. Target ``{target_path}`` already exists"
+            # TODO: ??? Change this behavior
+            raise FileExistsError(err_msg)
+
+        # Target parent directory does not exist
+        if not target_path.parent.exists():
             # TODO: we might consider creating it
-            raise Exception(
-                f"Target directory ``{target_path_parent_type}`` does not exist!"
-            )
+            err_msg = f"Unable to perform file operation. Parent directory of the target ``{target_path}`` does not exist"
+            raise FileNotFoundError(err_msg)
+
+        return True
 
 
 def copy_files(config):
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -191,7 +191,7 @@ class SimulationFile(dict):
         source_path = self[f"absolute_path_in_{source}"]
         target_path = self[f"absolute_path_in_{target}"]
 
-        # Checks
+        # general checks
         self._check_source_and_target(source_path, target_path)
 
         # Actual copy
@@ -240,7 +240,7 @@ class SimulationFile(dict):
         source_path = self[f"absolute_path_in_{source}"]
         target_path = self[f"absolute_path_in_{target}"]
 
-        # Checks
+        # general checks
         self._check_source_and_target(source_path, target_path)
 
         points_to_itself = source_path == target_path
@@ -276,7 +276,7 @@ class SimulationFile(dict):
         source_path = self[f"absolute_path_in_{source}"]
         target_path = self[f"absolute_path_in_{target}"]
 
-        # Checks
+        # general checks
         self._check_source_and_target(source_path, target_path)
 
         # Perform the movement:
@@ -318,7 +318,33 @@ class SimulationFile(dict):
             self[f"absolute_path_in_{key}"] = path.joinpath(self[f"name_in_{key}"])
 
 
-    # TODO: def to_path(): class method olabilir. Asagidaki type check leri buraya al
+
+    def convert_to_path(self, path: Union[str, pathlib.Path]) -> pathlib.Path: 
+        """
+        Converts the ``path`` to pathlib.Path object since ESM-Tools uses pathlib instead of str. This is a low level function to be called before higher level file I/O functions
+
+        Parameters
+        ----------
+        path : str, pathlib.Path
+            path to be checked
+
+        Returns
+        -------
+        path : pathlib.Path
+            converted path
+
+        Raises
+        ------
+        TypeError
+            - When input is not string or pathlike object. Eg. dict, integer, ...
+        """
+        if isinstance(path, pathlib.Path):
+            return path
+        if isinstance(path, str):
+            return pathlib.Path(path)
+        if not isinstance(path, (str, pathlib.Path)):
+            datatype = type(path).__name__
+            raise TypeError(f"Path ``{path}`` has an incompatible datatype ``{datatype}``. str or pathlib.Path is expected")
 
 
     def _path_type(self, path: pathlib.Path) -> Union[str, None]:
@@ -337,12 +363,11 @@ class SimulationFile(dict):
             If the path exists it returns its type as a string (``file``, ``dir``,
             ``link``). If it doesn't exist returns ``None``.
         """
+
+        # TODO: check docstring again
+
         # type check and casts
-        if isinstance(path, str):
-            path = pathlib.Path(path)
-        elif not isinstance(path, (str, pathlib.Path)):
-            datatype = type(path).__name__
-            raise TypeError(f"Path ``{path}`` has an incompatible datatype ``{datatype}``. str or pathlib.Path is expected")
+        path = self.convert_to_path(path)
 
         # NOTE: is_symlink() needs to come first because it is also a is_file()
         if not path.exists():
@@ -381,6 +406,10 @@ class SimulationFile(dict):
 
         # TODO: return True
         # TODO: add typing
+
+        # type check and casts
+        source_path = self.convert_to_path(source_path)
+        target_path = self.convert_to_path(target_path)
 
         # Types. Eg. file, dir, link, or None
         try:
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -110,6 +110,22 @@ class SimulationFile(dict):
             The full simulation configuration
         attrs_address : str
             The address of this specific file in the full config, separated by dots.
+
+        Note
+        ----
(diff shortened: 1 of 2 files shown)


src/esm_runscripts/filedicts.py
```diff
@@ -314,10 +317,39 @@ def _resolve_paths(self) -> None:
         for key, path in self.locations.items():
             self[f"absolute_path_in_{key}"] = path.joinpath(self[f"name_in_{key}"])
 
-    def _path_type(self, path: pathlib.Path) -> Union[str, bool]:
+    def convert_to_path(self, path: Union[str, pathlib.Path]) -> pathlib.Path:
```
pgierz: I really don't think we need this function. Have a look:
```
bash-5.1$ python
Python 3.9.5 (default, May 18 2021, 12:31:01)
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pathlib
>>> pathlib.Path(True)
  File "<stdin>", line 1, in <module>
  File "/Users/pgierz/opt/miniconda3/lib/python3.9/pathlib.py", line 1072, in __new__
    self = cls._from_parts(args, init=False)
  File "/Users/pgierz/opt/miniconda3/lib/python3.9/pathlib.py", line 697, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/Users/pgierz/opt/miniconda3/lib/python3.9/pathlib.py", line 681, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not bool
>>> pathlib.Path(123)
  File "<stdin>", line 1, in <module>
  File "/Users/pgierz/opt/miniconda3/lib/python3.9/pathlib.py", line 1072, in __new__
    self = cls._from_parts(args, init=False)
  File "/Users/pgierz/opt/miniconda3/lib/python3.9/pathlib.py", line 697, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/Users/pgierz/opt/miniconda3/lib/python3.9/pathlib.py", line 681, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not int
>>> pathlib.Path("a")
PosixPath('a')
>>> pathlib.Path(pathlib.Path("a"))
PosixPath('a')
>>>
```

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -159,6 +159,74 @@ def test_filedicts_basics(fs):
...
+    assert path_out == path_in 
+
+    # string goes in, Path goes out
+    path_out = simulation_file.convert_to_path(path_str)
+    assert path_out == path_in
+
+    # garbage in, exception out
+    path_str = 12345
+    with pytest.raises(TypeError):
+        path_out = simulation_file.convert_to_path(path_str)
```
pgierz: Unneeded, see above.

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -409,47 +477,22 @@ def test_ln_raises_exception_when_target_path_does_not_exist(simulation_file, fs
...
-
-    # Captures output (i.e. the user-friendly error)
-    with Capturing() as output:
-        with pytest.raises(SystemExit) as error:
-            sim_file = esm_runscripts.filedicts.SimulationFile(
-                config, "echam.files.jan_surf"
-            )
 
-    # error needs to occur as the path is not absolute
-    assert any(["ERROR: File Dictionaries" in line for line in output])
```
pgierz: Please do not remove this part. I think @mandresm has more plans for it.

src/esm_runscripts/filedicts.py
```diff
@@ -222,35 +240,20 @@ def ln(self, source: AnyStr, target: AnyStr) -> None:
...
 
+        # general checks
+        self._check_source_and_target(source_path, target_path)
+
         points_to_itself = source_path == target_path
         if points_to_itself:
             err_msg = (
                 f"Unable to create symbolic link: `{source_path}` is linking to itself"
             )
             raise OSError(err_msg)
```
pgierz: Not needed; `os.symlink` does this already for you.

src/esm_runscripts/filedicts.py
```diff
@@ -130,8 +147,9 @@ def __init__(self, full_config: dict, attrs_address: dict):
         if self["type"] not in ["restart", "outdata"]:
             self["name_in_work"] = self.get("name_in_work", self["name_in_computer"])
 
-        # Complete paths for all possible locations
-        self._resolve_paths()
+        # initialize the locations and complete paths for all possible locations
+        locations = dict.fromkeys(location_keys, None)
```
pgierz: Is this ever actually used?

src/esm_runscripts/filedicts.py
```diff
@@ -120,6 +120,23 @@ def __init__(self, full_config: dict, attrs_address: dict):
...
+        ]
+
+        # Explanation of the file keys:
+        # -----------------------------
+        # - LOCATION_KEY is one of the strings defined in LOCATION_KEY list
+        #
+        # - name_in<LOCATION_KEY> : file name (without path) in the LOCATION_KEY
+        #   - eg. name_in_work
+        # - absolute_path_in_<LOCATION_KEY> : absolute path in the LOCATION_KEY
+        #   - eg. absolute_path_in_run_tree
```
pgierz: I would directly put these sorts of things into the object doc-string instead of first as comments and then moving them later. A doc-string can be as long as you need. 

At least how I work, I normally have online documentation open on one side, and my program on the other. Since the online is also cross-linked in sphinx, I can look up other parts of the handbook on the fly, which I cannot do here.

src/esm_runscripts/filedicts.py
```diff
@@ -314,10 +317,39 @@ def _resolve_paths(self) -> None:
...
+        """
+        if isinstance(path, pathlib.Path):
+            return path
+        if isinstance(path, str):
+            return pathlib.Path(path)
+        if not isinstance(path, (str, pathlib.Path)):
+            datatype = type(path).__name__
+            raise TypeError(
+                f"Path ``{path}`` has an incompatible datatype ``{datatype}``. str or pathlib.Path is expected"
+            )
```
pgierz: If you want to put the actual input the user had given into the exception:
```suggestion
        try:
            return pathlib.Path(path)
        except TypeError:
            raise EsmToolsFileDictError(f"Unable to convert {path} from {type(path)} to a pathlib.Path")
```

src/esm_runscripts/filedicts.py
```diff
@@ -363,26 +403,42 @@ def _check_source_and_target(self, source_path, target_path):
...
+        source_path = self.convert_to_path(source_path)
+        target_path = self.convert_to_path(target_path)
+
+        # Types. Eg. file, dir, link, or None
+        try:
+            source_path_type = self._path_type(source_path)
+            target_path_type = self._path_type(target_path)
+        except TypeError as err:
+            print(err)
+            raise
```
pgierz: Why? Raise it directly from the function...

pgierz (changes requested):
I think there are still some things that could be cleaned up here :-) but it looks better than the first version and is more in line with what `mv` and `cp` are doing! Hooray!

tests/test_esm_runscripts/test_filedicts.py
denizural: I brought it back but still keep the MUT because this is better for unit testing.

src/esm_runscripts/filedicts.py
denizural: I removed the method `convert_to_path` and the test associated with that but now I am getting failing tests because we can still have non `str` or `Path` object and there are no checks covering that.

tests/test_esm_runscripts/test_filedicts.py
pgierz: My comment only referred to that line, not the entire method

src/esm_runscripts/filedicts.py
mandresm: agreed

src/esm_runscripts/filedicts.py
```diff
@@ -120,6 +120,23 @@ def __init__(self, full_config: dict, attrs_address: dict):
...
         self.path_in_computer = pathlib.Path(self["path_in_computer"])
 
+        # possible paths for files:
+        location_keys = [
+            "computer",  # pool/source directory (for input files)
+            "exp_tree",  # file in the category directory in experiment directory (eg. input, output, ...)
+            "run_tree",  # file in the experiment/run_<DATE>/<CATEGORY>/ directory
+            "work",  # file in the current work directory. Eg. experiment/run_<DATE>/work/
+        ]
+
```
mandresm: I find this redundant. I don't see the need of initializing this variable. And it is not used anywhere since `locations` down in the code are also not used anywhere.

src/esm_runscripts/filedicts.py
mandresm: I don't think this is needed, and it seems redundant to me. All we need is in `self._resolve_paths`

tests/test_esm_runscripts/test_filedicts.py
denizural: Removed it

src/esm_runscripts/filedicts.py
```diff
@@ -326,20 +329,25 @@ def _path_type(self, path: pathlib.Path) -> Union[str, bool]:
...
         -------
-        str or bool
+        str or None
             If the path exists it returns its type as a string (``file``, ``dir``,
-            ``link``). If it doesn't exist returns ``False``.
+            ``link``). If it doesn't exist returns ``None``.
         """
-        if path.is_file():
+
+        # TODO: check docstring again
```
mandresm: Please, resolve this before merge

src/esm_runscripts/filedicts.py
denizural: Forgot to put the `self.`

src/esm_runscripts/filedicts.py
denizural: Removed it

src/esm_runscripts/filedicts.py
denizural: Just to be more explicit about it. This will reraise the TypeError.

src/esm_runscripts/filedicts.py
```diff
@@ -363,26 +371,38 @@ def _check_source_and_target(self, source_path, target_path):
...
-        if not source_path_type:
-            raise Exception(f"Source file ``{source_path}`` does not exist!")
-        # Target exist
+        # Source does not exist
+        if source_path_type is None:
+            err_msg = f"Unable to perform file operation. Source ``{source_path}`` does not exist!"
+            raise FileNotFoundError(err_msg)
+
+        # Target already exists
+        target_exists = os.path.exists(target_path) or os.path.islink(target_path)
```
mandresm: This is not used. Please, remove it

src/esm_runscripts/filedicts.py
```diff
@@ -363,26 +371,38 @@ def _check_source_and_target(self, source_path, target_path):
...
         if target_path_type:
-            # TODO: Change this behavior
-            raise Exception(f"File ``{target_path_type}`` already exists!")
-        # Target dir exists
-        if not target_path_parent_type:
+            err_msg = f"Unable to perform file operation. Target ``{target_path}`` already exists"
+            # TODO: ??? Change this behavior
```
mandresm: Remove this TODO when solved

src/esm_runscripts/filedicts.py
```diff
@@ -363,26 +371,38 @@ def _check_source_and_target(self, source_path, target_path):
...
+            raise FileExistsError(err_msg)
+
+        # Target parent directory does not exist
+        if not target_path.parent.exists():
             # TODO: we might consider creating it
```
mandresm: I wouldn't do that, it has the potential to create target directories in crazy places if the user makes a mistake on typing the paths. I would only do that in the case of `name_in_{location}` containing a folder. But that would go somewhere else.

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -159,6 +159,60 @@ def test_filedicts_basics(fs):
...
+    output = simulation_file._path_type(mylink)
+    assert output == "link"
+
+    # check for file
+    output = simulation_file._path_type(simulation_file["absolute_path_in_computer"])
+    assert output == "file"
+
+    # check for directory
+    output = simulation_file._path_type(simulation_file.locations["work"])
+    assert output == "dir"
```
mandresm: Shouldn't we split all of this in different unit tests?

mandresm: test descriptions are missing

src/esm_runscripts/filedicts.py
denizural: Moved them into the docstring with some real examples

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -159,6 +159,60 @@ def test_filedicts_basics(fs):
...
+    source_path = Path("/this/does/not/exist")
+    target_path = simulation_file.locations["work"]
+    with pytest.raises(FileNotFoundError):
+        simulation_file._check_source_and_target(source_path, target_path)
+
+    # target does not exist
+    source_path = simulation_file.locations["computer"] 
+    target_path = Path("/this/does/not/exist")
+    with pytest.raises(FileNotFoundError):
+        simulation_file._check_source_and_target(source_path, target_path)
```
mandresm: I think this can also be devided in different tests. One can turn the comments into test descriptions once the tests are separated.

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -409,47 +463,28 @@ def test_ln_raises_exception_when_target_path_does_not_exist(simulation_file, fs
...
-            jan_surf:
-                type: input
-                name_in_computer: T63CORE2_jan_surf.nc
-                name_in_work: unit.24
-                path_in_computer: pool/ECHAM/T63
-        experiment_input_dir: /work/ollie/pgierz/some_exp/input/echam
-        thisrun_input_dir: /work/ollie/pgierz/some_exp/run_20010101-20010101/input/echam
-    """
-    config = yaml.safe_load(dummy_config)
+    simulation_file.path_in_computer = Path("foo/bar")
```
mandresm, line 645: I'm okay with having a dictionary defined somewhere else in the tests to be used as default, but I don't like that the object is instanced outside of the test. IMO the `SimulationFile` object should be instanced in the test, after we have modified the default `dict` manually on the test too. This method is closer to our real case scenario: 1) user defines dict, 2) ESM-Tools handles that dict. Right now you are doing: 1) test defines dict, 2) test instances dict, 3) test modifies that instance using the object methods, 4) test handles the operation of the test

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -409,47 +463,28 @@ def test_ln_raises_exception_when_target_path_does_not_exist(simulation_file, fs
...
 
     # Captures output (i.e. the user-friendly error)
     with Capturing() as output:
         with pytest.raises(SystemExit) as error:
-            sim_file = esm_runscripts.filedicts.SimulationFile(
-                config, "echam.files.jan_surf"
-            )
+            simulation_file._check_path_in_computer_is_abs()
```
mandresm, line 650: same as above

mandresm (changes requested):

src/esm_runscripts/filedicts.py
denizural: Comment is not from me. This is basically a guard against creating that target whose parent does not exist.

mandresm: True, this comment is mine XD sorry

tests/test_esm_runscripts/test_filedicts.py
denizural: Good idea. I wrote the test when I saw that the function was not covered. It is a good idea to refactor the tests and split them when they get large.

tests/test_esm_runscripts/test_filedicts.py
denizural: Good idea, I will do it.

src/esm_runscripts/filedicts.py
denizural: This comment is outdated. I already moved it to docstrings.

src/esm_runscripts/filedicts.py
denizural: 🤣 yes, that happens all the time. I think the solution is placing the name initial before the comment 😄

src/esm_runscripts/filedicts.py
denizural: Sorry, that was a typo. I am fixing it.

src/esm_runscripts/filedicts.py
denizural: `self.locations` is now initialized with `None` values

src/esm_runscripts/filedicts.py
denizural: solved

tests/test_esm_runscripts/test_filedicts.py
denizural: Newest version has separate test cases for the `_path_type_` method.

tests/test_esm_runscripts/test_filedicts.py
denizural: Already implemented and pushed.

src/esm_runscripts/filedicts.py
```diff
@@ -120,6 +147,9 @@ def __init__(self, full_config: dict, attrs_address: dict):
         self.component = component = attrs_address.split(".")[0]
         self.path_in_computer = pathlib.Path(self["path_in_computer"])
 
+        # possible paths for files:
+        location_keys = ["computer", "exp_tree", "run_tree", "work"]
```
mandresm: Would it make sense to move this below the `# Initialize the locations and complete paths`? In this way the whole initialization is together and clearer.

tests/test_esm_runscripts/test_filedicts.py
denizural: I see your point, but the **unit test** aims to test a **specific** function/method behaviour with different edge cases. Testing the behaviour of the object is also important (eg. functional test). I propose to have the following:
- have very specific unit tests for every single method to increase the coverage.
- have buggy YAML files with descriptive names. Eg. `config_no_absolute_paths.yaml` and write **functional tests** that test the higher level behaviour of the object when these YAML files are parsed.

What do you guys think @mandresm @pgierz

src/esm_runscripts/filedicts.py
```diff
@@ -173,12 +204,13 @@ def cp(self, source: str, target: str) -> None:
         source_path = self[f"absolute_path_in_{source}"]
         target_path = self[f"absolute_path_in_{target}"]
 
-        # Checks
+        # general checks
+        # TODO (deniz): need to add higher level exception handler (eg. user_error)
```
mandresm, line 225:
```suggestion
```
No please, we agreed we will only use user_error for "extended-yaml syntax" problems

src/esm_runscripts/filedicts.py
```diff
@@ -193,7 +225,7 @@ def cp(self, source: str, target: str) -> None:
 
     @_allowed_to_be_missing
     def ln(self, source: AnyStr, target: AnyStr) -> None:
-        """creates symbolic links from the path retrieved by ``source`` to the one by ``target``
+        """creates symbolic links from the path retrieved by ``source`` to the one by ``target``. ``source`` and ``target`` keys will be transformed into their corresponding absolute paths and ``target`` will be created.
```
mandresm: I don't understand the extra documentation here. Nothing gets transformed in this function right?

src/esm_runscripts/filedicts.py
```diff
@@ -222,35 +254,14 @@ def ln(self, source: AnyStr, target: AnyStr) -> None:
...
-            err_msg = (
-                f"Unable to create symbolic link: `{source_path}` is linking to itself"
-            )
-            raise OSError(err_msg)
-
-        if not os.path.exists(source_path):
-            err_msg = f"Unable to create symbolic link: source file `{source_path}` does not exist"
-            raise FileNotFoundError(err_msg)
+        # general checks
+        # TODO (deniz): need to add higher level exception handler (eg. user_error)
```
mandresm, line 275:
```suggestion
```
As above, nothing to be done here

src/esm_runscripts/filedicts.py
```diff
@@ -273,7 +284,8 @@ def mv(self, source: str, target: str) -> None:
         source_path = self[f"absolute_path_in_{source}"]
         target_path = self[f"absolute_path_in_{target}"]
 
-        # Checks
+        # general checks
+        # TODO (deniz): need to add higher level exception handler (eg. user_error)
```
mandresm, line 305:
```suggestion
```
Same

src/esm_runscripts/filedicts.py
```diff
@@ -362,27 +407,41 @@ def _check_source_and_target(self, source_path, target_path):
...
-        source_path_type = self._path_type(source_path)
-        target_path_type = self._path_type(target_path)
-        target_path_parent_type = self._path_type(target_path.parent)
+        # Types. Eg. file, dir, link, or None
+        try:
+            source_path_type = self._path_type(source_path)
+            target_path_type = self._path_type(target_path)
+        except TypeError as err:
+            print(err)
+            raise
```
mandresm: What is the purpose of this change?

src/esm_runscripts/filedicts.py
```diff
@@ -362,27 +407,41 @@ def _check_source_and_target(self, source_path, target_path):
...
+
+        # Target parent directory does not exist
+        if not target_path.parent.exists():
+            # TODO: we might consider creating it (Miguel)
+            err_msg = f"Unable to perform file operation. Parent directory of the target ``{target_path}`` does not exist"
+            raise FileNotFoundError(err_msg)
+
+        # if source is a broken link. Ie. pointing to a non-existing file
+        if source_path_type == FileTypes.BROKEN_LINK:
+            err_msg = f"Unable to create symbolic link: `{source_path}` points to a broken path: {source_path.resolve()}"
```
```suggestion
            err_msg = f"Unable to create symbolic link: ``{source_path}`` points to a broken path: {source_path.resolve()}"
```

mandresm (approved):
After the minor changes I suggested I think this is ready to be merged

src/esm_runscripts/filedicts.py
denizural: I think I had to write a better comment there. I meant that we could wrap this in a `try-except` block and exception would call the `user_error`

src/esm_runscripts/filedicts.py
denizural: It was re-raising the exception. Just removed it now. Tests are still passing

PR #791, opened 2022-08-17
sprint/filedicts/default names and allow missing path_in_computer

mandresm:
Generalization needed for reduced syntax in the yamls. Output files have different requirements in terms of syntax than input files, and this is already contemplated on the syntax checker, however, the previous version of the `__init__` did not allow for this syntax flexibility. By setting default names and allowing `path_in_computer` to be missing in output files, this problem is solved.

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -440,7 +440,7 @@ class SimulationFile(dict):
             user_error("File Dictionaries", f"{error_text}\n{missing_vars}")
 
     def _check_path_in_computer_is_abs(self):
-        if self.path_in_computer and not self.path_in_computer.is_absolute():
+        if self.path_in_computer is not None and not self.path_in_computer.is_absolute():
             user_error(
                 "File Dictionaries",
                 "The path defined for "
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -129,14 +129,7 @@ class SimulationFile(dict):
         self._check_file_syntax()
 
         # Complete tree names if not defined by the user
-        if self["type"] in self.input_file_types:
-            default_name = self["name_in_computer"]
-        elif self["type"] in self.output_file_types:
-            default_name = self["name_in_work"]
-        self["name_in_computer"] = self.get("name_in_computer", default_name)
-        self["name_in_run_tree"] = self.get("name_in_run_tree", default_name)
-        self["name_in_exp_tree"] = self.get("name_in_exp_tree", default_name)
-        self["name_in_work"] = self.get("name_in_work", default_name)
+        self._complete_file_names()
 
         # Complete paths for all possible locations
         self._resolve_paths()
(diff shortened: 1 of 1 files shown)


src/esm_runscripts/filedicts.py
```diff
@@ -122,20 +122,21 @@ def __init__(self, full_config: dict, attrs_address: dict):
...
-        self["name_in_run_tree"] = self.get(
-            "name_in_run_tree", self["name_in_computer"]
-        )
-        self["name_in_exp_tree"] = self.get(
-            "name_in_exp_tree", self["name_in_computer"]
-        )
-        if self["type"] not in ["restart", "outdata"]:
-            self["name_in_work"] = self.get("name_in_work", self["name_in_computer"])
+        if self["type"] in self.input_file_types:
+            default_name = self["name_in_computer"]
```
pgierz: Not `"name_in_exp_tree"`...?

src/esm_runscripts/filedicts.py
```diff
@@ -122,20 +122,21 @@ def __init__(self, full_config: dict, attrs_address: dict):
...
+        elif self["type"] in self.output_file_types:
+            default_name = self["name_in_work"]
+        self["name_in_computer"] = self.get("name_in_computer", default_name)
+        self["name_in_run_tree"] = self.get("name_in_run_tree", default_name)
+        self["name_in_exp_tree"] = self.get("name_in_exp_tree", default_name)
+        self["name_in_work"] = self.get("name_in_work", default_name)
```
pgierz: An idea: Make this into a mini function; all the logic kinda relates to the same things.

src/esm_runscripts/filedicts.py
```diff
@@ -309,7 +310,7 @@ def _resolve_paths(self) -> None:
         """
         self.locations = {
             "work": pathlib.Path(self._config["general"]["thisrun_work_dir"]),
-            "computer": pathlib.Path(self["path_in_computer"]),
+            "computer": self.path_in_computer, # Already Path type from _init_
```
pgierz, line 320: I would leave it like it is. Conversion of `pathlib.Path` to `Path` again is a zero-operation, and then it is consistent with the other ones, and you save yourself a comment 😉

src/esm_runscripts/filedicts.py
```diff
@@ -436,7 +440,7 @@ def _check_file_syntax(self):
             user_error("File Dictionaries", f"{error_text}\n{missing_vars}")
 
     def _check_path_in_computer_is_abs(self):
-        if not self.path_in_computer.is_absolute():
+        if self.path_in_computer and not self.path_in_computer.is_absolute():
```
pgierz: Mini style thing:
```suggestion
        if self.path_in_computer is not None and not self.path_in_computer.is_absolute():
```

pgierz (approved):
Looks good. I have one little recommendation, but that's up to you.

src/esm_runscripts/filedicts.py
mandresm: you read my mind :)

src/esm_runscripts/filedicts.py
mandresm: I think the users think more of things coming from the pool to the experiment/work, not from the exp_dir to the work directory.

src/esm_runscripts/filedicts.py
mandresm: It breaks if `self["path_in_computer"]` is `None` and that's the problem.

pgierz: That's annoying.

mandresm: yeah, I wish there was a PathNone or something like that...

src/esm_runscripts/filedicts.py
mandresm: Done!

src/esm_runscripts/filedicts.py
denizural: I have a good solution for that. Check for my `FileTypes` enumeration

denizural: Looks good so far. I can refactor your `path` `None` things with my `FileTypes` later on.

denizural (approved):
👏

PR #792, opened 2022-08-17
Sprint/filedicts/globbing

mandresm:
- [x] unit tests
- [x] Functionality
    - [x] wildcard patterns
    - [x] globbing loop decorator
- [x] cleanup, chunk globbing into smaller functions
- [x] docstrings

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -73,6 +73,24 @@ def _allowed_to_be_missing(method):
     return inner_method
 
 
+def globbing(method):
+    @functools.wraps(method)
+    def inner_method(self, source, target, *args, **kwargs):
+        source_name = self[f"name_in_{source}"]
+        target_name = self[f"name_in_{target}"]
+        if "*" in source_name or "*" in target_name:
+            source_pattern, target_pattern = self.wild_card_check(
+                source_name, target_name
+            )
+            print(source_name, target_name)
+            
+        else:
+            return method(self, source, target, *args, **kwargs)
+
+    return inner_method
+
+
+
 class SimulationFile(dict):
     """
     Describes a file used within a ESM Simulation.
@@ -166,6 +184,7 @@ class SimulationFile(dict):
         """
         return self.get("allowed_to_be_missing", False)
 
+    @globbing
     @_allowed_to_be_missing
     def cp(self, source: str, target: str) -> None:
         """
@@ -360,8 +379,22 @@ class SimulationFile(dict):
             raise Exception(f"Cannot identify the path's type of {path}")
 
     @classmethod
-    def wild_card_renaming(self, source_path, target_path) -> Tuple[list, list]:
-        return [], []
+    def wild_card_check(self, source_name, target_name) -> Tuple[list, list]:
+        wild_card_source = source_name.split("*")
+        wild_card_target = target_name.split("*")
+        # Check for syntax mistakes
+        if len(wild_card_target) != len(wild_card_source):
+            user_error(
+                "Wild card",
+                (
+                    "The wild card pattern of the source "
+                    + f"{wild_card_source} does not match with the "
+                    + f"target {wild_card_target}. Make sure the "
+                    + f"that the number of * are the same in both "
+                    + f"sources and targets."
+                ),
+            )
+        return wild_card_source, wild_card_target
 
     def _check_file_syntax(self):
         """
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -432,8 +432,8 @@ class SimulationFile(dict):
         else:
             raise Exception(f"Cannot identify the path's type of {path}")
 
-    @classmethod
-    def wild_card_check(self, source_pattern: list, target_pattern: list) -> True:
+    @staticmethod 
+    def wild_card_check(source_pattern: list, target_pattern: list) -> True:
         """
         Checks for syntax mistakes. If any were found, it notifies the user about these
         errors in the syntax using ``esm_parser.error``.
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -77,6 +77,7 @@ def _allowed_to_be_missing(method):
 def globbing(method):
     @functools.wraps(method)
     def inner_method(self, source, target, *args, **kwargs):
+        method_name = method.__name__
         source_name = self[f"name_in_{source}"]
         target_name = self[f"name_in_{target}"]
         if "*" in source_name or "*" in target_name:
@@ -118,7 +119,8 @@ def globbing(method):
                 glob_dict[f"name_in_{target}"] = glob_target_name
                 glob_file = SimulationFile(glob_config, self.attrs_address)
                 # Use method
-                method(glob_file, source, target, *args, **kwargs)
+                this_method = getattr(glob_file, method_name)
+                this_method(source, target, *args, **kwargs)
         else:
             return method(self, source, target, *args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


src/esm_runscripts/filedicts.py
```diff
@@ -365,8 +373,8 @@ def _check_file_syntax(self):
         missing_vars = ""
         types_text = ", ".join(self.all_model_filetypes)
         this_filedict = copy.deepcopy(self._original_filedict)
-        input_file_types = ["config", "forcing", "input"]
-        output_file_types = [
+        self.input_file_types = input_file_types = ["config", "forcing", "input"]
```
denizural: why do we have `self.input_file_types` and `input_file_types`?

mandresm: I'll solve that, there is not a good reason, I was just lazy to change all `input_file_types` in that method by `self.input_file_types`

mandresm: I was checking that method and there seems to be a good reason to do that: improve readability, it's used multiple times in the method in already long lines. Is this a valid reason for having both defined here?

denizural (approved):
👏

src/esm_runscripts/filedicts.py
```diff
@@ -73,6 +74,78 @@ def inner_method(self, *args, **kwargs):
...
+            # Check wild cards syntax
+            self.wild_card_check(source_pattern, target_pattern)
+
+            # Obtain source files
+            glob_source_paths = self.find_globbing_files(source)
+
+            # Extract globbing source names
+            glob_source_names = []
+            for glob_source_path in glob_source_paths:
+                glob_source_names.append(pathlib.Path(glob_source_path).name)
```
pgierz: Could be simpler:
```suggestion
            # Extract globbing source names
            glob_source_names = [glob_source_path.name for glob_source_path in glob_source_paths]
```

src/esm_runscripts/filedicts.py
```diff
@@ -73,6 +74,78 @@ def inner_method(self, *args, **kwargs):
...
+                glob_source_names, glob_target_names
+            ):
+                # Create a new simulation file object for this specific glob file
+                glob_config = copy.deepcopy(self._config)
+                glob_dict = dpath.util.get(
+                    glob_config, self.attrs_address, separator=".", default={}
+                )
+                glob_dict[f"name_in_{source}"] = glob_source_name
+                glob_dict[f"name_in_{target}"] = glob_target_name
+                glob_file = SimulationFile(glob_config, self.attrs_address)
```
pgierz, line 136: This part will bite me when I get the class constructor finished, but that's for me to fix, not you.

src/esm_runscripts/filedicts.py
```diff
@@ -73,6 +74,78 @@ def inner_method(self, *args, **kwargs):
...
+                # Use method
+                this_method = getattr(glob_file, method_name)
+                this_method(source, target, *args, **kwargs)
+        else:
+            return method(self, source, target, *args, **kwargs)
```
pgierz, line 141: You could save yourself a bit of code here (double check the indents, GitHub's online review suggest thingy isn't good about that):
```suggestion
                # Use method
                method = getattr(glob_file, method_name)
        return method(self, source, target, *args, **kwargs)
```

src/esm_runscripts/filedicts.py
```diff
@@ -359,7 +436,67 @@ def _path_type(self, path: pathlib.Path) -> Union[str, bool]:
         else:
             raise Exception(f"Cannot identify the path's type of {path}")
 
-    def _check_file_syntax(self):
+    @classmethod
+    def wild_card_check(self, source_pattern: list, target_pattern: list) -> True:
```
pgierz: Class methods should use `cls` instead of `self` as the first argument. The actual argument name does not matter, so it is just for clarity. Here is an example:

For regular methods, you get an instance of the object, e.g.:
```python
class MyThing:
    def do_stuff(self):
        print(f"Hooray, I am {self}")

    @classmethod
    def do_class_stuff(cls):
        print(f"Woohoo, I am {cls}")

a = MyThing()
a.do_stuff()
MyThing.do_class_stuff()
a.do_class_stuff()
```

Example output:
```
Hooray, I am <__main__.MyThing object at 0x7fa636dd7fa0>
Woohoo, I am <class '__main__.MyThing'>
Woohoo, I am <class '__main__.MyThing'>
```

And in fact, given that you never use `self` at all, this can just be a `staticmethod`

```suggestion
    @staticmethod 
    def wild_card_check(source_pattern: list, target_pattern: list) -> True:
```

src/esm_runscripts/filedicts.py
```diff
@@ -359,7 +436,67 @@ def _path_type(self, path: pathlib.Path) -> Union[str, bool]:
...
+                    + f"sources and targets."
+                ),
+            )
+
+        return True
+
+    def find_globbing_files(self, location : str) -> list:
+        """
+        Lists the files matching the globbing path of the given ``location``, and
+        notifies the user if none were found, via ``esm_parser.error``.
```
pgierz:
```suggestion
        notifies the user if none were found, via ``esm_parser.user_error``.
```

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -633,3 +633,207 @@ def test_resolve_paths_old_config():
...
+    target_pattern = target_name.split("*")
+
+    assert(esm_runscripts.filedicts.SimulationFile.wild_card_check(
+        source_pattern, target_pattern
+    ))
+
+def test_wild_card_check_fails():
+    """
+    Tests that, when given an incorrect wildcard pattern, a ``user_error`` is reported
+    """
```
pgierz:
```suggestion
    """
    Tests that, when given an incorrect wildcard pattern (more wildcards in source than in target), a ``user_error`` is reported
    """
```

tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -633,3 +633,207 @@ def test_resolve_paths_old_config():
...
+                type: outdata
+        experiment_outdata_dir: /work/ollie/pgierz/some_exp/input/oifs
+        thisrun_outdata_dir: /work/ollie/pgierz/some_exp/run_20010101-20010101/input/oifs
+    """
+    config = yaml.safe_load(dummy_config)
+    files = [
+        "input_11_DATE_12.nc",
+        "input_21_DATE_22.nc",
+        "input_31_DATE_32.nc",
+    ]
```
pgierz:
```suggestion
    config = yaml.safe_load(dummy_config)
    # These files are incorrect since they do not have input_expid and instead have input_<NUMBER>
    files = [
        "input_11_DATE_12.nc",
        "input_21_DATE_22.nc",
        "input_31_DATE_32.nc",
    ]
```

pgierz (commented):
Only some small things, looks good 👍🏻

src/esm_runscripts/filedicts.py
mandresm: The problem here was that you need `pathlib.Path` because `glob` return a list of strings, not Paths.

pgierz: Then instead:
```suggestion
            # Extract globbing source names
            glob_source_names = [pathlib.Path(glob_source_path).name for glob_source_path in glob_source_paths]
```

src/esm_runscripts/filedicts.py
mandresm: I still don't understand what's the purpose of the class constructor :S

src/esm_runscripts/filedicts.py
mandresm: Good point!

src/esm_runscripts/filedicts.py
mandresm: I see! This clarifies a lot, thanks! I'll copy the link to this comment to my technical notes for the time I forget ^^

PR #796, opened 2022-08-19
create subfolders in target names

mandresm:
- [x] unit tests
- [x] Functionality
- [x] docstrings

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -473,7 +473,7 @@ class SimulationFile(dict):
             # probably, this will not happen
             raise TypeError(f"{path} can not be identified")
 
-    @staticmethod 
+    @staticmethod
     def wild_card_check(source_pattern: list, target_pattern: list) -> True:
         """
         Checks for syntax mistakes. If any were found, it notifies the user about these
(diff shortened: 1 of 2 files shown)


tests/test_esm_runscripts/test_filedicts.py
```diff
@@ -929,3 +929,36 @@ def test_globbing_ln(fs):
...
+                name_in_work: ICMGG_input_expid+in_work
+                name_in_exp_tree: out/date/folder/ICMGG_input_expid
+                type: outdata
+        experiment_outdata_dir: /work/ollie/pgierz/some_exp/input/oifs
+        thisrun_outdata_dir: /work/ollie/pgierz/some_exp/run_20010101-20010101/input/oifs
+    """
+
+    config = yaml.safe_load(dummy_config)
+
+    fs.create_file(Path(config["general"]["thisrun_work_dir"]).joinpath(f"ICMGG_input_expid+in_work"))
```
denizural, line 1093: One handy thing about `pathlib` is that `/` operator is overloaded for concatenating paths. Eg.
```python
new_path = some_path / another_path
```

src/esm_runscripts/filedicts.py
```diff
@@ -527,6 +533,19 @@ def find_globbing_files(self, location : str) -> list:
...
+            if not parent_path.exists():
+                location = self.locations[name_type]
+                if location.exists():
+                    os.makedirs(parent_path)
+                else:
+                    raise FileNotFoundError(
+                        f"Unable to perform file operation. Path for ``{name_type}`` "
+                        f"({location}) does not exist!"
+                    )
+
```
pgierz: Only a comment: we should stay consistent with the parameter names. Here we have `name_type`, on the outside we have `target`. It would make it easier to read. Not a hard-must-have, but just an idea.

pgierz (approved):

denizural (approved):

src/esm_runscripts/filedicts.py
mandresm: I changed the name here, because we might in the future use this function for the source? I thought it was just more general.

pgierz: That makes sense 👍

PR #797, opened 2022-08-19
adds logger. adds __str__

denizural:


src/esm_runscripts/filedicts.py
```diff
@@ -115,7 +120,7 @@ class SimulationFile(dict):
         >>> sim_file.cp_to_exp_tree()  # doctest: +SKIP
     """
 
-    def __init__(self, full_config: dict, attrs_address: dict):
+    def __init__(self, full_config: dict, attrs_address: dict, separator="."):
```
pgierz: Don't make this an init parameter please. Will we ever in any case use a different one?
```suggestion
    def __init__(self, full_config: dict, attrs_address: dict):
```

src/esm_runscripts/filedicts.py
```diff
@@ -149,8 +154,10 @@ def __init__(self, full_config: dict, attrs_address: dict):
         super().__init__(attrs_dict)
         self._original_filedict = copy.deepcopy(attrs_dict)
         self._config = full_config
-        self.name = attrs_address.split(".")[-1]
-        self.component = component = attrs_address.split(".")[0]
+        self._separator = separator
+        self._attrs_address = attrs_address
+        self.name = attrs_address.split(separator)[-1]
+        self.component = component = attrs_address.split(separator)[0]
```
pgierz:
```suggestion
        self.name = attrs_address.split(".")[-1]
        self.component = component = attrs_address.split(".")[0]
```

pgierz (changes requested):

mandresm (approved):

PR #798, opened 2022-08-19
fixes most of the mypy warnings / errors

denizural:


src/esm_runscripts/filedicts.py
```diff
@@ -282,7 +282,7 @@ def __init__(self, full_config: dict, attrs_address: str):
         # possible paths for files:
         location_keys = ["computer", "exp_tree", "run_tree", "work"]
         # initialize the locations and complete paths for all possible locations
-        self.locations = dict.fromkeys(location_keys, None)
+        self.locations = dict.fromkeys(location_keys, pathlib.Path("dev/null"))
```
```suggestion
        self.locations = dict.fromkeys(location_keys, None)
```
`/dev/null` exists and it might be dangerous if we are checking `if self.locations` somewhere else in the code.

mandresm (approved):
Nice, but please fix the comment :)

src/esm_runscripts/filedicts.py
denizural: Good point. I had to stop mypy from complaining. Maybe it is ok to leave some of the type warnings there

PR #800, opened 2022-08-19
cleanup: private names/public names of methods

mandresm:
I find almost all methods are correctly defined in terms of public/private, except my newest one.


denizural (approved):

pgierz:
What about the wildcard one?

mandresm:
It's used outside, in the globbing.

PR #801, opened 2022-08-19
adds docstrings to undocumented methods

pgierz:
Docstrings everywhere

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -1,5 +1,5 @@
 """
-The file-dictionary implementation
+This module contains the description of a SimulationFile.
 
 Developer Notes
 ---------------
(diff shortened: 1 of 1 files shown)


src/esm_runscripts/filedicts.py
```diff
@@ -287,7 +323,7 @@ def __init__(self, full_config: dict, attrs_address: str):
         # possible paths for files:
         location_keys = ["computer", "exp_tree", "run_tree", "work"]
         # initialize the locations and complete paths for all possible locations
-        self.locations = dict.fromkeys(location_keys, None)
+        self.locations = dict.fromkeys(location_keys, pathlib.Path("/dev/null"))
```
```suggestion
        self.locations = dict.fromkeys(location_keys, None)
```
That file exists, and if we ever check `if self.locations` it will return True. I think it's dangerous

denizural (approved):

mandresm (approved):

src/esm_runscripts/filedicts.py
pgierz: I still don't understand why we initialize `self.locaitons` here at all, to be honest. I think I will just remove it entirely.

PR #804, opened 2022-08-23
Feat/dual hemisphere levante

ackerlar:
Some changes for the pism awiesm coupling on levante. Mainly in /esm_tools/runscripts/awicm-pism/ so not affecting the actual esm_tools besides the corresponding changes in echam, jsbach, awiesm and fesom config files


configs/components/echam/echam.yaml
```diff
@@ -628,6 +647,7 @@ streams:
         - sp6h
         - glim
         - spim
+        - ism
```
pgierz, line 650: I would rather have this in an `add_streams` block, since not every model variant will produce this and then we will get warnings about missing files. @mandresm and I are considering how to rework the stream business from the ground up at some point.

For right now, you could do some thing like:
```yaml
choose_coupled_ice:
    True:
        add_streams:
            - ism
```

configs/components/echam/echam.yaml
```diff
@@ -746,6 +766,9 @@ namelist_changes:
                 parctl:
                         nproca: "${nproca}"
                         nprocb: "${nprocb}"
+                submodelctl:
+                        lgfw: "${lgfw}"
+                        lupdate_orog: "${lupdate_orography}"
```
pgierz, line 772: Good, but make sure this is only done for the ECHAM version that supports these features, otherwise you will get a FORTRAN error with an unrecognised namelist entry.

configs/components/echam/echam.yaml
```diff
@@ -791,7 +814,23 @@ grids:
...
+                        gfw_atmo: "gfw_atmo.nc"
+
+choose_lupdate_orography:
+        True:
+                add_input_sources:
+                        target_oro: "${general.experiment_couple_dir}/latest_target_orography.nc"
+                add_input_files:
+                        target_oro: target_oro
+                add_input_in_work:
+                        target_oro: "target_orography_echam6.nc"
```
pgierz, line 833: 👍🏻

configs/components/fesom/fesom-2.1.yaml
```diff
@@ -200,7 +200,8 @@ restart_out_sources:
 outdata_sources:
         fesom: "*.fesom.*.nc"
 outdata_targets:
-        fesom: "*.fesom.*${start_date!smonth}.${start_date!sday}.nc"
+        #fesom: "*.fesom.*${start_date!smonth}.${start_date!sday}.nc"
+        fesom: "*.fesom.*.nc"
```
pgierz, line 204: @mandresm can you double check this? I am not sure if this is what we want...

configs/components/jsbach/jsbach.datasets.yaml
```diff
@@ -2,6 +2,7 @@ input_sources:
         lctlib: "${echam.model_dir}/lctlib_nlct21.def"
         "jsbach_1850": "${adj_input_dir}/jsbach_${echam.resolution}${echam.ocean_resolution}_11tiles_5layers_1850${jsbachnc_file_ending}.nc"
         "jsbach_1976": "${adj_input_dir}/jsbach_${echam.resolution}${echam.ocean_resolution}_11tiles_5layers_1976${jsbachnc_file_ending}.nc"
+        "jsbach_ismc": "${general.experiment_couple_dir}/latest_jsbach_init_file.nc"
```
pgierz, line 5: Can you put a comment here what ismc means? I suppose ice sheet model coupled, but I will probably look at this in a year or two and completely forget what it is.

```suggestion
        # Input for JSBACH with Ice Sheet Model Coupling Enabled
        "jsbach_ismc": "${general.experiment_couple_dir}/latest_jsbach_init_file.nc"
```

configs/components/jsbach/jsbach.yaml
```diff
@@ -31,7 +31,7 @@ config_sources:
 
 # normally, except awicm on ollie :-(((
 no_dynveg_file_ending: "_no-dynveg"
-dynveg_file_ending: "_dynveg"
+dynveg_file_ending: "" #_dynveg"
```
pgierz, line 34: Careful here, I think this is machine dependent because the pools might not be the same. If they are not, I can take care of that.

runscripts/awicm-pism/coupling/_stdout_stderr
```diff
@@ -0,0 +1,3 @@
+NAG Fortran Compiler Release 7.1(Hanzomon) Build 7101
+MOD_PRE.f90:
+Error: Cannot open file "MOD_PRE.f90" - No such file or directory
```
pgierz, line 3: I don't think we need this file checked in, do we?

runscripts/awicm-pism/coupling/coupling_awiesm2pism.functions
```diff
@@ -4,22 +4,14 @@ awiesm2pism() {
...
     . ${FUNCTION_PATH}/pism_helpers.functions
     . ${FUNCTION_PATH}/general_lists.functions
+
+    if [ -f ${COUPLE_DIR}/latest_atmo_forcing_file.nc ]; then
+        rm ${COUPLE_DIR}/latest_atmo_forcing_file.nc
+    fi
     
+    if [ -f ${COUPLE_DIR}/latest_ocean_forcing_file.nc ]; then
+        rm ${COUPLE_DIR}/latest_ocean_forcing_file.nc
+    fi
```
pgierz, line 14: Would it be useful to instead cache these files with a time stamp? For debugging?

runscripts/awicm-pism/coupling/coupling_echam2ice.functions
```diff
@@ -190,14 +197,14 @@ echam_ice_select_relevant_variables() {
         for file in ${echam6_file_list_for_ice}; do
                 filename=$(basename $file)
                 ${cdo} -s -f nc \
-                        -expr,"orog=geosp/9.81; aprt=aprl+aprc; tsurf=tsurf; swd=srads-sradsu; cc=aclcov; \
-                               emiss=(trads-tradsu)/(5.6703744e-8*tsurf^4); \
+                        -expr,"orog=geosp/9.81; aprt=aprl+aprc; temp2=temp2; tsurf=tsurf; swd=srads-sradsu; cc=aclcov; \
+                               emiss=(trads-tradsu)/(5.6703744e-8*temp2^4); \
```
pgierz, line 201: In general, I do not really like "magic numbers". Give them names in variables before hand ;-) Stefan Boltzmann, Earth gravity, etc. (who knows. maybe we want to simulate ice sheets on Mars one day)

Also, ECHAM's gravitation value is a bit more precisely defined. It probably will never make a difference in the simulation, but at least you are then consistent with the internal representation. Look for `ECHAM_g` (or something similar) I had that somewhere if I remember correctly...

runscripts/awicm-pism/coupling/coupling_ice2echam.functions
```diff
@@ -5,14 +5,34 @@ ice2echam() {
         echo "                =================================================================="
         echo "                 *** S T A R T I N G    ice2echam *** "; echo
 
-        . /home/ollie/lackerma/esm-master/esm-runscripts/functions/all/general_lists.functions
+        if [ "$MACHINE" == "mistral" ]; then
+            module load defaults
+        fi
```
pgierz, line 10: Can be removed, mistral no longer exists.

runscripts/awicm-pism/coupling/coupling_ice2echam.functions
```diff
@@ -137,17 +163,17 @@ update_orography_run_calnoro() {
         fi
         # Make sure dimension order is correct:
         change_lonlat topodata.nc OROMEA
-        module unload netcdf
-        module load netcdf/4.6.0_gnu
+        export LD_LIBRARY_PATH=/sw/spack-levante.2022-02-17/netcdf-fortran-4.5.3-jlxcfz/lib/:$LD_LIBRARY_PATH
```
pgierz, line 166: This looks specific to levante, so it needs a if/then around it

runscripts/awicm-pism/coupling/coupling_ice2echam.functions
```diff
@@ -252,7 +280,7 @@ set_glacial_mask_echam_restart_file() {
         dfile=${dfile%.*}_${RES_echam}grid.nc
 
         cdo -s -setmisstoc,0 $ifile tmp 2>> cdo_stderr_ice2echam; mv tmp $ifile 
-        cdo -s -gec,0.5 $ifile tmp 2>> cdo_stderr_ice2echam; mv tmp $ifile   # PG: Ensure the mask is binary after regridding.
+        cdo -s -gec,${ECHAM_GLACIAL_THRESHOLD} $ifile tmp 2>> cdo_stderr_ice2echam; mv tmp $ifile   # PG: Ensure the mask is binary after regridding.
```
pgierz, line 283: Thanks....that should have been configurable from the outside all along 🤦🏻‍♂️

runscripts/awicm-pism/coupling_dual-hemisphere/_stdout_stderr
```diff
@@ -0,0 +1 @@
+./make_debm.sh: line 67: MOD_PRE.f90: command not found
```
pgierz, line 1: This file can also be removed

mandresm:
I will review this on Monday.

pgierz (approved):
As far as I can see this looks good, but there are a lot of things I only skimmed over. Here it might be good to incorporate some kind of automatic test: if we agree that what is produced in the settings now are OK, we can at least guard against that being changed.

PR #821, opened 2022-09-13
Update from GEOMAR: new FOCIOIFS version and FOCI auto tools

seb-wahl:
This contains the latest developments from GEOMAR. I merged the latest release into the GEOMAR branch, so I don't get why there is a "Can't automatically merge." message shown here. Anyways, what's new:
* FOCIOIFS V2.1 with new runoff mapper, and some other fixes
* FOCI compilation with auto tools. I put this in a while ago to have the same compile as in our old runtime environment. Not used in any production simulations at the moment.
* Improvements in GEOMAR's postprocessing and monitoring.


pgierz:
Resolved merge conflicts (stupid whitespace errors in commented out git status check during the run, which does not work yet.)

mandresm:
I'm updating the tests information with current release. This will take until tomorrow noon. Then we will be able to have a look at the tests here and see what really changed.

configs/components/oifs/oifs.yaml
```diff
@@ -203,6 +220,8 @@ read_icmcl: 0
 generate_namelist: 1
 ensemble: 0
 post_processing: 0
+use_ocean_currents: 0
+solarspectrum: 0
```
mandresm:
```suggestion
use_ocean_currents: False
solarspectrum: False
```

configs/components/oifs/oifs.yaml
```diff
@@ -761,6 +780,27 @@ choose_wam:
...
+        1:
+                add_namelist_changes:
+                        fort.4:
+                                NAEPHY:
+                                        LECURR: ".true."
+        0:
+                add_namelist_changes:
+                        fort.4:
+                                NAEPHY:
+                                        LECURR: ".false."
```
mandresm: This makes sure we use a boolean here on resolution of the `choose`.
```suggestion
        True:
                add_namelist_changes:
                        fort.4:
                                NAEPHY:
                                        LECURR: ".true."
        False:
                add_namelist_changes:
                        fort.4:
                                NAEPHY:
                                        LECURR: ".false."
```

configs/components/oifs/oifs.yaml
```diff
@@ -761,6 +780,27 @@ choose_wam:
...
+
+# For 43r3v2 we can choose solar spectrum
+#choose_solarspectrum: 
+#        1:
+#                add_namelist_changes:
+#                        fort.4:
+#                                NAERAD:  
+#                                        NSOLARSPECTRUM: 1
+
```
mandresm:
```suggestion
```

configs/components/oifs/oifs.yaml
```diff
@@ -856,6 +896,56 @@ choose_version:
                 add_outdata_files:
                         ICMUA: ICMUA
 
+
+# If 1, use Coddington spectrum which reduced stratospheric warm bias
+# Only available in 43r3-v2
+choose_solarspectrum:
+        1:
```
mandresm:
```suggestion
        True:
```

configs/components/oifs/oifs.yaml
```diff
@@ -193,6 +209,7 @@ nlev: ${levels_number}
 prepifs_expid: gu5a
 prepifs_startdate: 19820101
 mip: "cmip5"
+o3_scheme: "cmip6"
```
mandresm: Wouldn't it be better that this is set to `null` and to specify the scenario from the yaml of FOCI? In this way we don't interfere with the AWICM3 configuration, which is mostly based on getting the namelist with most of the predefined parameters.

configs/components/oifs/oifs.yaml
mandresm: @JanStreffing, did you deleted this on the past? Any comments?

configs/components/oifs/oifs.yaml
```diff
@@ -761,6 +780,27 @@ choose_wam:
 # OTHER CONFIGS AND SWITCHES
 #======================================================================================
 
+# For coupled setups we can use ocean currents
+choose_use_ocean_currents:
+        1:
```
mandresm:
```suggestion
        True:
```

configs/components/oifs/oifs.yaml
```diff
@@ -761,6 +780,27 @@ choose_wam:
...
+                add_namelist_changes:
+                        fort.4:
+                                NAEPHY:
+                                        LECURR: ".true."
+        0:
```
mandresm:
```suggestion
        False:
```

mandresm (changes requested):

seb-wahl (commented):
Accepted all of Miguels suggestions.

configs/components/oifs/oifs.yaml
seb-wahl: `o3_scheme: "cmip6"` is the default. Joakim added the new option `cariolle`. So setting `"cmip6"` here is required to be backward compatible.

mandresm: This option adds the changes to the tested `fort.4` in AWICM3, for example https://github.com/esm-tools/esm_tools/actions/runs/3134531104/jobs/5089180243#step:8:1023

I think we need the input from @JanStreffing, in any case, if this is a problem in the end, it's as simple as defining `o3_scheme: null` and setting it `o3_scheme: cmip6` as default in `foci-oifs.yaml`

configs/components/oifs/oifs.yaml
```diff
@@ -856,6 +887,56 @@ choose_version:
...
+                        fort.4:
+                                NAMGFL:
+                                        YO3_NL%LGP: '.true.'
+                                NAEPHY:
+                                        LEO3CH: '.true.'
+                                NAERAD:
+                                        LEPO3RA: '.true.' # use progn O3 in radiation
+                                        LCMIP6O3: '.true.'                                        
+                                        NGHGRAD: 20       # overwrite O3 by progn O3
+
```
mandresm: Hi @joakimkjellsson and @JanStreffing,

This is the last conflict missing a resolution in the merging of `geomar` branch into `release`, and the only that is not straight forward to solve. The problem is the next: AWI-CM3 setups make extensive use of different namelists templates to handle most of the namelists variables, while GEOMAR models use switches in combination with `namelist_changes`. The conflicts occur with the extra switches added to this PR: `use_ocean_currents`, `solarspectrum` and `o3_scheme`.

For both, `solarspectrum` and `o3_scheme` the problem posed for AWI-CM3 is not sever as setting the selector variables to `null` or empty will mean no case of the `choose_` is met. However, for `use_ocean_currents`, `null` and `false` will lead to modifications on the namelists, and this will result in unwanted changes on the namelists for AWI-CM3.

My suggestion here is that, for maximum security and compatibility, we put all of these switches (`use_ocean_currents`, `solarspectrum` and `o3_scheme`) inside a new switch `choose_extra_switches`, and make `extra_switches: false` for AWI-CM3, so that the inner choose_ blocks are never reached by AWI-CM3. What do you guys think about this idea?

Please, let me know if you'd like to have a short meeting about this, if you think this cannot be resolved on GitHub.

Any other ideas?

Would you like to have the `extra_switches` into a `further_reading` for keeping the `oifs.yaml` cleaner, or have them in the `oifs.yaml` as a more explicit approach?

mandresm (changes requested):

configs/components/oifs/oifs.yaml
JanStreffing: Could the problem be solved by setting `use_ocean_currents: true` for the AWI-CM3 versions that currently use default templates where the ocean current coupling is turned on? (v3.1 and master)

joakimkjellsson: Hm, maybe we should have a chat about this. I've been unsure how to deal with "solarspectrum" myself since it only works for OpenIFS 43r3v2 (the namelist variable does not exist in v1). 

My thoughts are: 
1) I don't like the idea of namelist templates, and I thought the whole point with ESM-Tools would be to use one single template, and then control everything via switches. 
We use templates for the xml files in XIOS (since switches don't work) and it's really terrible! 

2) I've tried to make it so that setting any of these switches to "default" or "0" or "no" would give you the default OpenIFS behaviour. For example: Unless you set use_ocean_currents, it will be "0", which makes "LECURR = false" and ocean surface velocity never goes into the computation of surface wind stress, as is ECMWF default. Similarly "o3_scheme: default" gives you ECMWF default behaviour (which is not CMIP compatible). 

Let's have a chat about this. Friday 7th at 9am? 

Best wishes
Joakim

mandresm: Hi Joakim, that time would work for me.

seb-wahl: Same here. I'm available Friday at 9am. Let's hope that @JanStreffing has time too.

configs/components/oifs/oifs.yaml
```diff
@@ -856,6 +887,56 @@ choose_version:
...
+# There are two different O3 concentrations in OpenIFS:
+# One is used for output, and one that is used by radiation scheme
+# By default, the two are NOT the same. 
+# By default, prognostic O3 from the Cariolle scheme is written to output
+# but climatology is used by radiation scheme
+# 
+# Here the scheme can be changed in two ways:
+# cmip6: Use O3 from CMIP6 (input4MIP) for picontrol, historical or scenarios
+# cariolle: Use the prognostic O3 from Cariolle in radiation
+choose_o3_scheme:
```
mandresm, line 896: Add a default option:
```yaml
               add_namelist_changes:
                        fort.4:
                                NAERAD:
                                        NGHGRAD: 21
```

mandresm: `oifs.yaml` -> `o3_scheme: "default"`

configs/components/oifs/oifs.yaml
```diff
@@ -761,6 +780,18 @@ choose_wam:
...
+        True:
+                add_namelist_changes:
+                        fort.4:
+                                NAEPHY:
+                                        LECURR: ".true."
+        False:
+                add_namelist_changes:
+                        fort.4:
+                                NAEPHY:
+                                        LECURR: ".false."
```
mandresm: Move it to `focioifs.yaml`.

mandresm (changes requested):

JanStreffing (approved):

mandresm:
Hi @JanStreffing, I guess it could be resolved that way too. Would you prefer that (i.e. leave the `choose_use_oceancurrents` in `oifs.yaml` and set `use_oceancurrents: true` in the `awicm3.yaml` inside the `master` and `v3.1` cases)?

configs/components/oifs/oifs.yaml
```diff
@@ -856,6 +875,61 @@ choose_version:
...
+#
+# Here the scheme can be changed in two ways:
+# cmip6: Use O3 from CMIP6 (input4MIP) for picontrol, historical or scenarios
+# cariolle: Use the prognostic O3 from Cariolle in radiation
+choose_o3_scheme:
+        default:
+                add_namelist_changes:
+                        fort.4:
+                                NAERAD:
+                                        NGHGRAD: 21
```
mandresm, line 901: @JanStreffing: This is adding to the `fort.4` `NGHGRAD: 21` for `AWICM3` simulations that did not have this parameter (`v3.0`, `v3.1`, `frontiers-xios`). If that's okay we can leave it as it is. If it's not okay, we will need to set `o3_scheme: null` in `awicm3.yaml`, so that none of this cases are selected.

mandresm: You can see the changes in the `fort.4` in the work folder: https://github.com/esm-tools/esm_tools/actions/runs/3203196145/jobs/5232999271#step:8:666

mandresm: The default value in the source code it's 21 anyway, so we will keep the default configuration as it currently is.

mandresm:
#approve-changes

mandresm (approved):

mandresm:
#bump #minor

PR #824, opened 2022-09-18
Echam6 wiso nudging

pgierz:
Includes new code required for ECHAM6 to be run in nudging mode as a scenario. 

@mwerner-awi, you can use this pull request to keep an eye on our progress putting this in (should be fast once Miguel and I clarify how much we want this in the test suite, I would guess. More info on that below)

@mandresm: Martin has 2 projects (if I remember correctly) that will require nudging, and while that is not a particularly common use case among the majority of our current users, it might be good to also get this under the testing suite. What would you need to get that to happen? I guess an example runscript? For those who do not know, the `esm-tests` program which Miguel designed runs against every change in the code, and ensures that the generated compile and run scripts match against a version which we know "works" (barring any weird hardware things which we cannot really test against)

There were also a few "pain points" Martin and I discussed in getting this implemented, so maybe we could briefly sit together (also with Martin, and virtually if needed) to see how we can make it easier to have end-users do a form of "self service implementation" for new things we did not think of. Maybe also @JanStreffing could give us some hints, but (fortunately for us) Jan is way deeper into the core than the everyday user. Still, it would good to know if we need documentation updates, better examples, and so forth.

@seb-wahl for the Kiel side: Nothing I saw here would bite with the GEOMAR runs using ECHAM, but maybe just for info in case anyone in Kiel also wants to use that sort of setup, this will be available in the main branch soon.


pgierz:
For now, I will keep this PR in draft mode until we have a brief chance to discuss.

configs/components/echam/echam.yaml
```diff
@@ -352,6 +352,9 @@ choose_scenario:
                         swflux: 1950swflux
                         volcir: 1950volcir
                         volcsw: 1950volcsw
+        # new scenario NUDG for ECHAM6-wiso nudged simulations (M. Werner, 09/2022)
+        NUDG:
```
pgierz: To my knowledge, there is no requirement for scenario names to be short, so we can be explicit here. Small level peanuts changes:

```suggestion
        NUDGING:
```

configs/components/jsbach/jsbach.yaml
```diff
@@ -297,6 +297,10 @@ choose_version:
                                 forcing_files:
                                         LU: LU_1950
                                         LU_trans: LU_trans_none
+                        # new scenario NUDG for ECHAM6-wiso nudged simulations (M. Werner, 09/2022)
+                        NUDG:
```
pgierz: As above:
```suggestion
                        NUDGING:
```

namelists/echam/6.3.05p2-wiso/NUDG/namelist.echam
```diff
@@ -0,0 +1,131 @@
...
+    lpost = 0
+/
+&set_stream ! 'default veg'
+    stream = 'veg'
+    lpost = 0
+/
+&set_stream ! 'default yasso'
+    stream = 'yasso'
+    lpost = 0
+/
```
pgierz: We might need to check that these output file settings match up with the movement (copy, move, or link) of output to the correct folders

pgierz (commented):
All changes make sense, but there are some small considerations we might need to think about before we integrate into the main branch.

mandresm:
> Martin has 2 projects (if I remember correctly) that will require nudging, and while that is not a particularly common use case among the majority of our current users, it might be good to also get this under the testing suite. What would you need to get that to happen? I guess an example runscript?

Essentially that's the only thing I need. If you have different runscripts for support in different computers (where the differences are mainly the pool paths) I'll be interested in having all of them.

> There were also a few "pain points" Martin and I discussed in getting this implemented, so maybe we could briefly sit together (also with Martin, and virtually if needed) to see how we can make it easier to have end-users do a form of "self service implementation" for new things we did not think of.

Sure, it would be great to get Martin's input.

configs/components/echam/echam.yaml
```diff
@@ -352,6 +352,9 @@ choose_scenario:
...
+                dataset: r0007
```
mandresm: Doesn't `r0008` have nudging files? If it does, but you want to use `r0007` for a given experiment, wouldn't this belong to the runscript instead, and not to a `NUDG` scenario?

configs/components/jsbach/jsbach.yaml
mandresm: Similarly to the comment above: is this  a general thing for nudging or is just specific to the experiment you are using for the implementation of the nudging support?

runscripts/echam/echam-nudging-ollie-restart-monthly.yaml
```diff
@@ -0,0 +1,258 @@
...
+      - glday
+      - aclcim
+      - rad
+      - sp6h
+      - glim
+      - spim
+      
+    forcing_sources:
+    
+        # ERA5 fording files
```
mandresm:
```suggestion
        # ERA5 forcing files
```

runscripts/echam/echam-nudging-ollie-restart-monthly.yaml
```diff
@@ -0,0 +1,258 @@
...
+        
+        era5_atm: "/work/ollie/paleodyn/nudging/ERA5/atmos/T63/era5${echam.res}${echam.levels}_${echam.current_date!syear}${echam.current_date!smonth}.nc"
+        
+        sst_previous: "/work/ollie/paleodyn/nudging/ERA5/sea/T63/T63_era5sst_$(( ${echam.current_date!syear} - 1 )).nc"
+        sst_current:  "/work/ollie/paleodyn/nudging/ERA5/sea/T63/T63_era5sst_${echam.current_date!syear}.nc"
+        sst_next:     "/work/ollie/paleodyn/nudging/ERA5/sea/T63/T63_era5sst_$(( ${echam.current_date!syear} + 1 )).nc"       
+        sst_next2:    "/work/ollie/paleodyn/nudging/ERA5/sea/T63/T63_era5sst_$(( ${echam.current_date!syear} + 2 )).nc"
```
mandresm: You can use the following to simplify this type of syntax:
```yaml
echam:
    add_forcing_sources:
        sst: "/work/ollie/paleodyn/nudging/ERA5/sea/T63/T63_era5sst_@YEAR@.nc"
    
    add_forcing_additional_information:
        sst:
                - need_year_before
                - need_year_after
                - need_2years_after
```

namelists/echam/6.3.05p2-wiso/NUDG/namelist.echam
```diff
@@ -0,0 +1,131 @@
...
+    out_expname = 'NUDGING_ERA5_T63L47'
+    rerun_filetype = 4
+    delta_time = 450
+    putdata = 6, 'hours', 'last', 0
+    lamip = .true.
+    lnudge = .true.
+    lcouple = .false.
+    lcouple_co2 = .false.
+    default_output = .false.
+    l_orbvsop87 = .true.
```
mandresm: We could control the deviations from a standard namelist, needed for nudging through a `choose_with_nudging` switch. In this way we can combine the different scenarios already available with nudging (a more modular approach)

mandresm (changes requested):
Is nudging conceptually an scenario? If it's not I would suggest that we make a `choose_` bloc specifically for the nudging and avoid mixing it with the scenarios:
```
echam:
    with_nudging: False # Default false, can be set to true within the runscript
    choose_with_nudging:
        True:
            add_namelist_changes:
                [ ... ]
            add_forcing_files:
                [ ... ]
            [ ... ]
```
This should allow to separate information that is specific to scenarios from the information that is specific to the nudging.

A lot of the `forcing_sources` and `forcing_in_work` can probably be generalized for Ollie and Levante, and moved into the `configs/components/echam/echam.dataset.yaml` and `configs/components/jsbach/jsbach.dataset.yaml`.

I'm not entirely sure that all my comments make sense, since they are based on the assumption that nudging is not conceptually an scenario (you can have ssp scenarios with nudging right?).

IMO, whether you would want to generalize and modularize the nudging, or not, depends entirely on how many people will be using this feature, how often they would use it with different scenarios, how many machines... It's of course always better to have it well generalized from the beginning, a good example for a bad generalization its consequences is the current state of the ECHAM and JSBACH yamls.

pgierz:
From my side: I would say even if it is only a handful of people for right now, it  would be useful to generalise this. One never knows how many other people might want this in the future. Additionally, having this in the tools directly is more likely to be around in 2 years (or, well, X amount of time) than in someone's private runscript.

pgierz:
To the point scenario or not: I can here only speculate as my own scientific work never used nudging. However I would (gut feeling) categorise this more as a simulation "feature" than a "scenario". For me that falls broadly into the same bucket as enabling freshwater perturbation (hosing) or enabling isotopes.

mwerner-awi:
Hi Paul,

Thanks fro taking care of this.

During the next week I am fully occupied with a European training school that we are organising at AWI.

I will check the file movements and answer your other questions the week after. There is no hurry to bring it into the main esm_tools branch, but I have just already pushed an example runscript to the new echam6-wiso-nudging branch.

Janina can meanwhile start to work with the version in the echam6-wiso-nudging branch.

Best, Martin



Am 18.09.2022 um 20:33 schrieb Paul Gierz ***@***.******@***.***>>:



@pgierz commented on this pull request.

All changes make sense, but there are some small considerations we might need to think about before we integrate into the main branch.

________________________________

In configs/components/echam/echam.yaml<https://github.com/esm-tools/esm_tools/pull/824#discussion_r973752482>:

> @@ -352,6 +352,9 @@ choose_scenario:
                         swflux: 1950swflux
                         volcir: 1950volcir
                         volcsw: 1950volcsw
+        # new scenario NUDG for ECHAM6-wiso nudged simulations (M. Werner, 09/2022)
+        NUDG:


To my knowledge, there is no requirement for scenario names to be short, so we can be explicit here. Small level peanuts changes:

⬇️ Suggested change

-        NUDG:
+        NUDGING:


________________________________

In configs/components/jsbach/jsbach.yaml<https://github.com/esm-tools/esm_tools/pull/824#discussion_r973752524>:

> @@ -297,6 +297,10 @@ choose_version:
                                 forcing_files:
                                         LU: LU_1950
                                         LU_trans: LU_trans_none
+                        # new scenario NUDG for ECHAM6-wiso nudged simulations (M. Werner, 09/2022)
+                        NUDG:


As above:

⬇️ Suggested change

-                        NUDG:
+                        NUDGING:


________________________________

In namelists/echam/6.3.05p2-wiso/NUDG/namelist.echam<https://github.com/esm-tools/esm_tools/pull/824#discussion_r973759239>:

> +&mvstreamctl ! 'default spm glm g3bm'
+    filetag = 'echam'
+    source = 'sp', 'gl', 'g3b'
+/
+&mvstreamctl ! 'default co2m'
+    filetag = 'co2'
+    source = 'co2'
+/
+&mvstreamctl ! 'default surfm'
+    filetag = 'surf'
+    source = 'surf'
+/
+&mvstreamctl ! 'default tracerm'
+    filetag = 'tracer'
+    source = 'tracer'
+/
+&mvstreamctl ! 'default jsbachm'
+    filetag = 'jsbach'
+    source = 'jsbach'
+/
+&mvstreamctl ! 'default landm'
+    filetag = 'land'
+    source = 'land'
+/
+&mvstreamctl ! 'default vegm'
+    filetag = 'veg'
+    source = 'veg'
+/
+&mvstreamctl ! 'default yassom'
+    filetag = 'yasso'
+    source = 'yasso'
+/
+&set_stream ! 'default jsbach'
+    stream = 'jsbach'
+    lpost = 0
+/
+&set_stream ! 'default land'
+    stream = 'land'
+    lpost = 0
+/
+&set_stream ! 'default veg'
+    stream = 'veg'
+    lpost = 0
+/
+&set_stream ! 'default yasso'
+    stream = 'yasso'
+    lpost = 0
+/


We might need to check that these output file settings match up with the movement (copy, move, or link) of output to the correct folders

configs/components/echam/echam.yaml
pgierz: Which files are directly needed for nudging? Maybe it is an idea to add these as a separate listing instead of from a dataset? That would gain some independent control of the dataset and the nudging files, unless there is a good reason to link these two together.

runscripts/echam/echam-nudging-ollie-restart-monthly.yaml
pgierz: Is the `need_2years_after` a standard? I had not seen that before.

namelists/echam/6.3.05p2-wiso/NUDG/namelist.echam
pgierz: Agree with Miguel: we should only get the exact changes required that differ between a standard namelist and a nudging friendly one. I'll make a comparison later this afternoon.

mwerner-awi:
Hi @pgierz and @mandresm,

I've finally found some time to answer your questions (see my specific comments above).

Just two more general remarks:

- I fully agree that nudging is not a scenario, but rather an optional mode how to run the ECHAM model. Nudging can be applied to different scenarios. I just added the `NUDG` scenario, because I didn't find any other possibility to change the default dataset linked to a specific scenario. E.g., the `PI` scenario uses the `r008` dataset, but for our specific experiment we needed `r007`. If there is a way to set/change the dataset via the yaml runscript, this would be a much better solution.
- Most changes in `NUDG/namelist.echam` can probably be done via the yaml runscript of a specific experiment. However, I did not know any solution how to change setting of different ECHAM output streams via the runscript. If this is somehow possible, the new `NUDG/namelist.echam` could become obsolete.

If these two issues can be solved, one might not even need to implement an own `choose_` bloc specifically for the nudging (https://github.com/esm-tools/esm_tools/pull/824#pullrequestreview-1112518365), as everything could be controlled by an experiment-specific yaml runscript.

mandresm:
Hi @mwerner-awi,



> * I fully agree that nudging is not a scenario, but rather an optional mode how to run the ECHAM model. Nudging can be applied to different scenarios. I just added the `NUDG` scenario, because I didn't find any other possibility to change the default dataset linked to a specific scenario. E.g., the `PI` scenario uses the `r008` dataset, but for our specific experiment we needed `r007`. If there is a way to set/change the dataset via the yaml runscript, this would be a much better solution.

Every variable defined in a choose in a configuration file (`esm_tools/configs`) can be changed already through the runscript, no matter if that is part of a choose_ block in the configuration files. The way to do that from the runscript will simply be:
```yaml
echam:
    dataset: r0007
```

> * Most changes in `NUDG/namelist.echam` can probably be done via the yaml runscript of a specific experiment. However, I did not know any solution how to change setting of different ECHAM output streams via the runscript. If this is somehow possible, the new `NUDG/namelist.echam` could become obsolete.

It is not yet possible, but we are already designing a feature for that, and we will start prototyping it in the next months. In the mean time, I suggest we keep the `NUDG/namelist.echam` and you can actually select the namelist dir of the namelist also from the runscript with:
```yaml
echam:
    namelist_dir: "${general.esm_namelist_dir}/echam/${version}/NUDG"
```

> If these two issues can be solved, one might not even need to implement an own `choose_` bloc specifically for the nudging ([#824 (review)](https://github.com/esm-tools/esm_tools/pull/824#pullrequestreview-1112518365)), as everything could be controlled by an experiment-specific yaml runscript.

That is true, to me the question is if there are enough things that can be generalized for nudging, that is actually worth reducing the length of the runscript.

mwerner-awi:
Hi @mandresm,

My questions/requests have already been solved. See my comments for the NUDG/namlist.echam and other file changes of my latest commit.

> That is true, to me the question is if there are enough things that can be generalized for nudging, that is actually worth reducing the length of the runscript.

Given the few number of persons running nudged simulations at the moment, I don't think that's necessary.

pgierz:
Hi @mandresm: since @mwerner-awi is happy, what do we still need to clean up to put this in the main branch?

mandresm (commented):
I only have one comment, it seems that the `NUDG` folders for the namelists have been removed, except for `namelists/jsbach/3.20/NUDG/no_dynveg`. Should it be moved to `namelists/jsbach/3.20/HIST/no_dynveg`? @mwerner-awi ?

mwerner-awi:
@mandresm No, it can just be deleted.

Am 17.11.2022 um 11:46 schrieb Miguel ***@***.******@***.***>>:



@mandresm commented on this pull request.

I only have one comment, it seems that the NUDG folders for the namelists have been removed, except for namelists/jsbach/3.20/NUDG/no_dynveg. Should it be moved to namelists/jsbach/3.20/HIST/no_dynveg? @mwerner-awi<https://github.com/mwerner-awi> ?

mandresm (approved):
Thanks Martin!

mandresm:
#bump #patch

PR #829, opened 2022-09-28
Medusa suggestions

mandresm:
Hi Ying,

Find here my suggestions. This won't let you run fesom-medusa yet, but is a step closer to start testing the offline coupling tomorrow. I left some comments on the editions that should answer your questions.

As soon as you approve these changes I will merge them into `feat/medusa`. Once this is merged, can you test that `runscripts/medusa-fesom/fesom-recom-run.yaml` works as a standalone runscript (comenting** the `workflow` part) and that everything looks as it should? This is necessary so that in the next step we can focus on the coupling and we are not distracted by fesom not doing the right thing.


a270105 (approved):
Thank you!

a270105:
Hi Miguel, it is not clear to me where to add the part calling postprecessing_fesom.sh with the input variables. Should it be in fesom.yaml?

mandresm:
> Hi Miguel, it is not clear to me where to add the part calling postprecessing_fesom.sh with the input variables. Should it be in fesom.yaml?

No, just in that line, if that script needs input parameters you can write just on that line the parameters, e.g.:
```yaml
    sript: postprocessing_fesom.sh <param1> <param2> ...
```
The parameters can be esm-tools variables (i.e. `${experiment_dir}`) and they will be substituted before calling the script.

PR #839, opened 2022-10-13
Fix/conda env

mandresm:
Adds an `environment.yaml` for conda. Instructions can be found in the comments of the `utils/environment.yaml`.


pgierz (approved):

mandresm:
@pgierz, just to be clear, I don't think conda is the solution to our problems, but I have found that installing ESM-Tools via conda in my mac was a good way of protecting myself from conflicts with autosubmit. Therefore, I thought I will just push this changes.

mandresm:
#bump

pgierz:
This already went in, but to follow up: "Shielding" is also what the `venv` is supposed to do. I guess the clear message here is that we need some sort of strategy for doing this (even if it is only for developers), and it should be one that we agree on and know how to support. I'm happy to volunteer to write dev-friendly documentation on that, and deal with confused users if we choose a strategy.

I personally have no problem against more active publishing to PyPI for `pip` support (actually, I still do not know why there was so much friction against this from Dirk), or even publishing to the conda service, but I would need to figure out how to do that (and how to do it automatically, in an ideal case).

PR #840, opened 2022-10-13
Feat/ssp generalization

mandresm:
Fixes several problems with ssp simulations in multiple versions of ECHAM.

**TODO**
- [ WON'T DO] Integration tests for standalone-ECHAM ssp simulations
- [x] Antonio's tests
- [x] Pengyang tests
- [x] Check Jule's namelists reported by Christian in #673 
- [x] Check 2014 historical run

@antoniofis10000, @PengyangSong and @l975421700, I'd like to ask you for a last effort on testing that your simulations can run on this branch. Then it will be merge into `release` and the problem should be solved, or at least have set a good base for future ssp issues to be fastly resolved.


pgierz:
Update of "test-correctness" seems to be need. @mandresm, how did that work? Do I just do `number-sign approve_changes`? 

If I nuke anything with this, please forgive me
#approve_changes

pgierz:
Or is it?
#approve-changes

christian-stepanek:
Okay. @l975421700 is trying to run a scenario in standalone-atmosphere configuration. I am pretty sure that this has not been done before. So far our scenario simulations have been coupled atmosphere-ocean. So, potentially one needs to use the SSP5 namelist settings that I posted here (based on fully coupled AO) https://github.com/esm-tools/esm_tools/discussions/673#discussioncomment-3877382 and "build back" to atmosphere-only settings. This will certainly involve providing in addition the SST and SIC forcing files (unit.20 for sst, unit.96 for sic), which of course have to stem from a transient simulation that has been run with interactive ocean.

configs/components/echam/echam.yaml
```diff
@@ -382,26 +382,41 @@ choose_scenario:
...
+                        message: "
+                            You are trying to run a ``ssp`` simulation with the
+                            ``echam.dataset: ${dataset}``. Some input files might be
+                            missing for ssp simulations in that dataset. It's
+                            recommended that you use the ``r0008`` dataset instead.\n\n
+                            If you wish to continue with the current dataset answer
+                            ``Y`` to the question. If you want to switch to
+                            ``r0008`` dataset, answer ``n`` and modify your runscript
+                            accordingly (``echam.dataset: r0008``)"
+                        ask_user_to_continue: True
```
pgierz: While this works, what about others? Dataset 9, 10, 11 (11 doesn't exist yet, but who knows). I would recommend rather than doing a simple string comparison, we do some kind of greater-than check, but that might mean digging into the parser, which I do not what to recommend.

configs/components/echam/echam.yaml
```diff
@@ -382,26 +382,41 @@ choose_scenario:
...
-                scenario_type: cmip6
-
-        ssp245:
-                scenario_type: cmip6
-
-        ssp534os:
-                scenario_type: cmip6
-
         cmip6hist:
                 scenario_type: cmip6
```
pgierz, line 399: This syntax annoys me. We should have some way of doing an "or", but that requires touching the parser, which I do not want to do, unless I get to blow it up completely and re-write it from scatch.

configs/components/jsbach/jsbach.yaml
```diff
@@ -118,7 +102,7 @@ choose_version:
                         false:
                                 namelist_dir: "${general.esm_namelist_dir}/jsbach/${version}/${echam.scenario}/no_dynveg"
                                 jsbachnc_file_ending: "${no_dynveg_file_ending}"
-                                dynveg: "_no_dynveg"
+                                dynveg: "_no-dynveg"
```
pgierz, line 105: This discrepancy really needs to go away at some point; please check as it might be badly named between Levante and Ollie.

namelists/echam/6.3.04p1/cmip6/namelist.echam
```diff
@@ -0,0 +1,139 @@
...
+	target = 'sp6h'
+	source = 'sp'
+	variables = 'st:mean', 'svo:mean', 'lsp:mean', 'sd:mean'
+/
+&mvstreamctl
+	interval = 1, 'hours', 'last', 0
+	target = 'g3b1hi'
+	source = 'g3b'
+	variables = 'u10:inst', 'v10:inst', 'wimax:max'
+/
```
pgierz, line 118: We have talked about this, but I at some point really would be happy to separate the model physical settings and the output choice settings, at least on the tools side as far as possible. Miguel suggested to merge namelists together with add and remove functionally at some point, I would strongly advocate to follow that way of thinking. But it will be tricky...

Determine model (I/)O (I guess O) is something I really at some point want to get a decent handle on.

configs/components/echam/echam.yaml
mandresm: On the other hand this is an error message very specific to our current situation, so generalizing might not be a good idea, at least for now. One could always generalize it like this: `data_set_lower_than_r0008: "$(( int('${dataset}'[-4:]) < 8 ))"`, but I don't think we really need that right now.

configs/components/echam/echam.yaml
mandresm: This syntax is already removed and already substituted by this:
https://github.com/esm-tools/esm_tools/blob/71cc07e251fba517902e01f6347e8afeab7f917b/configs/components/echam/echam.yaml#L388-L393

It's not the most terrible solution and we don't need to touch the parser.

namelists/echam/6.3.04p1/cmip6/namelist.echam
mandresm: I think the tricky part would be to get an agreement on what we should provide as defaults selections for the output, and again defaults for the "physical" parts of the default echam and jsbach namelists. In technical terms, we have already a function that does something very similar, and in the end is a matter of extending it a bit to append some text. We can open a mini project for that.

mandresm:
> Okay. @l975421700 is trying to run a scenario in standalone-atmosphere configuration. I am pretty sure that this has not been done before. So far our scenario simulations have been coupled atmosphere-ocean. So, potentially one needs to use the SSP5 namelist settings that I posted here (based on fully coupled AO) [#673 (comment)](https://github.com/esm-tools/esm_tools/discussions/673#discussioncomment-3877382) and "build back" to atmosphere-only settings. This will certainly involve providing in addition the SST and SIC forcing files (unit.20 for sst, unit.96 for sic), which of course have to stem from a transient simulation that has been run with interactive ocean.

Thanks for this insight @christian-stepanek. In this case, this is then more of an issue for the user to solve as this is part of the experiment configuration itself. @l975421700, of course we will be willing you to assist you with the ESM-Tools parts of how to do this, but firstly you will need to know which forcing files you want to use for SST and SIC and build back the atmosphere-only settings from the echam namelist as Christian said. Consequently, the merge of this branch won't wait until your problem is solved, and it will be merged as soon as the reviews are finished.

QinggangGao:
Hi Miguel

Thanks. I understand and I will work on the atmospheric-only ssp585 simulation later.

Regards, Qinggang

mandresm:
Email from @PengyangSong:
> Dear all,
>
> I still have some questions regarding running SSP scenario with ESM-tools.
>
> I restart from a 500-yr spinup, and then run from 1851-2100.
>
> There are mainly two problems:
>
>1. In year 2014, cannot find file ozone2015 from folder /work/ollie/pool/ECHAM6/input/r0007/T63/ozone/...>>
>
>2. In year 2099, cannot find file ozone2100 from folder /work/ollie/pool/ECHAM6/input/r0008/T63/ozone/...>
>
>
>My solution is below for echam.scenario: "ssp585":
>
>    from 1851-2013: echam.dataset = r0007
>
>    2014: echam.dataset = r0008
>
>              forcing_sources.greenhouse_hist_scenario: "/work/ollie/pool/ECHAM6/input/r0007/greenhouse_historical.nc"
>
>              forcing_sources.MAC-SP_hist_scenario: "/work/ollie/pool/ECHAM6/input/r0007/MAC-SP.nc"
>
>    2015-2098: echam.dataset = r0008
>
>    2099-2100: cannot run....
>
>
>Is it also right from your sides? that a future scenario should start at year 1851 and stop at year 2098?
>
>Kind regards,
>
>Pengyang

christian-stepanek:
> Email from @PengyangSong:
> 
> > Dear all,
> > I still have some questions regarding running SSP scenario with ESM-tools.
> > I restart from a 500-yr spinup, and then run from 1851-2100.
> > There are mainly two problems:
> > 
> > 1. In year 2014, cannot find file ozone2015 from folder /work/ollie/pool/ECHAM6/input/r0007/T63/ozone/...>>
> > 2. In year 2099, cannot find file ozone2100 from folder /work/ollie/pool/ECHAM6/input/r0008/T63/ozone/...>
> > 
> > My solution is below for echam.scenario: "ssp585":
> > from 1851-2013: echam.dataset = r0007
> > 2014: echam.dataset = r0008
> > ```
> >          forcing_sources.greenhouse_hist_scenario: "/work/ollie/pool/ECHAM6/input/r0007/greenhouse_historical.nc"
> > 
> >          forcing_sources.MAC-SP_hist_scenario: "/work/ollie/pool/ECHAM6/input/r0007/MAC-SP.nc"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 2015-2098: echam.dataset = r0008
> > 2099-2100: cannot run....
> > Is it also right from your sides? that a future scenario should start at year 1851 and stop at year 2098?
> > Kind regards,
> > Pengyang

My feeling is that the problem is related to the fact that ECHAM needs ozone data sets that go beyond the simulation period. If SSPs are defined until 2100, and the forcing is available until 2100, then per definition ECHAM cannot be run for the whole period. We have seen this problem also in other cases (e.g. the work by Jule Schrepfer). It would be interesting to know if anybody has taken the endeavor to overcome this problem by preparing additional forcing files for a few years beyond the simulation period. Such forcings may exist, but maybe have never been prepared for the use with ECHAM6. I do not know whether the problem is critical, whether one can just be satisfied with two years of simulation missing at the end of the scenario run. One could think that there may be a problem if one ever wanted to run simulations for an extended period, based on fixed forcings of 2100 or similar, which then likely would not fully correspond to the forcing defined for the scenario.

mandresm:
Note to @pgierz: I've reviewed the tests differences, with special attention to AWICM1-REcoM, AWIESM-2 and FOCI. The differences are just new variables used for the generalization, but they make no impact in the namelists (unchanged) and which files are copied in the supported tests.

mandresm:
Another note for @pgierz: running `/home/ollie/mandresm/my_scripts/psong_1851_r0008.yaml` which has `echam.dataset: r0008` leads to the following missing files (which are not missing in `r0007` folder. Those files are neither present in the `r0008` folder of Levante:
```
- missing source: /work/ollie/pool/ECHAM6/input/r0008/greenhouse_historical.nc
- missing target: /work/ollie/mandresm/psong_1851_r0008/run_18510101-18510131/forcing/echam/greenhouse_gases.nc

- missing source: /work/ollie/pool/ECHAM6/input/r0008/MAC-SP.nc
- missing target: /work/ollie/mandresm/psong_1851_r0008/run_18510101-18510131/forcing/echam/MAC-SP.nc
```

Those missing files are probably the reason why @PengyangSong uses `dataset: r0007` for before 2014.

I've looked in the FOCI pool in HLRN, they have a `MAC-SP.nc` and a `greenhouse_historical.nc` in the `r0008` dataset. Here is where I lack the knowledge, are the r0008 dataset folders in Levante and Ollie incomplete?

pgierz:
Currently, it seems that if you do not specify anything, you get dataset revision `r0007`. I would recommend to remove this default setting and tie each version of `ECHAM` to a particular dataset (that is also what was done for `MPI-ESM`). See 466c1fbd.

mandresm:
> Currently, it seems that if you do not specify anything, you get dataset revision `r0007`. I would recommend to remove this default setting and tie each version of ECHAM to a particular dataset (that is also what was done for `MPI-ESM`). See [466c1fb](https://github.com/esm-tools/esm_tools/commit/466c1fbdcfb219a62e1b6dd4a78b1a3c4059869b).

That is awesome, thanks!

pgierz:
Hi @christian-stepanek, please see eb9bc912 for a fix for the ozone problem. You will now always get the ozone for year 2099 for any ozone forcing beyond 2099.

christian-stepanek:
Hey Paul, great! Thanks a lot.
Was ozone the only file that caused a problem? Or was there more to be fixed?
I will bump Pengyang, who raised the issue.

pgierz:
As far as I am aware, ozone was the only problematic file. Of course we can expand the modifications to include other "repeating last year" cases if something else acts up.

mandresm:
I am comparing the namelists posted by @christian-stepanek [here](https://github.com/esm-tools/esm_tools/discussions/673#discussioncomment-3877382) (belonging to Jule's simulation), with the namelists that are provided by default with AWIESM-2.1. For the `namelist.jsbach` (`esm_tools/namelist/jsbach/3.20/ssp126/dynveg/namelist.jsbach`) the one provided by Christian is missing the following lines:
```diff
&jsbach_ctl
    standalone = .false.
    ntiles = 11
    use_bethy = .true.
    use_phenology = .true.
    use_albedo = .true.
    with_yasso = .true.
    with_hd = .true.
    use_roughness_lai = .true.
    use_roughness_oro = .false.
    veg_at_1200 = .false.
    use_dynveg = .true.
    use_disturbance = .true.
    with_nitrogen = .true.
    lcc_forcing_type = 'transitions'
    lcc_scheme = 2 
    test_cconservation = .true.
    test_nconservation = .true.
/
&soil_ctl
    nsoil = 5 
/
&hydrology_ctl
-    gethd = 192, 'steps', 'exact', 0
-    puthd = 192, 'steps', 'exact', -450
    diag_water_budget = .true.
/
&cbalance_ctl
    read_ndepo = .true.
/
&disturbance_ctl
    fire_name = 'thonicke'
    fuel_name = 'thonicke'
    lburn_pasture = .true.
/
&input_ctl
    var_name = 'lightning_frq'
    var_file_name = 'HRMC_COM_FR'
    file_name = 'lightning.nc'
    action_cycle = 'all'
/
-&input_ctl
-    var_name = 'population_density'
-    dt_update = 0 
-    init_rec = 1 ! only read first record of population_density file (i.e. year 1850)
-/
```

Should we keep as default what we already have, or do we need to remove the lines for the default version?

namelists/echam/6.3.04p1/cmip6/namelist.echam
mandresm: I have added a project for this, and therefore I'll mark this by resolved.

mandresm:
> Another note for @pgierz: running `/home/ollie/mandresm/my_scripts/psong_1851_r0008.yaml` which has `echam.dataset: r0008` leads to the following missing files (which are not missing in `r0007` folder. Those files are neither present in the `r0008` folder of Levante:
> 
> ```
> - missing source: /work/ollie/pool/ECHAM6/input/r0008/greenhouse_historical.nc
> - missing target: /work/ollie/mandresm/psong_1851_r0008/run_18510101-18510131/forcing/echam/greenhouse_gases.nc
> 
> - missing source: /work/ollie/pool/ECHAM6/input/r0008/MAC-SP.nc
> - missing target: /work/ollie/mandresm/psong_1851_r0008/run_18510101-18510131/forcing/echam/MAC-SP.nc
> ```
> 
> Those missing files are probably the reason why @PengyangSong uses `dataset: r0007` for before 2014.
> 
> I've looked in the FOCI pool in HLRN, they have a `MAC-SP.nc` and a `greenhouse_historical.nc` in the `r0008` dataset. Here is where I lack the knowledge, are the r0008 dataset folders in Levante and Ollie incomplete?

@pgierz, @christian-stepanek, I still have this problem. In summary, `greenhouse_historical.nc` and `MAC-SP.nc` are missing in the `r0008` dataset of the pool, but weirdly, not in the `r0007`. I think that's also the case in Levante. I don't really know what to do about this, specially if that's the case in the Levante pool: isn't that the one that was put together by MPI?

christian-stepanek:
> I am comparing the namelists posted by @christian-stepanek [here](https://github.com/esm-tools/esm_tools/discussions/673#discussioncomment-3877382) (belonging to Jule's simulation), with the namelists that are provided by default with AWIESM-2.1. For the `namelist.jsbach` (`esm_tools/namelist/jsbach/3.20/ssp126/dynveg/namelist.jsbach`) the one provided by Christian is missing the following lines:
> 
> ```diff
> &jsbach_ctl
>     standalone = .false.
>     ntiles = 11
>     use_bethy = .true.
>     use_phenology = .true.
>     use_albedo = .true.
>     with_yasso = .true.
>     with_hd = .true.
>     use_roughness_lai = .true.
>     use_roughness_oro = .false.
>     veg_at_1200 = .false.
>     use_dynveg = .true.
>     use_disturbance = .true.
>     with_nitrogen = .true.
>     lcc_forcing_type = 'transitions'
>     lcc_scheme = 2 
>     test_cconservation = .true.
>     test_nconservation = .true.
> /
> &soil_ctl
>     nsoil = 5 
> /
> &hydrology_ctl
> -    gethd = 192, 'steps', 'exact', 0
> -    puthd = 192, 'steps', 'exact', -450
>     diag_water_budget = .true.
> /
> &cbalance_ctl
>     read_ndepo = .true.
> /
> &disturbance_ctl
>     fire_name = 'thonicke'
>     fuel_name = 'thonicke'
>     lburn_pasture = .true.
> /
> &input_ctl
>     var_name = 'lightning_frq'
>     var_file_name = 'HRMC_COM_FR'
>     file_name = 'lightning.nc'
>     action_cycle = 'all'
> /
> -&input_ctl
> -    var_name = 'population_density'
> -    dt_update = 0 
> -    init_rec = 1 ! only read first record of population_density file (i.e. year 1850)
> -/
> ```
> 
> Should we keep as default what we already have, or do we need to remove the lines for the default version?

In principle I assume it is best to keep what you already have, provided it runs successfully. I remember that Jule and Fernanda were quite busy getting the SSP setup to run. I cannot exclude the possibility that some things were removed that do not necessarily have to be removed.

Yet, on a second thought, is there a good reason to read the population density file for 1850 for a future simulation? Why not that for the most recent year where such information is available? I will have to check what MPI-Met suggests.

christian-stepanek:
> 

I will have to check by getting the newest mpi-esm tarball and setting up a SSP simulation for reference. I thought we took the most recent configurations created by MPI-Met, but it does not hurt to check again. These checks cannot be done before Tuesday due to Levante downtime. (I would like to once more compare all settings against the pool on Levante.)

mandresm:
> I will have to check by getting the newest mpi-esm tarball and setting up a SSP simulation for reference. I thought we took the most recent configurations created by MPI-Met, but it does not hurt to check again. These checks cannot be done before Tuesday due to Levante downtime. (I would like to once more compare all settings against the pool on Levante.)

Thanks !

mandresm:
> > 
> 
> I will have to check by getting the newest mpi-esm tarball and setting up a SSP simulation for reference. I thought we took the most recent configurations created by MPI-Met, but it does not hurt to check again. These checks cannot be done before Tuesday due to Levante downtime. (I would like to once more compare all settings against the pool on Levante.)

Hi @christian-stepanek, any updates on that? @PengyangSong and myself checked whether these files were also missing in the Levante pool, and they are indeed missing. 

@seb-wahl, I've seen in `blogin` that the pool you use for FOCI contains these two files (`greenhouse_historical.nc` and `MAC-SP.nc`) in the `r0008` dataset. Do you have any hint on where did you got them from?

seb-wahl:
I'm pretty sure I got the from the mistral pool around 2017 when mistral was still alive. Feel free to copy them

mandresm:
Thanks @seb-wahl, weirdly enough, those files are not in the Levante pool... I thought they where the same as those from Mistral.

PengyangSong:
> Thanks @seb-wahl, weirdly enough, those files are not in the Levante pool... I thought they where the same as those from Mistral.

is there any folder in the levante pool named "from mistral"?

mandresm:
I have compared both files from @seb-wahl `r0008` dataset in blogin with those available in levante for `r0007`, and they are bit identical.

mandresm:
> is there any folder in the levante pool named "from mistral"?

No, or at least I cannot find it.

configs/components/echam/echam.yaml
chrisdane: It is nice but actually not needed since missing file checks are performed anyway?

configs/components/echam/echam.yaml
```diff
@@ -391,26 +392,41 @@ choose_scenario:
...
+                            ``echam.dataset: ${dataset}``. Some input files might be
+                            missing for ssp simulations in that dataset. It's
+                            recommended that you use the ``r0008`` dataset instead.\n\n
+                            If you wish to continue with the current dataset answer
+                            ``Y`` to the question. If you want to switch to
+                            ``r0008`` dataset, answer ``n`` and modify your runscript
+                            accordingly (``echam.dataset: r0008``)"
+                        ask_user_to_continue: True
+        after_2014: "$(( ${start_date!year} > 2014 ))" # TODO: double check with Sebastian, Antonio, Pengyang, Tido
+        choose_after_2014:
```
chrisdane: I am not sure if I like this approach with `iaero`. It only works in combination with the new file `cmip6/namelist.echam` (below). This namelist, viewed by itself, is very confusing: as a user I am asking myself "What is a general namelist.echam for cmip6? For which experiments will it be used?". I Agree that the only difference between the `historical` and `ssp*`-scenarios is `iaero`. But if you do it like this, then, theoretically, the `piControl`-case (and all other cases) could also be implemented here and it will be more difficult to follow what is going on. But this is just a comment.

configs/components/echam/echam.yaml
mandresm: Agreed... I'll remove it

configs/components/echam/echam.yaml
mandresm: I think the idea here, is that they are all "flavors of CMIP6" correct?

chrisdane: I am not sure if I understand your question. There are a lot of different cmip6-experiments. I would not opt for capturing those by the esm_tools. I like the namelist-approach more.

mandresm: I see what you mean. On the other hand, all this logic with the scenario-type was developed by the FOCI (ECHAM+NEMO) developers some time ago, and has greatly simplified the echam yamls and datasets and make them less error-prone. The idea behind it is that if you select `echam.scenario: ssp585` for example, and your start date is for example 2000, it starts with historical, and then switches to ssp input in 2014. Additionally, if you just want to run a historical simulation, then you just need to select `echam.scenario: cmip6hist`. The `scenario_type` is set automatically in this yaml file for you, and for ssp scenarios or cmip6hist the value that it takes automatically is `cmip6`, which I agree, it's maybe not the most accurate name.

That been said, the ssp logic was developed first in ESM-Tools by people in GEOMAR some time ago, I don't feel like removing that logic, specially because it works pretty well (in fact, I believe this is already part of your ssp simulations configuration correct?)

mandresm: On the other hand, it would be very easy to change the `namelist_dir` here instead of changing `iaero` with a `namelist_changes` and that won't give problems to anyone.

However, the nice thing about having it controlled through a namelist_changes instead of the namelist_dir, is that if you define your own `namelist_dir` in the runscript, the namelist changes takes care of switching iaero once it reaches 2014, otherwise, you'll still run with whatever your original namelist was.

chrisdane: Yes I see the advantages. However, I think the less complex the better. And 
> your start date is for example 2000, it starts with historical, and then switches to ssp input in 2014

means that the namelist will change within one experiment? I would vote absolutely against this. This makes debugging, analysis, etc. way more difficult since it removes the assumption that namelists (and similar stuff) are constant for one experiment (read: a directory). But jeah, keep it as it is if that is the community concensus.

mandresm: IMO this is totally up to the users, I really don't have a strong opinion in favor of one or another, and I am willing to change whatever is needed. I agree  automating everything is not always a good idea. However, there are some users already using this feature (or worse, not having this feature, having gone within the same simulation from a historical scenario to a ssp without changing the iaero), and it seems it's something the MPI-ESM scripts also do (please, @christian-stepanek correct me if I'm wrong).

mandresm: > This makes debugging, analysis, etc. way more difficult since it removes the assumption that namelists

I don't agree with this, all namelist for all runs are saved as `<base_dir>/<exp_id>/config/echam/namelist.echam_<timestamp>`.

mandresm: @christian-stepanek , @PengyangSong , @pgierz , @seb-wahl , can we make a decision here to the question:

- Do we want automatic switching of iaero for simulations that go from historical to ssp times?

pgierz: This may sound trivial but "the user should get the run she/he expected to happen". If I say SSP 585 then that should just be what is done. Or is that not what is currently happening? If not, why not, and how do I help to get that to work?

pgierz: What is the "official" policy on aerosols?

christian-stepanek: My personal opinion is: historical and any of the different SSP scenarios are completely independent simulations. They should therefore be run independently. That means, one runs a historical simulation, and thereafter sets up an SSP simulation that is just initialized from the last time step of the historical simulation.

I am not very familiar with the automatic year-dependent switching of forcing types. There may be reasons that speak for it, but I also see some reasons that speak against it. For example with the SSP scenarios. At the moment it seems that all SSP scenarios use the same iaero data set (I am verifying the data sets on levante - sorry that it takes so long, I had to figure out how to set up the mpi-esm infrastructure on levante as there is absolutely no update to the documentation on the MPI-Met side regarding their modelling infrastructure). But, are we sure that this will always be the case? Maybe in CMIP7 somebody will come up with forcing data sets other than ghg that differ between the various SSP scenarios. How would esm-tools decide which of the various forcing data sets should be chosen at the end of year 2014?

In the end I think there is a tradeoff between flexibility and safety. How flexible shall the esm-tools "make decisions themselves", and how intensely do we ask the modelers to think about their setup - whether everything is the way they expect it to be. If we keep the  switch in 2014 to be done automatically, all this will have to be updated in esm-tools for CMIP7, as it is very likely that the CMIP7 historical simulation will have another run period defined - and maybe other forcing data sets, etc.

In summary: Do we want to hide the details in the esm-tools program structure, or in the scenario definitions. If you asked me, then I actually would opt for the latter.

mandresm: I see.

About the historical to ssp transition: independently of the automatic setting of iaero, nothing prevents the user to start a run in historical period and move into ssp. Should we warn the users that they are doing something that is not recommended?

About the CMIP7 ssp files I have actually a recommendation, and that is we don't know how CMIP7 is going to look like in terms of files, generalizing too much will make the echam.yaml difficult to read, so I would keep things for ssp files as short and simple as possible. When CMIP7 comes we can implement all the complexity needed behind a cmip7 switch and give it it's own template namelists.

christian-stepanek: IMHO a historical simulation should end when the historical forcing ends. So a break at the end of 2014 would be the expected behavior. If somebody wants to run the simulation longer, then they should explicitly think about under which conditions the simulation should continue - and that somehow implies that the esm-tools interrupt the simulation at 2014.

Again, this is just my point of view.

PengyangSong: Notice if we have enough input files for year 2014, because it requires some forcing files at 2015.

pgierz:
I have contacted the DKRZ Helpdesk to clarify the `MAC-SP.nc` and `greenhouse_historical.nc` file issues. Since we are trying to maintain an exact copy of the DKRZ pool, I would like to know if there are good reasons those files are missing in `r0008` (missing) vs `r0007` (present). I could simply make links, however there are probably good reasons for the files to be missing...? In any case, someone at DKRZ can either fix the problem and it will be synced, or they can tell me who at MPI-Met to contact to clarify everything.

configs/components/echam/echam.yaml
chrisdane: > Notice if we have enough input files for year 2014, because it requires some forcing files at 2015.

Yes but thats a purely technical problem.

mandresm: I don't think the question of Pengyang it's purely a technical problem. We can all code a solution for that in the yaml files, it's pretty straight forward. The problem is that we need to make a decision on which files are taken for a 2014 historical simulation for the forcing files that require 2-year-after, do we use the ssp ones? which ones? do we use 2014 twice?

chrisdane: This is a specific technical question and I documented this, its in release: [echam.datasets.yaml:L96-100](https://github.com/esm-tools/esm_tools/blob/release/configs/components/echam/echam.datasets.yaml#L96).

mandresm: Oh true, I did not remember that. So that situation is already handle with the current state of the scenarios. It seems that we have reached a conclusion: by default ESM-Tools will return an error if you try to run historical simulations of echam passed 2014, and the case of  ssp simulations will be the other way around. I'll do that tomorrow if no one complains against this solution. For FOCI people, who want to run simulations from the historical period into the ssp one, this restriction will be removed in the `foci.yaml` file.

mandresm:
We have a reply from MPI!

> version 0007 was an intermediate version of the input data and was not used in CMIP6.

>  For the CMIP6 version, r0008, we decided to append scenario data (2015+) to the historical data (-2014). In ECHAM6, data for the end of 2014 is interpolated from 2014 and 2015 values, so this avoids incontinuities between historical and scenario experiments. All ssp versions contain the same data for the historical period (0-2014). As the default continuation, we used SSP 2 RCP 4.5 (ssp245).

> In effect, when using r0008, uses of MAC-SP.nc or greenhouse_historical.nc should be replaced by MAC-SP_ssp245.nc or greenhouse_ssp245.nc, respectively. You may want to choose other scenarios as your default.

> Note that the data has not been updated to reflect later changes or extensions in the input4MIPs data.

So what I understand from this is that if we are in the `historical` period using `r0008` dataset we need to use the ssp245 files. I'll give a strong push to this branch on Friday, as most of the blockers have been removed.

christian-stepanek:
Great! We still do not not about the iaero files, though, am I right? Those were not available in r0008.

Good that we found out that way. I tried yesterday afternoon successlessly to get the MPI-ESM tools running on levante in order to have a look how an SSP scenario as suggested by MPI-ESM in the newest version of the model would look like. No success, mkexp depends on python2 and there is no python2 on Levante. I was not even able to compile their model anymore ...

configs/components/jsbach/jsbach.yaml
mandresm: the discrepancy is gone :)

configs/components/echam/echam.yaml
mandresm: @PengyangSong, @chrisdane, would this be an acceptable solution then?
Screenshot 2022-11-25 at 19 04 03

mandresm:
@PengyangSong, after the last commit you should not have problems with files missing anymore, the `greenhouse` and `MAC-SP` issues were solved following the advice Kalle (MPI). Let me know if that works for you.

mandresm:
> Great! We still do not not about the iaero files, though, am I right? Those were not available in r0008.
> 
> Good that we found out that way. I tried yesterday afternoon successlessly to get the MPI-ESM tools running on levante in order to have a look how an SSP scenario as suggested by MPI-ESM in the newest version of the model would look like. No success, mkexp depends on python2 and there is no python2 on Levante. I was not even able to compile their model anymore ...

You're right @christian-stepanek, that's the last piece of the puzzle. Thanks for the efforts! l For the python, you could always create an anaconda enviroment that contains a python2 installation. Let me know if you need help with that.

pgierz:
Christian and I worked on that. The solution unfortunately requires installing your own private copy of miniconda and we had a meeting before I could get to that step.

configs/components/echam/echam.yaml
chrisdane: Yeah ok. Maybe also print the initial date that raised the error.

mandresm: done!

pgierz (approved):
Looks good! 👍

mandresm:
Thanks @pgierz, running a last set of actual tests. I'll be merging tomorrow after the tests run.

mandresm:
#approve-changes

mandresm:
#bump

PR #845, opened 2022-11-12
fix/iterative coupling setup separation

mandresm:
# Problem

When doing offline coupling (iterative coupling), the different models run inside the same general experiment folder. When the workflow manager was put together, it was for vilma-pism where the model dates for vilma and pism do not advance together. This means that for the first vilma chunk the run folder will be named differently than the pism run folder, simply because the time stamp will be different.

With fesom-medusa coupling (being done by @a270105), at least in the first testing phase, we are running both fesom and medusa for the same dates, and that means fesom and medusa both run inside the same `run_` folder , which IMO is unsafe and confusing. Furthermore, some files such as the valuable `finished_config.yaml` (and other logging files) are not stamped with the setup name, which means the second model to run in the current chunk will rewrite the previous files. 

Furthermore, even when the model clocks don't run synchronized, it is pretty confusing to recognize to which model a run/finished_config.yaml/log file belongs.

# Solution
The new `config["general"]["iterative_coupled_model"]` takes values of an empty string when there is not iterative coupling, and with iterative coupling it takes the value `<setup_name>_`, and then this is plugged in into the name of the problematic folders and files, for example:

| No-offline coupling    | Offline coupled fesom        | Offline coupled medusa        | 
|------------------------|------------------------------|-------------------------------|
| run_DATE             | run_fesom_DATE             | run_medusa_DATE             | 
| *_finished_config.yaml | *_fesom_finished_config.yaml | *_medusa_finished_config.yaml |


mandresm:
The only difference in all `esm_tests` is the `iterative_coupled_model` variable.

src/esm_runscripts/chunky_parts.py
```diff
@@ -10,7 +10,7 @@ def setup_correct_chunk_config(config):
     if not config["general"].get("iterative_coupling", False):
         return config
 
-    print("Starting the iterative coupling business")
+    print("Initializing iterative coupling")
```
pgierz, line 13: Thank you! Let's keep it professional 👍

pgierz (approved):
There could be some small things to make this more readable, but that is not so important. Only if you want: you often have a pattern such as

```python
f'{config["general"]["experiment_dir"]}/run_'
         f'{config["general"]["iterative_coupled_model"]}'
         f'{config["general"]["run_datestamp"]}'
```

It might be easier to read to just extract these things beforehand and skip the whole config blah blah blah:

```python
experiment_dir = config["general"]["experiment_dir"]}
coupled_model_name = config["general"]["iterative_coupled_model"]
datestamp = config["general"]["run_datestamp"]

full_name = f"{experiment_dir}/run_{coupled_model_name}_{datestamp}"
```

So, could be a change, but strictly not needed.

mandresm:
> There could be some small things to make this more readable, but that is not so important. Only if you want: you often have a pattern such as
> 
> ```python
> f'{config["general"]["experiment_dir"]}/run_'
>          f'{config["general"]["iterative_coupled_model"]}'
>          f'{config["general"]["run_datestamp"]}'
> ```
> 
> It might be easier to read to just extract these things beforehand and skip the whole config blah blah blah:
> 
> ```python
> experiment_dir = config["general"]["experiment_dir"]}
> coupled_model_name = config["general"]["iterative_coupled_model"]
> datestamp = config["general"]["run_datestamp"]
> 
> full_name = f"{experiment_dir}/run_{coupled_model_name}_{datestamp}"
> ```
> 
> So, could be a change, but strictly not needed.

Those are everywhere, but you are right, I will change that at least for these new lines, so that we make sure what is new is better coded.

mandresm:
#approve-changes

mandresm:
#bump

PR #849, opened 2022-11-20
fix/esm_master repo errors

mandresm:
# Problems
1. No clear error when using `esm_master get/install-<model>-<version>` and the folder for that model exists.
2. No clear error when using `esm_master get/install-<model>-<version>` and the user introduce incorrectly the user name/password or doesn't have the required permissions.

# Solution
1. If there is a permissions/user identification problem, puts together a summary of those problems and reports it, and adds the `contact` information so that the user knows who to contact about permission problems:
Screenshot 2022-11-20 at 12 54 52
2. If there are no permissions/user identification problems, but the folder already exists, reports a summary of those problems:
Screenshot 2022-11-20 at 12 59 13

# TODO
- [x] @pgierz, @seb-wahl, @JanStreffing, can you try to break this feature (not catching some error on the cloning of the repos) and see if it behaves as you'd like to?
- [x] Add `contact` information in all yamls
- [x] Write docstrings


pgierz:
It's Sunday, go and take a walk while it's not snowing. But also: thank you thank you thank you. This is one part of the tool chain that had driven me crazy so I am very happy for it being more manageable for us on the development side :-)

pgierz:
Awiesm works fine. I'll try again from the dummy account which has no permissions to see what happens.

mandresm:
> Awiesm works fine. I'll try again from the dummy account which has no permissions to see what happens.

Did you try that in the end @pgierz ?

pgierz:
Yup, that worked as well. From my side, this can go over from draft mode to open PR to let the others have a look.

src/esm_master/task.py
```diff
@@ -534,3 +550,181 @@ def output_steps(self):
...
+        """
+        Defines ``self.num_get_commands`` which accounts for the total number of get
+        commands for the current ``esm_master`` operation
+        """
+        self.num_get_commands = 0
+        for command in self.command_list:
+            for subtask in self.subtasks:
+                if command in subtask.package.command_list.get("get", []):
+                    self.num_get_commands += 1
+                    break
```
pgierz, line 566: Should this thing also return something?
```suggestion
    def num_of_get_commands(self):
        """
        Defines ``self.num_get_commands`` which accounts for the total number of get
        commands for the current ``esm_master`` operation
        """
        self.num_get_commands = 0
        for command in self.command_list:
            for subtask in self.subtasks:
                if command in subtask.package.command_list.get("get", []):
                    self.num_get_commands += 1
                    break
        return self.num_get_commands # (Not strictly needed, but might be nice?)
```

mandresm:
Adding contact info to all the models

src/esm_master/task.py
```diff
@@ -534,3 +550,181 @@ def output_steps(self):
...
+
+        Returns
+        -------
+        repo : dict
+            A dictionary containing the following information:
+            - ``package``: the package object that contains all the package info
+            - ``is_repo_operation``: boolean that indicates whether ``command`` is a
+              repo operation
+            - ``is_last_repo``: boolean that indicates whether this is the last repo
+              operation or not
```
pgierz, line 587: Thank you (as always) for nice doc-strings. 👍🏻 🎉

src/esm_master/task.py
```diff
@@ -534,3 +550,181 @@ def output_steps(self):
...
+
+        The ``repo_errors`` attribute itself is a ``dict`` with keys being the failed
+        ``command``s, each of them containing the following variables:
+        - ``model``: model name
+        - ``repo``: url of the repo
+        - ``destination``: destination folder
+        - ``error``: "destination path exist" if the destination exist, and the value
+          of the ``error`` parameter if it the destination does not exist
+
+        Note: because the ``error`` is caught from an ``except`` of a ``subprocess``
```
pgierz: Probably not important, but you can use this:
```
Note
----
```
It might show up in a "more beautiful" way in Sphinx.
```suggestion
        Note
        ---- 
        Since the ``error`` is caught from an ``except`` of a ``subprocess``
```

src/esm_master/task.py
```diff
@@ -534,3 +550,181 @@ def output_steps(self):
...
+            self.repo_errors = {}
+
+        destination = repo["package"].destination
+        # Checks whether the destination path exists to save a "destination path exist"
+        # error
+        if os.path.isdir(destination):
+            error = "destination path exists"
+        for folder in self.folders_after_download:
+            if destination in folder:
+                full_destination = folder
```
pgierz, line 666: I don't entirely understand this part...?

pgierz (changes requested):
Only minor corrections, and I did not understand one part. See the specific comments.

On another note: we should consider blowing this up and re-doing it, as whenever I start looking at `esm-master` it seems to be overly complicated. However, maybe I also just am not deep enough in the code for that part. Perhaps a topic for a hackathon after the holiday break?

src/esm_master/task.py
pgierz: Correction: What happens is clear, but I do not understand why it needs to happen....

src/esm_master/task.py
mandresm: No need to return anything, it's all contained in attributes of the object

src/esm_master/task.py
mandresm: the user tries to install awicm3 but has not permissions for oifs, xios and rnfmap, however, fesom clones because it has no restirctions. The user then gets the persmission error for oifs, xios, rnfmap, contacts Jan or myself and we grant the necessary permisions to the user. Then the user runs the command again and can clone oifs, xios and rnfmap, but git returns an error for fesom because the destination already exist. With this, we can catch this problem and tell the user what to do.

mandresm:
I checked the `esm_tests` differences and the only change is the new `contact` variable. So all good from my side.

mandresm:
#approve-changes

mandresm:
@pgierz gave verbal green light for this :)

mandresm:
#approve-changes

mandresm:
#bump

PR #852, opened 2022-11-21
Docs/esm tests

mandresm:
ESM-Tests documentation. I'll be filling it out slowly, in between meetings...

Here is the link in readthedocs so that people can start using it: https://esm-tools.readthedocs.io/en/docs-esm_tests/esm_tests.html

Please, comment down here if you want a particular feature of `esm_tests` included asap.


pgierz:
I really like this documentation. This is what we should be doing for basically everything 👍🏻 

No bump, no accept new changes, this is just a doc change.

Miguel, you can kill off the branch showing up on readthedocs once everything goes through.

PR #858, opened 2022-11-26
adds an user-friendly error for an incorrect extension of the runscript

### LINKED ISSUE #856: UnboundLocalError: local variable 'user_config' referenced before assignment
**Describe the bug**
```
esm_runscripts runscript -e test -c
```
raises the error
```
UnboundLocalError: local variable 'user_config' referenced before assignment
```
But the same command works when the runscript is called `runscript.yaml` instead, i.e. the file ending is provided.

**Expected behavior**
Successfull `esm_runscripts`-call independent of the fiename of the runscript. 

```
version = esm_tools, version 6.17.1
branch  = release
hash    = 57ef6a69
src     = /home/a/a270073/esm/esm_tools (levante2.atos.local)
```

mandresm:
Closes #856 

Adds an user-friendly error for an incorrect extension of the runscript.

### DIFF
--- a/src/esm_parser/esm_parser.py
+++ b/src/esm_parser/esm_parser.py
@@ -237,6 +237,7 @@ def look_for_file(model, item, all_config=None):
 
 
 def initialize_from_yaml(filepath):
+    user_config = {}
     for file_ending in YAML_AUTO_EXTENSIONS:
         if filepath.endswith(file_ending) and not file_ending == "":
             user_config = yaml_file_to_dict(filepath)
@@ -247,6 +248,13 @@ def initialize_from_yaml(filepath):
             user_config["general"]["runscript_abspath"] = filepath
 
             user_config = complete_config(user_config)
+    if not user_config:
+        user_error(
+            "Incorrect extension",
+            f"The runscript provided (``{filepath}``) does not have a valid file "
+            "extension. The following are valid file extensions for runscripts: "
+            f"{YAML_AUTO_EXTENSIONS[1:]}"
+        )
     return user_config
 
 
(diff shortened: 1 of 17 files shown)


src/esm_parser/esm_parser.py
```diff
@@ -247,6 +248,13 @@ def initialize_from_yaml(filepath):
             user_config["general"]["runscript_abspath"] = filepath
 
             user_config = complete_config(user_config)
+    if not user_config:
+        user_error(
+            "Incorrect extension",
+            f"The runscript provided (``{filepath}``) does not have a valid file "
+            "extension. The following are valid file extensions for runscripts: "
+            f"{YAML_AUTO_EXTENSIONS[1:]}"
```
pgierz, line 256: Why the slice from 1 to the end and not just everything?

pgierz (commented):
One question. Otherwise great! Slow progress towards better errors 🥳

src/esm_parser/esm_parser.py
mandresm: because the first `YAML_AUTO_EXTENSION` is an empty string, and an empty string is not a valid extension. I did not know why `YAML_AUTO_EXTENSION` includes an empty string, but I did not want to dive into that right now. But maybe I should just remove the empty string from there and see what happens?

pgierz: Ah. I remember why it's empty: if the user gives a path that is a valid yaml, it doesn't need to append anything when trying to open the runscript. If not, it goes through yaml, yml, YAML, YML. 

Maybe not the best design choice...

pgierz (approved):

mandresm:
#bump

PR #859, opened 2022-11-28
fixes the default prev_date for the model if no time_step is specified

mandresm:
When a model had no `time_step` variable specified the `prev_date` of the model was set to the `current_date` as default, which is IMO a very bad default. The `prev_date` of general instead is always setup to be one day before the current date. Here, I propose to set the default value of `prev_date` for a model to be the same as `general` (one day before the current date). If the model contains a `time_step` variable, `prev_date` for the model is `current_date` - `time_step` as usual.


mandresm:
#bump

mandresm:
#approve-changes

PR #867, opened 2022-12-05
Feat/albedo wip

pgierz:
@mandresm, here is a cleaned up version of the albedo YAML. Maybe we can also briefly discuss the status so far, and if I need to install anything else via Spack.

### DIFF
--- a/configs/machines/albedo.yaml
+++ b/configs/machines/albedo.yaml
@@ -1,138 +1,149 @@
-# The albedo.awi.de YAML config file
+################################################################################
+# The albedo.awi.de yaml config file for esm-tools
+#
+# P. Gierz, Nov/Dec 2022
+# M. Andres-Martinez, Nov/Dec 2022
+################################################################################
 name: albedo
-# Could be usefully extraced from /etc/os-distro or something
+# Could be useful if this is extraced from /etc/os-distro:
 operating_system: "linux-rocky"
-# Should not be defined here:
-#jobtype: compute NOTE(PG): ...why?? The computer configuration should not need to know anything about the jobtype.
 sh_interpreter: "/usr/bin/bash"
-
-
 # Batch system and configuration:
 batch_system: "slurm"
-accounting: false
+accounting: true
 further_reading:
   - batch_system/slurm.yaml
 partitions:
-        compute:
-                name: mpp
-                cores_per_node: 36
-        pp:
-                name: smp
-                cores_per_node: 1
+  compute:
+    name: mpp
+    cores_per_node: 36
+  pp:
+    name: smp
+    cores_per_node: 1
 
-# (compute) node hardware information
+# compute node hardware information:
 logical_cpus_per_core: 2
 threads_per_core: 1
 
+# pool directory information:
 pool_directories:
-        pool: "/dev/null"
-        projects: "/dev/null"
-        focipool: "/dev/null"
-# PG: Why is this...? Can't we just reference a particular pool from the list each time?
+  pool: "/albedo/pool"
+  projects: "/albedo/work/projects"
+  focipool: "/dev/null"
 pool_dir: "${computer.pool_directories.pool}"
 
-hyper_flag: "" # No hyperthreading on albedo
+hyper_flag: "" # No hyperthreading on albedo is supported in esm-tools!
 choose_heterogeneous_parallelization:
-        True:
-                choose_taskset:
-                    False:
-                        add_export_vars:
-                            I_MPI_SLURM_EXT: 0
-                    "*":
-                        taskset_chs: ""
-                add_unset_vars:
-                        - "SLURM_DISTRIBUTION"
-                        - "SLURM_NTASKS"
-                        - "SLURM_NPROCS"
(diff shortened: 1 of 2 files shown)


configs/machines/albedo.yaml
```diff
@@ -1,138 +1,180 @@
...
 operating_system: "linux-rocky"
-# Should not be defined here:
-#jobtype: compute NOTE(PG): ...why?? The computer configuration should not need to know anything about the jobtype.
 sh_interpreter: "/usr/bin/bash"
-
-
 # Batch system and configuration:
 batch_system: "slurm"
-accounting: false
+accounting: true
```
mandresm, line 13: I did not know Albedo will have accounts for the computing time... good to know.

configs/machines/albedo.yaml
```diff
@@ -1,138 +1,180 @@
...
+            gcc:
+              HDF5ROOT: /albedo/soft/sw/spack-sw/hdf5/1.12.2-rgostku
+            intel-oneapi:
+              HDF5ROOT: /albedo/soft/sw/spack-sw/hdf5/1.12.2-fknnmx5
+        "intel-oneapi-mpi":
+          choose_compiler_suite:
+            gcc: 
+              HDF5ROOT: /albedo/soft/sw/spack-sw/hdf5/1.12.2-lnfmhhc
+            intel-oneapi:
+              HDF5ROOT: /albedo/soft/sw/spack-sw/hdf5/1.12.2-mt53hvh
```
mandresm: Are these choose blocks resolved?

mandresm (changes requested):
Didn't we agree on indentation to be 4 spaces? Other than that, good to go from my side :)

pgierz:
> Didn't we agree on indentation to be 4 spaces? Other than that, good to go from my side :)

Yeah, I have not correctly configured my vim yet on albedo :-(

configs/machines/albedo.yaml
pgierz: Well, I just made a compile script with that version, so it seems to work. I'd rather have those variables set when you load the modules.

configs/machines/albedo.yaml
pgierz: Just for reporting purposes; basically we want to have some way to reliably figure out how much compute time is being used by which group.

mandresm:
Are now gonna be using the gcc versions? I thought you said it was going to be the intel-oneapi compiler ones

mandresm:
Just checked, the `choose_` blocks inside the `export_vars` are not resolved. I think ESM-Tools only resolves choose blocks when they are at the base of a section (1st level of indentation). We should probably resolve the HDFROOT stuff outside into a variable. Do you want me to take care of that?

configs/machines/albedo.yaml
mandresm: so there you would just write your awi user right? not your project?

configs/machines/albedo.yaml
mandresm: they don't work, see comment below

pgierz: Sorry....monday morning. I was using an old compile script 🤦🏻‍♂️

pgierz:
> Just checked, the `choose_` blocks inside the `export_vars` are not resolved. I think ESM-Tools only resolves choose blocks when they are at the base of a section (1st level of indentation). We should probably resolve the HDFROOT stuff outside into a variable. Do you want me to take care of that?

yes please

configs/machines/albedo.yaml
```diff
@@ -123,58 +124,17 @@ export_vars:
...
 
-  # NetCDF environment Variables:
+  # HDF5 exported environment variables:
+  HDF5ROOT: "$(module show hdf5/${hdf5_version}-${mpi_implementation_suffix}-${compiler_suffix} | CMAKE_PREFIX_PATH | cut -d " " -f 2)"
+  HDF5_C_INCLUDE_DIRECTORIES: $HDF5ROOT/include
+  HDF5_ROOT: $HDF5ROOT
+
+  # NetCDF environment variables:
+  NETCDFFROOT: "$(module show netcdf-fortran/${netcdf_fortran_version}-${mpi_implementation_suffix}-${compiler_suffix} | CMAKE_PREFIX_PATH | cut -d " " -f 2)"
+  NETCDFROOT: "$(module show netcdf-c/${netcdf_c_version}-${mpi_implementation_suffix}-${compiler_suffix} | CMAKE_PREFIX_PATH | cut -d " " -f 2)"
```
pgierz: Still needs a `grep`?

mandresm: yep, commitng now

pgierz:
I also fixed the idiotic broken Python 3.6 test here. Seems we were using an ancient version of the setup Python action, plus Python 3.6 had end-of-support beginning of 2022.

pgierz:
You're kidding me. I forgot a second F?!

mandresm:
That doesn't fix our netcdf problems though :(

mandresm (approved):

mandresm:
I'm merging this one now, so that we can resume work from the `feat/albedo` branch

PR #886, opened 2023-01-09
fix/esm_tests broken namelist links

mandresm:
fixes problems with namelists in the experiments being broken links because the file has been moved to the general experiment config


mandresm:
#bump

PR #896, opened 2023-02-02
better error message in case of missing model folder

JanStreffing:


JanStreffing:
Some changes on blogin from foci. Maybe the test script needs updating of the correct state here?

pgierz (approved):

pgierz:
I use neither foci nor the blogin computer, so I can't comment on that ;-)

mandresm:
There is a silly issue with `esm_tests` affecting exclusively to FOCI. So you can ignore the error here. I'll be fixing that issue later today or next week.

mandresm:
#bump

PR #915, opened 2023-03-14
Feat/albedo

mandresm:
Merges current state of Albedo support


configs/setups/awiesm/awiesm.yaml
```diff
@@ -183,7 +183,12 @@ echam:
                 OIFS_OASIS_INCLUDE: '"-I$OIFS_OASIS_BASE/build/lib/psmile -I$OIFS_OASIS_BASE/build/lib/psmile/scrip -I$OIFS_OASIS_BASE/build/lib/psmile/mct -I$OIFS_OASIS_BASE/build/lib/psmile/mct/mpeu"'
                 OASIS3MCT_FC_LIB: "$(pwd)/lib/"
                 OASIS3MCTROOT: "$(pwd)/oasis/"
-                configure_opts: '--with-coupler=oasis3-mct'
+                configure_opts: "'--with-coupler=oasis3-mct --without-regard-for-quality'"
+            # Unset FFLAGS, they are set in mh-linux in echam (booo this is horrible)
```
mandresm: @Paul XD made me laugh really hard. I'm okay with leaving it like this, I just don't understand why we need to set them on on the echam file and here off.

mandresm (commented):
All good from my side. Only one question in the echam.yaml.

src/esm_environment/esm_environment.py
```diff
@@ -541,7 +549,7 @@ def write_dummy_script(self, include_set_e=True):
         with open("dummy_script.sh", "w") as script_file:
             # Write the file headings
             script_file.write(
-                f'#!{self.config.get("sh_interpreter", "/bin/bash")} -l\n'
+                f'#!{self.config.get("sh_interpreter", "/bin/bash")}\n'
```
mandresm: I catch this one with `esm_tests`: This flag "makes bash to act as it has being invoked as a login shell". This means that the comp scripts and run scripts are shielded against whatever exports the user had made on their current session, ensuring that the ESM-Tools environment defined on those scripts prevail (only the environment defined by the user in the bash_profile, bash.rc... is loaded). I am afraid that this is the type of change that could result in weird errors difficult to reproduce. I would rather have the `-l` back unles there are very good reasons that I'm not aware of to remove it. What do you think @pgierz?

pgierz (approved):

configs/setups/awiesm/awiesm.yaml
mandresm: @pgierz, I'm concern that this change will affect other computers: https://github.com/esm-tools/esm_tools/actions/runs/4415382184/jobs/7738274369#step:8:205

Both the `--without-regard-for-quality` and the `unset` of variables. Should I make this happen exclusively for Albedo or do you want that these changes happen in all machines?

pgierz: Hmmm. I think only for albedo for now. Since the other machines work fine, I would avoid tampering with it.

mandresm:
All changes in `esm_tests` are expected and do not have an effect on namelists, `.run` scripts or `comp` scripts in computers different from Albedo, therefore #approve-changes

mandresm:
#bump

PR #919, opened 2023-03-17
Fix pytest

mandresm:
Fixes a wild card typo and the pytest


pgierz (approved):
This should be a private method (Python does not really have private, so we normally use a prefix underscore) This fixes our breaking tests, right?

mandresm:
> This fixes our breaking tests, right?

yes, it does

PR #923, opened 2023-03-22
is date in interval method

mandresm:


mandresm:
The MOTD fix was so that ESM-Tools doesn't crash because it has not connection to internet (e.g. on the plane)

src/esm_calendar/esm_calendar.py
```diff
@@ -895,3 +895,38 @@ def sub_tuple(self, to_sub):
...
+                f"The time interval ${interval} is not well defined for the separator "
+                f"${sep}"
+            )
+        interval = [Date(date) for date in interval]
+
+        # Check whether the ``self`` date falls within the interval
+        if self > interval[0] and self < interval[1]:
+            return True
+        else:
+            return False
```
pgierz: Keep it simple: 
```suggestion
        # Check whether the ``self`` date falls within the interval
        return self > interval[0] and self < interval[1]
```

src/esm_calendar/esm_calendar.py
```diff
@@ -895,3 +895,38 @@ def sub_tuple(self, to_sub):
...
+        """
+        Finds if a date falls in an open interval.
+
+
+        Parameters
+        ----------
+        interval : str
+            A date interval string in which the 2 dates are separated by a ``sep``
+            string.
+        sep : str
```
pgierz: Just pass a tuple directly:


```suggestion
        interval : tuple of Date
            A date interval in which there are 2 Dates objects.
```

src/esm_calendar/esm_calendar.py
```diff
@@ -895,3 +895,38 @@ def sub_tuple(self, to_sub):
 
         new_date = self.from_list(result)
         return new_date
+
+    def is_in_interval(self, interval, sep="->") -> bool:
```
pgierz:
```suggestion
    def is_in_interval(self, interval) -> bool:
```

You probably can make a type annotation of `interval: Tuple[Date, Date]`, but I can't remember the exact syntax

src/esm_calendar/esm_calendar.py
```diff
@@ -895,3 +895,38 @@ def sub_tuple(self, to_sub):
...
+        Returns
+        -------
+        True :
+            If the date of ``self`` is within the open interval defined by ``interval``
+        False :
+            If the date of ``self`` is out of the open interval defined by ``interval``            
+        """
+
+        # Split the interval into 2 dates
+        interval = interval.split(sep)
```
pgierz: Can be removed if you want to go with passing in a tuple directly

```suggestion
```

src/esm_calendar/esm_calendar.py
```diff
@@ -895,3 +895,38 @@ def sub_tuple(self, to_sub):
...
+        if len(interval) != 2:
+            raise TypeError(
+                f"The time interval ${interval} is not well defined for the separator "
+                f"${sep}"
+            )
```
pgierz:
```suggestion
            raise TypeError(f"You must have {interval} be a tuple of two Dates!")         
```

tests/test_esm_calendar/test_esm_calendar.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+
+import pytest
+
+import esm_calendar
+
+def test_date_is_in_interval():
+    date1 = esm_calendar.Date("1949-12-31")
+    date2 = esm_calendar.Date("1950-01-01")
+    interval = "1850-01-01->1950-01-01"
```
pgierz:
```suggestion
    interval = ("1850-01-01", "1950-01-01")
```

src/esm_calendar/esm_calendar.py
```diff
@@ -895,3 +895,38 @@ def sub_tuple(self, to_sub):
...
+        interval = [Date(date) for date in interval]
```
pgierz: This line might need to get checked. It currently *needs* `date` to be a string. It might be that `date` is already a `Date`.

(Re-reading that made my head hurt)

pgierz (commented):
I have some suggestions, feel free to use them or not, however you want :-)

src/esm_calendar/esm_calendar.py
mandresm: I'm out of practice...

src/esm_calendar/esm_calendar.py
mandresm: thanks for the review, I'll take care of it later. I gave the interval as a string with a separator, because that's how it will be the same as the key in the `file` object. But I agree with you: let's keep this as independent as possible from `filedicts` and make the string transformations there

src/esm_calendar/esm_calendar.py
```diff
@@ -926,7 +926,4 @@ def is_in_interval(self, interval, sep="->") -> bool:
         interval = [Date(date) for date in interval]
 
         # Check whether the ``self`` date falls within the interval
-        if self > interval[0] and self < interval[1]:
-            return True
-        else:
-            return False
+        return self > interval[0] and self < interval[1]
```
pgierz:
```suggestion
        return interval[0] < self < interval[1]
```

src/esm_calendar/esm_calendar.py
mandresm: I removed this, it was not necessary

mandresm:
Hi @pgierz, I apply the changes you suggested and adjust the test, not because we are on a sprint, but just so that this is at least finished for the next one. I think it's ready to be merged.

pgierz:
OK, I will push the button ;)

PR #931, opened 2023-03-28
fix/monorepo fixing script

mandresm:
The transition between the multirepo version 5.x to the monorepo version 6.x was smoothly done if one would use `esm_versions upgrade`. That triggered a python script in `esm_versions` that would clean all the multirepos and other metadata to avoid conflicts with the monorepo.

It is quite often the case, that users did never upgrade to version 6 and suddenly they decide to do it, but they just do a git checkout and `./install.sh`. This generates all sorts of weird behaviour as it means the multirepo and monorepo are installed and conflicting. To solve that I placed the cleanup python script inside `utils` so that users can easily clean their deprecated ESM-Tools package by simply running `python3 utils/monorepo_fix.py`.

This should solve the problems reported by @hegish in https://github.com/AWI-ESM/project_management/issues/131#issuecomment-1485137400

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -46,7 +46,7 @@ if hash git 2>/dev/null; then
 		echo $git_error_message
 		echo "git version found: ${git_version}"
 	else
-		if test ${minor_git_version} -lt "13"; then
+		if test ${minor_git_version} -lt "10"; then
 			echo $git_error_message
 			echo "git version found: ${git_version}"
 		fi
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.21.1
+current_version = 6.21.2
 commit = True
 tag = True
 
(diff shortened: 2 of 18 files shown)


mandresm:
I have managed to reproduced @hegish's issue and tested that the python script fixes it.

pgierz (approved):
If this works, sure. There are many things about it that make me want to bang my head against the wall, but more important that we get people on newer versions than making the script beautiful...

mandresm:
This script was reviewed in 2021 by a few people and widely used in the transition to version 6... it's horrible? yes, but gets rid of an even more horrible problem: the old multirepo :)

JanStreffing:
I've used this a few times. Good to have it done automatically. There might not be many old installations out there, but the longer it has been, the harder it will be to spot the strange way the normal install breaks. So yeah, approved.

JanStreffing (approved):

mandresm:
#bump

pgierz:
I'm still surprised we got away with someone (by the tone, either me or Miguel) writing "this function does all the magic". Hooray you wrote a doc-string. Except no, no you did not. Not even slightly.

PR #942, opened 2023-05-03
Sprint/filedicts/class constructor

pgierz:
There are far more commits in this PR than I know what to do with.

The last one is important. You can know build a simulation file in a few different ways:

```python
>>> from esm_runscripts import filedicts
>>> paths = {"computer": "/some/nice/location", "exp_tree": "/some/other/nice/location"
>>> name = "My Beautiful File"
>>> sim_file = filedicts.SimulationFile(paths=paths, name=name)
```

Alternatively, you can just give the config and location in the config:
```python
>>> from esm_runscripts import filedicts
>>> sim_file = filedicts.SimulationFile.from_config(path_to_config, "echam.files.jan_surf")
```

There is also introduced a `DatedSimulationFile`, which knows about its date. 

There may be some duplication of mini fixes also in #922 (grammar in the comments) which I did not bother to filter out nicely.


pgierz:
#approve-changes

PR #949, opened 2023-05-10
947 comments for generic maschine file

### LINKED ISSUE #947: Comments for generic maschine file
As mentioned in our Monday meeting, it would be good to have some comments for the generic machine file:
https://github.com/esm-tools/esm_tools/blob/release/configs/machines/generic.yaml

So that it can serve as a more self explanatory base for users trying to install models on machines we have no access to.

nwieters:
closes #947

### DIFF
--- a/configs/machines/generic.yaml
+++ b/configs/machines/generic.yaml
@@ -1,6 +1,13 @@
+################################################################################
+# The albedo.awi.de yaml config file for esm-tools
+#
+# N. Wieters, May 2023: Added comments and madatory entries
+################################################################################
+
 # GENERIC YAML CONFIGURATION FILES
 
 name: generic
+sh_interpreter: "/usr/bin/bash"
 accounting: false
 hetjob_flag: "?" # either hetjob or packjob
 batch_system: "slurm"
--- a/configs/machines/generic.yaml
+++ b/configs/machines/generic.yaml
@@ -48,8 +48,6 @@ use_hyperthreading: False
 #            mt_launcher_flag: ""
 #        False: 
 #            <set_any_variable_here_related_to_hypertherading>: <value>
-#        "anyother_scalar_value':
-#            <set_any_variable_here_related_to_hypertherading>: <value>
 
 # Set default partition ??? (e.g. compute)
 partition: <partition_name>
--- a/configs/machines/generic.yaml
+++ b/configs/machines/generic.yaml
@@ -1,51 +1,158 @@
-################################################################################
-# The albedo.awi.de yaml config file for esm-tools
+#######################################################################################
+# The dummy yaml machine config file for esm-tools
 #
-# N. Wieters, May 2023: Added comments and madatory entries
-################################################################################
+# This is a dummy machine file. It is intend to be used for creating a new config file 
+# for a new machine setup. It provides a minimal structure to be filled in with
+# the oppropriate values of the new HPC system. 
+#
+# Usage:  
+#     - Please replace the placeholder indicated by <...> 
+#       with your PHC specific values
+#     - Please uncomment sections of this yaml file if appropriate.
+#       Possible sections are indicating this (see below).
+#     - See also files for other machines that are already implemented in esm-tools.
+#     - After editing this file, save it as <machine>.yaml. Add also an entry in
+#       the file all_machines.yaml (in this folder) for the new machine.
+#
+# Last changes: N. Wieters, May 2023: Added comments and placeholders.
+#######################################################################################
 
 # GENERIC YAML CONFIGURATION FILES
+#
+# Set hostname of the machine 
+name: <hostname>
 
-name: generic
-sh_interpreter: "/usr/bin/bash"
-accounting: false
-hetjob_flag: "?" # either hetjob or packjob
-batch_system: "slurm"
+# General information about operating system
+# Operating system
+#operating_system: "<your-operating-system>"
+# Shell interpreter (e.g. "/usr/bin/bash")
+sh_interpreter: "<path-to-shell>"
 
-jobtype: compute
+# Information about the job scheduler
+# Set batch system (e.g. slurm)
+batch_system: "<your-job-scheduler>"
 
-partition: generic 
+# Set whether an account needs to be set for batch system (e.g. true or false). The actual account that will be used for the experiment will be set in the runscript.
+accounting: true 
+#
+#hetjob_flag: "?" # either hetjob or packjob
+#
+#jobtype: compute
+#
+# Information about the partitions of the machine
+#
+# Specify if hyperthreading should be used.
+use_hyperthreading: False
+# Set further options depending on hyperthreading variable 'use_hyperthreading' by uncommenting and editing the following 'choose_use_hyperthreading' section:
+#choose_use_hyperthreading:
+#        True: 
+#            hyperthreading_flag: ""
+#            mt_launcher_flag: ""
+#        False: 
+#            <set_any_variable_here_related_to_hypertherading>: <value>
+#        "anyother_scalar_value':
+#            <set_any_variable_here_related_to_hypertherading>: <value>
 
+# Set default partition ??? (e.g. compute)
+partition: <partition_name>
+#
+# Set detail information for available partitions on the machine.
+# These variables will be used in other parts of this yaml file.
 choose_partition:
-        generic:
-                partition_name: generic
-                partition_cpn: 8 
-
+        <partition_name>:
+                partition_name: "<partition_name>"
+                partition_cpn: <number_of_cores_per_node>  # integer number 
+#
+# Define available partitions of the machine for different jobtypes (e.g. for computing [compute], post-processing [PP], etc.)
+#
 partitions:
         compute:
-                name: ${computer.partition_name}
+                name: ${computer.partition_name}     # this uses the variable set in choose_partition section (see above)
+                #name: "compute"                     # or set the partition name as string
                 cores_per_node: ${computer.partition_cpn}
-
-logical_cpus_per_core: 2
-
-threads_per_core: 1
-
+#
+#logical_cpus_per_core: 2
+#
+#threads_per_core: 1
+#
+# Specify different pool directories for this machine.
+#
 pool_directories:
-        focipool: "/path/not/set/for/generic/yaml/"
-        pool: "/path/not/set/for/generic/yaml"
+#        focipool: "<set_path_to_pool_dir>"
+#        pool: "<set_path_to_pool_dir"
+#
+# Is this a dublicate ??? Can be set in runscript.
+pool_dir:  "<set_path_to_pool_dir>"
+#
+# Specify compiler (e.g. intel2022_openmpi)
+compiler_mpi: <compiler>
 
-pool_dir:  "/path/not/set/for/generic/yaml/"
+# Choose libraries. Define further libraries (depending on iolibraries value set in the next line) below in section choose_iolibraries
+iolibraries: system_libs
 
-# standard environment setup
+# Setup environment
+# Load necessary modules (e.g. git, cdo, nco). Edit and extent the following lines.
 module_actions:
-   - "load something"
+        - "purge"
+        - "load <module>"
+#        - "load <module>"
+#        - "load <module>"
+#
+# Export environment variables. Uncomment and extend the following lines to export needed environment variables.
+#export_vars:
+#        # Locale Settings
+#        LC_ALL: en_US.UTF-8
+#        <variables>: <value>
+#
+# Variables needed exclusively during runtime
+# NOTE: runtime_environment_changes does not work from the machine files
+#choose_general.run_or_compile:
+#    runtime:
+#        add_export_vars:
+#            var1: ...
+#
+# Set information depending on choosen value for variable compiler_mpi set above. Uncomment and extend the following lines.
+#choose_compiler_mpi:
+#       <compiler>:
+#           <variable>: <value>
+#           # load additional modules specific for compiler
+#           add_module_actions:
+#               - "load <module1>"
+#               - "load <module2>"
+#               - "load <module3>"
+#                 # add additional variables to exported: change and extend following lines:
+#           add_export_vars:
+#               FC: <mpif90>
+#               F77: <mpif90>
+#               MPICC: <mpicc>
+#               MPIFC: <mpif90>
+#               CC: <mpicc>
+#               CXX: <mpicxx>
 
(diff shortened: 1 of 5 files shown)


JanStreffing (approved):
I think this is a good improvement. We can keep adding to it as users of the template have questions and receive answers.

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
+# Usage:  
+#     - Please replace the placeholder indicated by <...> 
+#       with your PHC specific values
+#     - Please uncomment sections of this yaml file if appropriate.
+#       Possible sections are indicating this (see below).
+#     - See also files for other machines that are already implemented in esm-tools.
+#     - After editing this file, save it as <machine>.yaml. Add also an entry in
+#       the file all_machines.yaml (in this folder) for the new machine.
+#
+# Last changes: N. Wieters, May 2023: Added comments and placeholders.
```
pgierz:
```suggestion
```

We don't need this, that is what `git blame` does for you 🙂

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
+# Specify different pool directories for this machine.
+#
 pool_directories:
-        focipool: "/path/not/set/for/generic/yaml/"
-        pool: "/path/not/set/for/generic/yaml"
+#        focipool: "<set_path_to_pool_dir>"
+#        pool: "<set_path_to_pool_dir"
+#
+# Is this a dublicate ??? Can be set in runscript.
+pool_dir:  "<set_path_to_pool_dir>"
```
pgierz: No, that's not a duplicate. Setting it in the runscript would override the one that is set for you by default here.

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
-jobtype: compute
+# Information about the job scheduler
+# Set batch system (e.g. slurm)
+batch_system: "<your-job-scheduler>"
 
-partition: generic 
+# Set whether an account needs to be set for batch system (e.g. true or false). The actual account that will be used for the experiment will be set in the runscript.
+accounting: true 
+#
+#hetjob_flag: "?" # either hetjob or packjob
```
pgierz:
```suggestion
#hetjob_flag: "?" # either hetjob or packjob, this depends on your SLURM version. More info can be found: https://slurm.schedmd.com/heterogeneous_jobs.html
```

pgierz (approved):
Looks good! I had a few minor suggestions.

For info, if you are in the role of reviewer, you can add little suggestions by highlighting the lines, making a comment, and then the very fir button on the right hand side (after the write and preview tabs) is a little file with plus/minus. That automatically pastes the selected code and you can make modifications.

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
+use_hyperthreading: False
+# Set further options depending on hyperthreading variable 'use_hyperthreading' by uncommenting and editing the following 'choose_use_hyperthreading' section:
+#choose_use_hyperthreading:
+#        True: 
+#            hyperthreading_flag: ""
+#            mt_launcher_flag: ""
+#        False: 
+#            <set_any_variable_here_related_to_hypertherading>: <value>
+#        "anyother_scalar_value':
+#            <set_any_variable_here_related_to_hypertherading>: <value>
```
mandresm: In this instance, I'd leave the "1" and "0" as in Levante. That was only to prevent that someone defined`use_hyperthreading` as a 0 string and it will still be evaluated as true.

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
 
+# Set default partition ??? (e.g. compute)
```
mandresm:
```suggestion
# Set default partition (e.g. compute)
```
Yes, default partition

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
-        focipool: "/path/not/set/for/generic/yaml/"
-        pool: "/path/not/set/for/generic/yaml"
+#        focipool: "<set_path_to_pool_dir>"
+#        pool: "<set_path_to_pool_dir"
+#
+# Is this a dublicate ??? Can be set in runscript.
+pool_dir:  "<set_path_to_pool_dir>"
+#
+# Specify compiler (e.g. intel2022_openmpi)
+compiler_mpi: <compiler>
```
mandresm: I would remove this, it's not a feature variable, and it is only used as a case for adding variables to `module_actions` and `export_vars`, IMO it's up to the person who's writing the machine file to add the logic that they want for the switches (`choose_` blocks).

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
 
-pool_dir:  "/path/not/set/for/generic/yaml/"
+# Choose libraries. Define further libraries (depending on iolibraries value set in the next line) below in section choose_iolibraries
+iolibraries: system_libs
```
mandresm: Same as above, I recommend removing it.

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
+        - "purge"
+        - "load <module>"
+#        - "load <module>"
+#        - "load <module>"
+#
+# Export environment variables. Uncomment and extend the following lines to export needed environment variables.
+#export_vars:
+#        # Locale Settings
+#        LC_ALL: en_US.UTF-8
+#        <variables>: <value>
```
mandresm:
```suggestion
export_vars:
        # Locale Settings
        LC_ALL: en_US.UTF-8
        <variables>: <value>
```

`LC_ALL` should always be defined to avoid problems with python and the locale

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
-cxx: "unset"
+#
+## some yamls use computer.fc, etc to identify the compiler, so we need to add them
+fc: "$FC"
+cc: "$CC"
+mpifc: "$MPIFC"
+mpicc: "$MPICC"
+cxx: "$CXX"
+#
+#
```
mandresm: IMO is up to the person writing the machine file the logic they want to use for the compilers and MPI switches. Therefore, for the sake of simplicity, I recommend removing all this.

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
+#        # Locale Settings
+#        LC_ALL: en_US.UTF-8
+#        <variables>: <value>
+#
+# Variables needed exclusively during runtime
+# NOTE: runtime_environment_changes does not work from the machine files
+#choose_general.run_or_compile:
+#    runtime:
+#        add_export_vars:
+#            var1: ...
```
mandresm: I would remove this also for simplicity. In the end this is also a switch.

configs/machines/generic.yaml
nwieters: Ok. Thanks for clarification. I will remove my comment.

configs/machines/generic.yaml
nwieters: Thanks Miguel, I will remove the option for a scalar value.

configs/machines/generic.yaml
```diff
@@ -1,44 +1,158 @@
...
+# Specify if hyperthreading should be used.
+use_hyperthreading: False
+# Set further options depending on hyperthreading variable 'use_hyperthreading' by uncommenting and editing the following 'choose_use_hyperthreading' section:
+#choose_use_hyperthreading:
+#        True: 
+#            hyperthreading_flag: ""
+#            mt_launcher_flag: ""
+#        False: 
+#            <set_any_variable_here_related_to_hypertherading>: <value>
+#        "anyother_scalar_value':
```
nwieters:
```suggestion
```

configs/machines/generic.yaml
```diff
@@ -1,44 +1,156 @@
...
+#            <set_any_variable_here_related_to_hypertherading>: <value>
```
nwieters:
```suggestion
```

mandresm (changes requested):
Thanks for this contribution! Apart from the suggestions I made referring to code lines I have 3 extra requests:
- [x] We've agreed some time ago to go for a 4 spacing indentation at least for the new yaml files. Could you please modify this file accordingly?
- [x] The `configs/machines/generic.yaml` was set to be able to at least be able to run `esm_runscripts` with `-c` option in any computer. With the current changes, this option does not work anymore. What I suggest is that we have 2 files, the old `config/machines/generic.yaml` for the purpose of having a generic yaml file that allows for minimal testing of ESM-Tools functionality in other computers without having to edit any yaml file, and then that what this file as it is right now is copied to `configs/templates/machine_template.yaml`.
- [x] Apply the necessary changes in the following parts of the code, so that `esm_tools create-new-config machine` copies the `configs/templates/machine_template.yaml` and opens it to start editing.
  https://github.com/esm-tools/esm_tools/blob/release/src/esm_tools/cli.py#L36

mandresm:
> [ ] The `configs/machines/generic.yaml` was set to be able to at least be able to run `esm_runscripts` with `-c` option in any computer. With the current changes, this option does not work anymore. What I suggest is that we have 2 files, the old `config/machines/generic.yaml` for the purpose of having a generic yaml file that allows for minimal testing of ESM-Tools functionality in other computers without having to edit any yaml file, and then that what this file as it is right now is copied to `configs/templates/machine_template.yaml`.

Just to clarify this point, I would do the following for simplicity:
1. `cp config/machines/generic.yaml configs/templates/machine_template.yaml`
2. `git checkout release config/machines/generic.yaml` -> restore the file taking the one from `release` branch

pgierz:
> * We've agreed some time ago to go for a 4 spacing indentation at least for the new yaml files. Could you please modify this file accordingly?

I can give a demo on something called [`editorconfig`](https://editorconfig.org) which plugs into your editor of choice and takes care of those sorts of things for you.

configs/machines/generic.yaml
```diff
@@ -1,44 +1,154 @@
...
+mpifc: "$MPIFC"
+mpicc: "$MPICC"
+cxx: "$CXX"
+#
+#
+launcher_flags: "-l" # --distribution=block:cyclic"
 
+# Choose another configuration file, that should be evaluated (e.g. slurm.yaml)
 further_reading:
         - batch_system/slurm.yaml
```
mandresm: We also support PBS (in Aleph for example). So this line could also be `- batch_system/pbs.yaml`

nwieters: Added pbs.yaml in the comment.

nwieters:
Sorry, I saw some of your comments too late. I will put them in now, and push again.
I think I need a little bit more practise with github.

configs/machines/generic.yaml
nwieters: Removed comment sign.

configs/machines/generic.yaml
nwieters: Removed ???

configs/machines/generic.yaml
nwieters: I see the point to keep it as simple as possible. But on the other hand, this way of setting up the machine file is used in many other machine files. Therefore I think it could help someone to have an example of how such a file could be setup up, not the way it should or needs to be.

nwieters: I reduced this to only one example at the end of the template. I kept this in as an example of how to write and use a choose_ block in the machine file.

configs/machines/generic.yaml
nwieters: Removed this.

configs/machines/generic.yaml
nwieters: Removed

configs/templates/machine_template.yaml
```diff
@@ -0,0 +1,153 @@
...
+# the oppropriate values of the new HPC system. 
+#
+# Usage:  
+#     - Please replace the placeholder indicated by <...> 
+#       with your PHC specific values
+#     - Please uncomment sections of this yaml file if appropriate.
+#       Possible sections are indicating this (see below).
+#     - See also files for other machines that are already implemented in esm-tools.
+#     - After editing this file, save it as <machine>.yaml. Add also an entry in
+#       the file all_machines.yaml (in this folder) for the new machine.
```
mandresm:
```suggestion
#     - After editing this file, save it as <machine>.yaml in configs/machines. Add also
#       an entry in the file configs/machines/all_machines.yaml (in this folder) for the
#       new machine.
```

configs/templates/machine_template.yaml
```diff
@@ -0,0 +1,153 @@
...
+name: <hostname>
+
+# General information about operating system
+# Operating system
+#operating_system: "<your-operating-system>"
+# Shell interpreter (e.g. "/usr/bin/bash")
+sh_interpreter: "<path-to-shell>"
+
+# Information about the job scheduler
+# Set batch system (e.g. slurm)
```
mandresm:
```suggestion
# Set batch system (e.g. slurm/pbs)
```

configs/templates/machine_template.yaml
```diff
@@ -0,0 +1,153 @@
...
+
+## Some other yaml files use a file 'computer.fc', etc to identify the compiler, so we need to add them here.
+fc: "$FC"
+cc: "$CC"
+mpifc: "$MPIFC"
+mpicc: "$MPICC"
+cxx: "$CXX"
+#
+#
+launcher_flags: "-l" # --distribution=block:cyclic"
```
mandresm:
```suggestion
# Launcher flags (i.e. flags for the srun/mpirun/aprun command)
launcher_flags: "-l"
```
Distribution is too specific to the `srun` launcher, therefore I recommend to remove it. We support also `mpirun` launcher and `aprun` (Aleph).

configs/templates/machine_template.yaml
```diff
@@ -0,0 +1,153 @@
...
+
+
+####################################################################################
+# Further functionality: 
+# 
+# If there are modules or environment variables that are different for 
+# different e.g. compilers, you can use the funcionality of 
+# declaring choose_ blocks in this yaml file.
+# (For more details about yaml syntax please also see the documentation
+# https://esm-tools.readthedocs.io/en/latest/yaml.html)
```
mandresm:
```suggestion
# https://esm-tools.readthedocs.io/en/latest/yaml.html#switches-choose)
```

mandresm (approved):
I made some changes in the recipe and committed them. I have also made a couple of extra minimal suggestions. Other than that, all thumbs up after the suggestions are dealt with.

PR #957, opened 2023-05-19
Fix/echam albedo

mandresm:
Reported by @l975421700


mandresm:
I won't be waiting for a review since it's kind of urgent, it's in Albedo so there is not much that can be broken, and this was a silly mistake I just introduced on the previous  hotfix...

mandresm:
#approve-changes

mandresm:
#bump

PR #964, opened 2023-05-23
trigger an error if using esm_runscripts with generic.yaml without ch…

### LINKED ISSUE #959: not-detected (=generic) hostname should stop immediately
**Describe the bug**
If the esm_tools do not detect the hostname at the beginning of a job to model date1, the esm_parser shows a warning like this:
```
WARNING:root:The yaml file for this computer (l40248) could not be determined!
WARNING:root:Continuing with generic settings...
```
and continues the job. At the end of the job, if a subsequent job is about the be submitted to model date2, the same warning is triggered again and, in addition, since the srun command is crap due to the generic setting, srun has this error:
```
sbatch.bin: error: This does not look like a batch script.  The first
sbatch.bin: error: line must start with #! followed by the path to an interpreter.
sbatch.bin: error: For instance: #!/bin/sh
```
i.e., the job to model date2 is not submitted. After this the esm_runscripts process continues so that the job to model date1 ends successfully.

Does it? How can I know that the job to model date1 ended successfully? Maybe its flawed due to the generic setting? Maybe a default forcing file was used due to the generic setting so that the model run successfully finished but with a forcing file that I did not want to use. 

**Expected behavior**
I think the esm_parser should not raise a warning but an error at the beginning of the job to model date1. Ideally in a way so that 
```
computer:
    mail_type: "FAIL"
    mail_user: "<mymail>
```
is triggered.

Thanks and cheers,
Chris

mandresm:
Triggers an error if using esm_runscripts with generic.yaml without check mode off to prevent ESM-Tools not catching problems in the recognition of the machine file during runtime, as described in #959 

Closes #959 

One can replicate this error by running `esm_runscripts` in your local computer without `-c`.

### DIFF
--- a/configs/machines/generic.yaml
+++ b/configs/machines/generic.yaml
@@ -54,7 +54,7 @@ choose_general.check:
                     is only aimed for local testing of ESM-Tools functionalities and dry
                     check runs (``esm_runscripts -c``).\n
 
-                    If you are running in a HPC and you're seeing this message, it's
+                    If you are running on an HPC and you're seeing this message, it's
                     likely that the HOSTNAME of your machine does not matches the
                     pattern defined in ``configs/machines/all_machines.yaml``. Please,
                     check that your machine is included in that file and that the
--- a/configs/machines/generic.yaml
+++ b/configs/machines/generic.yaml
@@ -42,3 +42,24 @@ cxx: "unset"
 
 further_reading:
         - batch_system/slurm.yaml
+
+choose_general.check:
+    False:
+        error:
+            generic.yaml:
+                message: "
+                    You are trying to run an actual simulation with the environment from
+                    ``configs/machines/generic.yaml``. This is not allowed, you should
+                    always run using a specific machine file. The ``generic.yaml`` file
+                    is only aimed for local testing of ESM-Tools functionalities and dry
+                    check runs (``esm_runscripts -c``).\n
+
+                    If you are running in a HPC and you're seeing this message, it's
+                    likely that the HOSTNAME of your machine does not matches the
+                    pattern defined in ``configs/machines/all_machines.yaml``. Please,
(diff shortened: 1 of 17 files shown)


pgierz (approved):

mandresm:
#bump

configs/machines/generic.yaml
```diff
@@ -42,3 +42,24 @@ cxx: "unset"
...
+            generic.yaml:
+                message: "
+                    You are trying to run an actual simulation with the environment from
+                    ``configs/machines/generic.yaml``. This is not allowed, you should
+                    always run using a specific machine file. The ``generic.yaml`` file
+                    is only aimed for local testing of ESM-Tools functionalities and dry
+                    check runs (``esm_runscripts -c``).\n
+
+                    If you are running in a HPC and you're seeing this message, it's
+                    likely that the HOSTNAME of your machine does not matches the
```
nwieters:
```suggestion
                    likely that the HOSTNAME of your machine does not match the
```

configs/machines/generic.yaml
```diff
@@ -42,3 +42,24 @@ cxx: "unset"
...
+        error:
+            generic.yaml:
+                message: "
+                    You are trying to run an actual simulation with the environment from
+                    ``configs/machines/generic.yaml``. This is not allowed, you should
+                    always run using a specific machine file. The ``generic.yaml`` file
+                    is only aimed for local testing of ESM-Tools functionalities and dry
+                    check runs (``esm_runscripts -c``).\n
+
+                    If you are running in a HPC and you're seeing this message, it's
```
nwieters:
```suggestion
                    If you are running on an HPC and you're seeing this message, it's
```

mandresm:
#bump

PR #967, opened 2023-05-24
901 oasis restarts issue

### LINKED ISSUE #901: Oasis restarts issue
Dear developers,
ESM_tools seems to always copy the `a2o_flux` and `o2a_flux` files from `restart/oasis3mct/` when doing restart runs. However, these files are links to the latest year oasis restart files (for example, `a2o_flux_21000101-21001231`  and `o2a_flux_21000101-21001231`). This is not correct when we want to do restart from other years (like `2098` or `2099`). 
Therefore, I suggest to always use the proper year file `a2o_flux_*` and `o2a_flux_*` instead of `a2o_flux` and `o2a_flux` in esm_tools restart.
Pengyang

nwieters:


nwieters:
I tried to solve this issue by adapting the oasis part in the e.g. awicm3.yaml file. But after discussing this with @mandresm I skipped this solution to a solution inside the oasis.py -> add_restart_files routine.

nwieters:
Here is my first attempt for the solution to solve the oasis restart issue in `oasis.py`. Some checks failed. Do you see why?
For now I will open it for review. Any suggestions for improvement are welcome.

nwieters:
I checked oasis.py with `pycodestyle` and it complained a few lines. I can change this after review.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -711,10 +711,15 @@ oasis3mct:
...
-                                                rstos.nc: ${input_dir}/${fesom.nproc}/${rstos_file_name}.nc
+                                    add_restart_in_sources:
+                                        rmp: ${input_dir}/${fesom.nproc}/rmp_*.nc
+                                        rstas.nc: ${input_dir}/${fesom.nproc}/${rstas_file_name}.nc
+                                        rstos.nc: ${input_dir}/${fesom.nproc}/${rstos_file_name}.nc
+#                                false:
+#                                    add_restart_in_sources:
+#                                        #rmp: ${input_dir}/${fesom.nproc}/rmp_*.nc
+#                                        rstas.nc: rstas.nc_${start_date!syear!smonth!sday}-*
+#                                        rstos.nc: rstos.nc_${start_date!syear!smonth!sday}-*
```
pgierz: Please directly erase and don't just comment out. Git history can tell us what was there before if we need it again :-)

src/esm_runscripts/oasis.py
```diff
@@ -401,6 +401,9 @@ def add_output_file(self, lefts, rights, leftmodel, rightmodel, config):
             config["outdata_sources"][thisfile] = thisfile
 
     def add_restart_files(self, restart_file, fconfig):
+        import glob
+        import os
+
```
pgierz: Imports should go at the top of the file

src/esm_runscripts/oasis.py
```diff
@@ -439,8 +442,48 @@ def add_restart_files(self, restart_file, fconfig):
...
+        # branch-off date
+        if gconfig["run_number"] == 1 and config["lresume"]:  # if branchoff experiment
+            # Check if ini_parent_* or ini_restart_* is used. To make sure ini_restart_*
+            # is always not empty, since I will continue to work with ini_restart_*  here.
+            # This is a bit contrary to prepare.py because there it only is taking care
+            # about the opposite case, if only ini_restart_* is set, than set ini_parent_* 
+            # to the values of ini_restart_*. -> ini_parent_* is always present in config.
+            # ini_restart_dir seems to be always set, even if lresume for oasis is false.?
+            #-----------------------------------------------------------------------------
+            # The follwoing 3 line can be removed, if we switch to only use ini_restart_*
```
pgierz:
```suggestion
            # The following 3 line can be removed, if we switch to only use ini_restart_*
```

src/esm_runscripts/oasis.py
```diff
@@ -439,8 +442,48 @@ def add_restart_files(self, restart_file, fconfig):
...
+            #-----------------------------------------------------------------------------
+            # If set in config (oasis):
+            if "ini_restart_dir" in config and "ini_restart_date" in config:
+                # check if restart file with ini_restart_date in filename is in the restart
+                # folder of the parent experiment to be branched off from:
+                branchoff_restart_file = glob.glob(config["ini_restart_dir"] + \
+                    restart_file + "_????????-" + \
+                    str(config["ini_restart_date"].year) + \
+                    str(config["ini_restart_date"].month).zfill(2) + \
+                    str(config["ini_restart_date"].day).zfill(2))
```
pgierz: Can this be instead written down as an `f-string`? That might make it easier to read.

src/esm_runscripts/oasis.py
```diff
@@ -439,8 +442,48 @@ def add_restart_files(self, restart_file, fconfig):
...
+                    branchoff_restart_file = restart_file
+
+                config["restart_in_sources"][restart_file] = branchoff_restart_file
+            else:
+                # In case it is an initial run, but oasis restart files should be taken
+                # from different source (e.g. pool) as defined e.g. in awicm3.yaml
+                if not restart_file in config["restart_in_sources"]:
+                    config["restart_in_sources"][restart_file] = restart_file
+        else:
+            if not restart_file in config["restart_in_sources"]:
```
pgierz:
```suggestion
            if restart_file not in config["restart_in_sources"]:
```

src/esm_runscripts/prepare.py
pgierz, line 1: In this file, generally, please `if variable not in config` instead of `if not variable in config`

pgierz (changes requested):
I have some small suggestions. If you want to practice unit-testing, it might also be nice to write tests for these functions :-)

nwieters:
I will think about tests for these functions. But I think I will merge it first before starting with tests, or?

pgierz:
That is up to you. You can merge and make a separate PR for tests, or put the tests into this PR. 

In general, I would like to start the following strategy, unless it is too much of a pain. If a user notices a bug, we write that bug into a unit test so that we can fix it and have it documented. That would also help in the future to make sure any new changes we do do not re-introduce the bug.

src/esm_runscripts/oasis.py
nwieters: Review suggestions applied.

src/esm_runscripts/oasis.py
nwieters: Review suggestions applied.

src/esm_runscripts/oasis.py
nwieters: Review suggestions applied.

src/esm_runscripts/prepare.py
nwieters: Changed throughout the file.

src/esm_runscripts/oasis.py
nwieters: Used f-string.

nwieters:
I think I will merge it now and leave the tests for a later pull request. 
I implemented the review suggestions so far, but I think I still need someone to approve them to unblock the merging.

pgierz (approved):

pgierz:
Approved!! Go ahead and merge :-) nice work Nadine!

nwieters:
#bump #patch

src/esm_runscripts/oasis.py
```diff
@@ -98,12 +100,12 @@ def add_coupling(
...
 
         if export_mode == "DEFAULT":
-            if lresume == False:
+            if lresume is False:
                 export_mode = "EXPOUT"
             else:
                 export_mode = "EXPORTED"
 
-        if lresume == False:
+        if lresume is False:
```
mandresm, line 108: Have you considered that many times users define `lresume: 1` or `0`? If a user defines `lresume: 0` or `1` with the new syntax `lresume is False` the result will always be `False` unless the parser does the transformation of `lresume` first. Can someone check that `1` and `0` still work for these lines?

mandresm (commented):
@nwieters, maybe you can include the following comment below in the next PR for the Oasis fixes?

src/esm_runscripts/oasis.py
pgierz:
```
pgierz in 🌐 albedo0 in ~ via 🐍 v3.10.6
❯ cat bool_checks.py
#!/usr/bin/env python3
lresume_bool_off, lresume_bool_on = False, True
lresume_int_off, lresume_int_on = 0, 1
lresume_str_off, lresume_str_in = "False", "True"
for var in lresume_bool_off, lresume_bool_on, lresume_int_off, lresume_int_on, lresume_str_off, lresume_str_in:
    if var:
        print(f"{var=} is 'truthy'")

pgierz in 🌐 albedo0 in ~ via 🐍 v3.10.6
❯ ./bool_checks.py
var=True is 'truthy'
var=1 is 'truthy'
var='False' is 'truthy'
var='True' is 'truthy'
```

nwieters: That is already changed in the commit 5d0443a of this pull request.

pgierz: Note please the danger here:
```python
a = "False" # <-- str of letters spelling False
b = bool(a)
print(b)
True
```

mandresm: @pgierz, in your script you are using `if var` which is very different from `if var1 is var2`. The question here is whether `if 1 is True` and that returns `False`

src/esm_runscripts/oasis.py
```diff
@@ -98,12 +100,12 @@ def add_coupling(
             sep = ":"
 
         if export_mode == "DEFAULT":
-            if lresume == False:
+            if lresume is False:
```
mandresm, line 103: Please, revert this change in the new PR. As I said `0 == False` gives you `True`. `0 is False` gives you `False`

src/esm_runscripts/oasis.py
mandresm, line 108: Please, revert this change in the new PR. As I said `0 == False` gives you `True`. `0 is False` gives you `False`

Issue #826, opened 2022-09-22
OASIS and FESOM restarts not from last day of the experiment in AWI-CM1 and AWI-CM3


antoniofis10000:
Dear developers.
Currently, by default, I think if somebody wants to branch off an experiment using AWI-CM1 (and AWI-CM3 I think) in the middle of the year (or month), by default the restart states from FESOM and OASIS are readed from the last day of the experiment. E.g. We want to use an experiment running from 01/01/2019-31/01/2019 to branch off another experiment starting on 15/01/2019 the ECHAM, HDMODEL and JSBACH states are readed from 14/01/2019 but 31/01/2019 is used for FESOM and OASIS (as a file with a symbolic link to the last state is used).

I think I have implement a solution for AWI-CM1 in: /work/ba1264/a270148/esm-experiments/AWICM/IFDefinitivaContinuationFromChooseWithChooseFromGen/scripts/4KMarylouIFFromPreviousWithChooseFromGeneral.yaml 
that can be interesting for the community.

Thanks in advance.

nwieters:
Hi @antoniofis10000,

regarding OASIS restart files, this issue has been solved with version 6.21.23 (pull request #996).
I will do a test for FESOM restart files. 
Please let me know if you have more recent experience with such a branch-off experiment.

Thank you.

Issue #743, opened 2022-07-21
Oasis restarts


mandresm:
**Describe the bug**
For a restart, Oasis flux files (`a2o` and `o2a` for `AWICM` for example) are always copied from a file which path has no date-stamp, and this is hardcoded in the `src/esm_runscripts/oasis.py`. This is okay for restarts in ongoing simulations as the file without a time-stamp is a link to the latest time-stamped file. However, it's not correct when there is a need to branch-off from an already finished simulation, in some point earlier than the simulation's end. 

**Expected behavior**
Restarts of oasis are handled through the runscript in the same way the other components handle restarts.

**Additional context**
@denizural already started working on it and has a solution that will be pull-requested in a couple of weeks, after the end-of-July workshops.

@chrisdane reported this issue.

nwieters:
This issue has been solved with version 6.21.23 (pull request #996).

PR #971, opened 2023-06-01
Added selection of oasis restart file if more than one is found.

### LINKED ISSUE #969: Choose oasis restart file if not unambiguously clear
In case of a branchoff simulation: Let the correct oasis restart file be chosen by the user if not unambiguously clear from the content of the restart folder of the parent experiment.

nwieters:
closes #969

### DIFF
--- a/src/esm_runscripts/oasis.py
+++ b/src/esm_runscripts/oasis.py
@@ -470,20 +470,19 @@ class oasis:
                     # Note: What if there are more than one found? Let the user decide which one to take.
                     if len(branchoff_restart_file) == 1:
                         branchoff_restart_file = os.path.basename(branchoff_restart_file[0])
-                    elif len(branchoff_restart_file) > 1 and not gconfig["isinteractive"]:
-                        # If more than one restart file found that matches ini_restart_date,
-                        # ask the user to select from the result list:
-                        questions = [
-                            {
-                                "type": "select",
-                                "name": "OASIS restart file",
-                                "message": "More than one OASIS restart file was found for your branchoff experiment that matches the ini_restart_date you selected. Please select one of the following OASIS restart files:",
-                                "choices": branchoff_restart_file,
-                            },
-                        ]
-                        branchoff_restart_file = os.path.basename(questionary.prompt(questions)["OASIS restart file"])
-                    if not gconfig["isinteractive"]:
-                        print(f"OASIS will use {branchoff_restart_file} as restart file from parent experiment {config['ini_restart_dir']}.")
+                    else:
+                        if not gconfig["isinteractive"]:
+                            # If more than one restart file found that matches ini_restart_date,
+                            # ask the user to select from the result list:
+                            questions = [
+                                {
+                                    "type": "select",
+                                    "name": "OASIS restart file",
+                                    "message": "More than one OASIS restart file was found for your branchoff experiment that matches the ini_restart_date you selected. Please select one of the following OASIS restart files:",
+                                    "choices": branchoff_restart_file,
+                                },
+                            ]
+                            branchoff_restart_file = os.path.basename(questionary.prompt(questions)["OASIS restart file"])
                 else:
                     # If no restart file found for this particular ini_restart_dir, use default file 'rstas.nc'
                     branchoff_restart_file = restart_file
--- a/src/esm_runscripts/oasis.py
+++ b/src/esm_runscripts/oasis.py
@@ -457,41 +457,38 @@ class oasis:
             if "ini_restart_dir" in config and "ini_restart_date" in config:
                 # check if restart file with ini_restart_date in filename is in the restart
                 # folder of the parent experiment to be branched off from:
-                glob_search_file = f"{config['ini_restart_dir']}{restart_file}_????????-"\
-                                  +f"{config['ini_restart_date'].year}"                 \
-                                  +f"{config['ini_restart_date'].month:02}"             \
-                                  +f"{config['ini_restart_date'].day:02}"
-                branchoff_restart_file = glob.glob(glob_search_file)
-                branchoff_restart_file.sort()
-                if branchoff_restart_file:
+                glob_search_file = (
+                    f"{config['ini_restart_dir']}{restart_file}_????????-"
+                    f"{config['ini_restart_date'].year}"
+                    f"{config['ini_restart_date'].month:02}"
+                    f"{config['ini_restart_date'].day:02}"
+                )
+                glob_restart_file = glob.glob(glob_search_file)
+                glob_restart_file.sort()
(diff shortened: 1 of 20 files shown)


src/esm_runscripts/oasis.py
```diff
@@ -465,9 +465,25 @@ def add_restart_files(self, restart_file, fconfig):
...
+                        # ask the user to select from the result list:
+                        questions = [
+                            {
+                                "type": "select",
+                                "name": "OASIS restart file",
+                                "message": "More than one OASIS restart file was found for your branchoff experiment that matches the ini_restart_date you selected. Please select one of the following OASIS restart files:",
+                                "choices": branchoff_restart_file,
+                            },
+                        ]
+                        branchoff_restart_file = os.path.basename(questionary.prompt(questions)["OASIS restart file"])
```
pgierz: Here, we are missing an else. What happens not greater or equal 1 are found (so 0)? Then we should throw an error!

src/esm_runscripts/oasis.py
```diff
@@ -465,9 +465,25 @@ def add_restart_files(self, restart_file, fconfig):
...
+                    if not gconfig["isinteractive"]:
+                        print(f"OASIS will use {branchoff_restart_file} as restart file from parent experiment {config['ini_restart_dir']}.")
```
pgierz: This is fine, but at some point (not in this PR, certainly!) we want to move all of these kind of user-info messages into [`loguru.logger`](https://loguru.readthedocs.io/en/stable/).

pgierz (changes requested):
One little "finishing touch" recommendation, but otherwise, looks great! 👍🏻

src/esm_runscripts/oasis.py
nwieters: I placed len(branchoff_restart_file) > 1 back again in the else part. Since len(branchoff_restart_file) == 0 is captured by th outer if statement already.

src/esm_runscripts/oasis.py
nwieters: Removed

src/esm_runscripts/oasis.py
```diff
@@ -465,9 +465,24 @@ def add_restart_files(self, restart_file, fconfig):
...
                 branchoff_restart_file = glob.glob(glob_search_file)
-                # Note: What if there are more than one found?
+                branchoff_restart_file.sort()
                 if branchoff_restart_file:
-                    branchoff_restart_file = os.path.basename(branchoff_restart_file[0])
+                    # Note: What if there are more than one found? Let the user decide which one to take.
+                    if len(branchoff_restart_file) == 1:
+                        branchoff_restart_file = os.path.basename(branchoff_restart_file[0])
+                    else:
+                        if not gconfig["isinteractive"]:
```
pgierz: Shouldn't this be the other way around?
```suggestion
                        if gconfig["isinteractive"]:
```

src/esm_runscripts/oasis.py
```diff
@@ -465,9 +465,24 @@ def add_restart_files(self, restart_file, fconfig):
...
+                            # ask the user to select from the result list:
+                            questions = [
+                                {
+                                    "type": "select",
+                                    "name": "OASIS restart file",
+                                    "message": "More than one OASIS restart file was found for your branchoff experiment that matches the ini_restart_date you selected. Please select one of the following OASIS restart files:",
+                                    "choices": branchoff_restart_file,
+                                },
+                            ]
+                            branchoff_restart_file = os.path.basename(questionary.prompt(questions)["OASIS restart file"])
```
pgierz: I'm not entirely sure from memory, but I think you need `.ask()` still here:
```suggestion
                            branchoff_restart_file = os.path.basename(questionary.prompt(questions).ask()["OASIS restart file"])
```

pgierz (changes requested):
Two more little things I saw, have a look 👍🏻

src/esm_runscripts/oasis.py
nwieters: I thought so too. But (for a reason I do not understand yet) this will be executed twice, actually four time (for each restart file rstas.nc and rstos.nc) and than a second time which will overwrite everything again. At this second round gconfig["isinteractive"] is False. Thats why I ask here for 'if not'.

mandresm:
I cannot review this right now, please wait for my review before merging (on Monday), I have some comments on the code, that need to be addressed as there are a couple of things (coming from the previous merge) that might result in sneaky errors for the users.

src/esm_runscripts/oasis.py
nwieters: This was correct. But I changed it a bit and use questionary.form now. This makes it a few lines shorter.

pgierz:
> I cannot review this right now, please wait for my review before merging (on Monday), I have some comments on the code, that need to be addressed as there are a couple of things (coming from the previous merge) that might result in sneaky errors for the users.

I can make it so that merging requires two approvals, if you want. Since we are a team of three, then everyone has seen everything (unless I forgot how to count), and you could still use "god-mode" to merge for very small changes.

src/esm_runscripts/oasis.py
mandresm: It should be `if gconfig["isinteractive"]` because we want to avoid opening questionairs while ESM-Tools runs in the background on a computing node, that will stall the job. If in the second round `gconfig["isinteractive"] is False` as you said @nwieters, then changing this should not be a problem right?

src/esm_runscripts/oasis.py
```diff
@@ -465,9 +465,20 @@ def add_restart_files(self, restart_file, fconfig):
...
                 if branchoff_restart_file:
-                    branchoff_restart_file = os.path.basename(branchoff_restart_file[0])
+                    # Note: What if there are more than one found? Let the user decide which one to take.
+                    if len(branchoff_restart_file) == 1:
+                        branchoff_restart_file = os.path.basename(branchoff_restart_file[0])
+                    else:
+                        if not gconfig["isinteractive"]:
+                            # If more than one restart file found that matches ini_restart_date,
+                            # ask the user to select from the result list:
+                            message = "More than one OASIS restart file was found for your branchoff experiment that matches the ini_restart_date you selected. Please select one of the following OASIS restart files:"
```
mandresm: @pgierz what's the best Pythonic way to break down this string into multiple lines?

mandresm (changes requested):
I committed some minor changes, and also made a couple of comments. Please, make also sure to change back the `==` in the lresume `if`s of the previous merged PR.

mandresm:
Closes #901 #743 #826

mandresm:
#bump

mandresm (approved):

PR #973, opened 2023-06-07
Filelist-log file to yaml and with checksum, to be used with esm_tests for the new filedicts.py

mandresm:
Draft pull request for the new `log_used_files` method of the `filedicts`.

**TODO**

- [ ] Decide whether we use `kind` or `type` property for the files to define whether they are `input`, `output`...
- [ ] Standardize the variables used in the lo with the ones used at the `_gather_file_movements` (`src`, `dest`, `kind` vs `source`, `target`, `type` in the log)
- [ ] Cleanup the test
- [ ] Add the intermediate step

### DIFF
--- a/src/esm_runscripts/filedicts.py
+++ b/src/esm_runscripts/filedicts.py
@@ -12,6 +12,7 @@ import copy
 import functools
 import glob
 import inspect
+import hashlib
 import os
 import pathlib
 import shutil
@@ -1241,6 +1242,7 @@ class SimulationFileCollection(dict):
             config[sim_file_obj["component"]]["files"][sim_file_id]["src"] = (
                 sim_file_obj.paths[src]
             )
+            config[sim_file_obj["component"]]["files"][sim_file_id]["intermediate"] = None
             config[sim_file_obj["component"]]["files"][sim_file_id]["dest"] = (
                 sim_file_obj.paths[dest]
             )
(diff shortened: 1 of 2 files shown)


src/esm_runscripts/filedicts.py
```diff
@@ -1207,17 +1207,18 @@ def from_config(cls, config: dict):
...
-            movement_type = sim_file_obj.get("movement_type", "cp")
-            if movement_type == "mv":
+            movement_type = sim_file_obj.get("movement_type", "copy")
+            if movement_type == "move":
                 self[sim_file_id]["_filesystem_op"] = getattr(sim_file_obj, "mv")
-            elif movement_type == "cp":
+            elif movement_type == "copy":
                 self[sim_file_id]["_filesystem_op"] = getattr(sim_file_obj, "cp")
-            elif movement_type == "ln":
+            elif movement_type == "link":
```
mandresm, line 1222: @pgierz just so you know, this is a fix. Otherwise `_gather_file_movements` would crash.

src/esm_runscripts/filedicts.py
```diff
@@ -1231,12 +1233,19 @@ def execute_filesystem_operation(
...
             else:
                 raise ValueError(f"Incorrect jobtype specified for {sim_file_obj}")
             sim_file_obj["_filesystem_op"](src, dest)
+            config[sim_file_obj["component"]]["files"][sim_file_id]["src"] = (
+                sim_file_obj.paths[src]
+            )
+            config[sim_file_obj["component"]]["files"][sim_file_id]["intermediate"] = None
+            config[sim_file_obj["component"]]["files"][sim_file_id]["dest"] = (
+                sim_file_obj.paths[dest]
+            )
```
pgierz, line 1248: I'm not 100% sure about this, it's a design thing...

what do you think of the following. At this point:
```python
config[sim_file_obj["component"]]["files"][sim_file_id]
```
we have the `SimulationFile`. Should it know about the current phase of movement it is in, and potentially, if it needs an intermediate location?

I would say no, that job belongs somewhere else. Therefore that info also belongs somewhere else...here you are injecting extra info in the `SimulationFile`'s dictionary, right? ...????

Maybe I am overthinking it.

src/esm_runscripts/filedicts.py
```diff
@@ -1231,12 +1233,19 @@ def execute_filesystem_operation(
         for sim_file_id, sim_file_obj in self.items():
             logger.info(f"Processing {sim_file_id}")
             if config["general"]["jobtype"] == "prepcompute":
-                src, dest = "pool", "work"
+                src, dest = "computer", "work"
```
pgierz, line 1236: Good catch, thanks. I would have noticed that if I had tests.

src/esm_runscripts/filedicts.py
```diff
@@ -1258,3 +1267,66 @@ def resolve_file_movements(config: ConfigSetup) -> ConfigSetup:
...
+
+    Returns
+    -------
+    config : ConfigSetup
+        The complete simulation configuration, potentially modified.
+    """
+    if config["general"].get("verbose", False):
+        logger.info("\n::: Logging used files")
+
+    filetypes = config["general"]["relevant_filetypes"]
```
pgierz: To be discussed: If we use the Enum thing, we should use it everywhere

src/esm_runscripts/filedicts.py
```diff
@@ -1258,3 +1267,66 @@ def resolve_file_movements(config: ConfigSetup) -> ConfigSetup:
     sim_file_collection = SimulationFileCollection.from_config(config)
     config = sim_file_collection.execute_filesystem_operation(config)
     return config
+
+
+def log_used_files(config: ConfigSetup) -> ConfigSetup:
```
pgierz, line 1321: This looks great! The only thing I don't like about this function is that does two things: 1) Gathers all the model files and then 2) writes them to a location. I need number 1 in a separate place for my unknown files, so it would be nice to break this down into smaller pieces. I don't want to need to care (at this point at least) about the recipe order yet, or we will get into the situation that we rely on behaviour of one step of the recipe to even make the next one possible (and yes, I know we cannot avoid that entirely)

pgierz (commented):
I had some ideas, see below. I'm not going to push yes or no on the review yet, I think there are still some points we should discuss.

src/esm_runscripts/filedicts.py
mandresm: `relevant_filtypes` is a list of files that is defined depending on the step `tidy`, `compute`, `preprocess`, ... (`grep -r relevant_filtypes src/esm_runscripts`). I suggest we don't touch it for consistency with previous versions. Also it seems natural to me that the relevant files for the given ESM-Tools phase are defined in the phase itself and not in `filedicts.py`.

pgierz:
Hey @mandresm: want to re-activate work on FileDicts after the summer break? (very very slowly, on the side as a hobby project?)

PR #974, opened 2023-06-09
Sprint/filedices/provenance

mandresm:
I have fixed the errors that were taking place when trying Paul's tests. I have also included a couple of tests in the same file at the end (`src/esm_parser/provenance.py`) in the form of `assertion`. At the moment you'll need to run the whole thing as a script `python3 provenance.py`. Feel free to add more tests and try to break it. It can do updates and inheritates the provenance, which was precisely what we wanted. The thing is an absolute mess in terms of coding... feel free to commit. I'm still not sure if this thing has a future... but now is your turn guys (@nwieters, @pgierz) to judge :)


mandresm:
This is very raw, the tests needs to move to our unit testing folders, the function needs to be cleaned and simplify if possible, ...

pgierz:
Well, I broke it twice during the meeting ;-) I'll have a detailed look Monday after the candidate meetings. The code is certainly not 100% awful, I've looked at other Python that makes me cry on the inside. This is not that, so preliminarily until I have a deep dive: 👍

pgierz (commented):
Looks good as a start, now just move these tests into the `tests/test_esm_parser/test_proveneance` file. Have a look at the pattern we have set up for the runscripts :-)

mandresm:
@pgierz and @nwieters  I have rebased this branch into `release` (it was based on `sprint/filedicts/main` ), since this will be an independent feature from the `filedicts` and it will be merged earlier. I have also changed the target branch to be `feat/provenance` which just departed from `release` and it's to be used as our main branch for the `provenance` development. There we should also merge `ruamel-example`, so that tomorrow @nwieters has everything she needs to continue developing in one single place. @pgierz, all changes you requested for this branch (the one that adds the class `DictWithProvenace`) are solved. Would you approve?

PR #975, opened 2023-06-12
Bugfix in oasis.py if lresume is ...

nwieters:
I fixed the wrong `if lresume is` ... statements to `if lresume ==` ...

### DIFF
--- a/src/esm_runscripts/oasis.py
+++ b/src/esm_runscripts/oasis.py
@@ -105,7 +105,7 @@ class oasis:
             else:
                 export_mode = "EXPORTED"
 
-        if lresume == False:
+        if bool(lresume) is False:
             lag = str(0)
         else:
             lag = direction.get("lag", "0")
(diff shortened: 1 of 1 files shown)


src/esm_runscripts/oasis.py
```diff
@@ -100,12 +100,12 @@ def add_coupling(
             sep = ":"
 
         if export_mode == "DEFAULT":
-            if lresume is False:
+            if lresume == False:
```
pgierz:
```suggestion
            if bool(lresume) is False:
```

src/esm_runscripts/oasis.py
```diff
@@ -100,12 +100,12 @@ def add_coupling(
...
                 export_mode = "EXPOUT"
             else:
                 export_mode = "EXPORTED"
 
-        if lresume is False:
+        if lresume == False:
```
pgierz:
```suggestion
        if bool(lresume) is False:
```

pgierz (changes requested):
Small changes to ensure that `lresume` is transformed into a bool at checking time

mandresm (commented):
#bump!

PR #980, opened 2023-06-19
fix(esm_master/task.py): allow pip-based installed 'extra' packages t…

pgierz:
…o install in venvs


mandresm (approved):
Passes all tests except blogin and albedo (there are problems in esm_tests I did no yet fixed for those two so you can ignore them). Don't forget to bump :)

pgierz:
#bump

PR #982, opened 2023-06-22
Bugfix for selction of oasis restart file.

nwieters:
Bugfix: Asking to select a oasis restart file only when `if not gconfig["isinteractive"]`. This is because when asked when `isinteractive` is `true`, at the second time the `config` is assembled, this selection is forgotten. In such a case, the latest oasis restart file of the parent simulation will be taken.

I tested with this version

1.  normal restart simulations (initial runs with restart every month)
2. branchoff simulation for a given `ini_restart_date` without the need to select a specific oasis restart file and a restart after one month
3. branchoff simulation for a given `ini_restart_date` with selecting for a specific oasis restart file and a restart after one month

Tests 1 and 2 worked
Test 3 started with the correct oasis restart files, but does not make a following restart every month.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.21.18
+current_version = 6.21.19
 commit = True
 tag = True
 
--- a/setup.py
+++ b/setup.py
@@ -92,6 +92,6 @@ setup(
     test_suite="tests",
     tests_require=test_requirements,
     url="https://github.com/esm-tools/esm_tools",
-    version="6.21.18",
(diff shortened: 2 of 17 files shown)


nwieters:
Can someone please reproduce the behaviour of Test 3? Or do you have a hint why a following restart in this case is not started?

nwieters:
With this bugfix all three test simulations described above are running.

mandresm (approved):
Thanks for catching this bug and solving it! #bump!

mandresm:
Dear GitHub actions... can you #bump ?

PR #987, opened 2023-06-30
Feat/provenance loading

nwieters:


ruamel-examples/example.yaml
```diff
@@ -6,7 +6,7 @@ person:
     I am here to make your life impossible
   "
   # My comment
-  my_var: !ENV MY_VAR
```
mandresm, line 9: Is this still not working?

src/esm_parser/parse_it.py
```diff
@@ -0,0 +1,254 @@
+#!/usr/bin/env python3
```
mandresm: This whole file should disappear once we finish with this PR

src/esm_parser/yaml_to_dict.py
```diff
@@ -169,39 +173,51 @@ def yaml_file_to_dict(filepath):
...
                 # Check for duplicates
                 check_duplicates(yaml_file)
                 # Back to the beginning of the file
                 yaml_file.seek(0, 0)
+                yaml_load = {}
                 # Actually load the file
-                yaml_load = yaml.load(yaml_file, Loader=loader)  # yaml.FullLoader)
+#                yaml_load = yaml.load(yaml_file, Loader=loader)  # yaml.FullLoader)
+
+                esm_tools_loader = parse_it.EsmToolsLoader()
```
mandresm: I think the `EsmToolsLoader` class belongs in this file

src/esm_parser/parse_it.py
```diff
@@ -0,0 +1,254 @@
...
+                            config_prov[key].append(_extract_dict(elem)[1])
+                        else:
+                            config_prov[key].append(prov)
+                            config[key].append(elem)
+                else:
+                    prov = { }
+                    prov["line"], prov["col"], prov["yaml_file"], prov["category"] = value_prov
+                    config_prov[key] = prov
+                    config[key] = value
+            return (config, config_prov)
```
mandresm: I would suggest this is a method of the EsmToolsLoader, and not nested in `load`.

mandresm (commented):
Looking good so far :)

ruamel-examples/example.yaml
nwieters: It is working. But I turned it off, just because I was lazy to export MY_VAR all the time. But !ENV USER or !ENV ${USER} works.

src/esm_parser/parse_it.py
nwieters: Ok. Then I will integrate it in the yaml_to_dict. I had it already but re-winded all because I was confused at some point and started all over again.

src/esm_parser/yaml_to_dict.py
nwieters: Yes. I have this fixed in my local branch already. But I will move EsmToolsLoader etc. to yaml_to_dict.py anyway.

src/esm_parser/parse_it.py
nwieters: Done

ruamel-examples/example.yaml
nwieters: Actually no. Now `!ENV ${USER}` works, but not `!ENV USER`. Will fix that.

mandresm: It's normal that `!ENV USER` does not work. I think it should remain like that.

mandresm: Or was it working before?

src/esm_parser/yaml_to_dict.py
```diff
@@ -733,7 +730,7 @@ def load(self, stream):
         mapping_with_tuple_prov = super().load(stream)[0]
 
         config, provenance = self._extract_dict(mapping_with_tuple_prov)
-
+        self.env_variables = self.constructor.env_variables
```
nwieters, line 733: I couldn't find out, why `self.Constructor.env_variables` does not work. This attribute is set in the `EnvironmentConstructor` class. `self.env_variables = self.Constructor.env_variables` complaines that 

`type object 'ProvenanceConstructor' has no attribute 'env_variables'`

`self.Constructor` exists, but does not have attribute `env_variables`.

pgierz:
This is already merged: so, no review updates. Just tag me on new things if they pop up.

PR #989, opened 2023-07-03
Provenance/dict and list with provenance

mandresm:
Hey guys!

This PR will be key for tomorrow's demo and for integrating what @nwieters has been working on.

Provenance is inherited!!!! (after some dark magic)

I'll merge as soon as I get an approval. Then we can merge `feat/provenance` into Nadine's so that we can keep her branch up to date.

### DIFF
--- a/src/esm_parser/esm_parser.py
+++ b/src/esm_parser/esm_parser.py
@@ -592,7 +592,7 @@ def new_dict_merge(dct, merge_dct, winner="to_be_included"):
             new_dict_merge(dct[k], merge_dct[k], winner)
         else:
             if not (winner == "receiving" and k in dct):
-                assign_key_val_with_provenance(dct, merge_dct, k)
+                dct[k] = merge_dct[k]
 
 
 def new_deep_update(
@@ -681,9 +681,9 @@ def dict_merge(dct, merge_dct, resolve_nested_adds=False, **kwargs):
             # is empty and protection is requested
             if k in dct:
                 if dct[k] and not merge_dct[k] and dont_overwrite_with_empty_value:
-                    assign_key_val_with_provenance(merge_dct, dct, k)
+                    merge_dct[k] = dct[k]
 
-            assign_key_val_with_provenance(dct, merge_dct, k)
+            dct[k] = merge_dct[k]
 
 
 def deep_update(chapter, entries, config, blackdict={}):
@@ -1882,6 +1882,7 @@ def determine_regex_list_match(test_str, regex_list):
     return any(result)
 
 
+@keep_id_provenance
 def find_variable(tree, rhs, full_config, white_or_black_list, isblacklist):
     raw_str = rhs
     if not tree[-1]:
@@ -2242,6 +2243,7 @@ def determine_computer_from_hostname():
     # )
 
 
+@keep_id_provenance
 def do_math_in_entry(tree, rhs, config):
     if not tree[-1]:
         tree = tree[:-1]
@@ -2346,6 +2348,7 @@ def do_math_in_entry(tree, rhs, config):
     return convert(entry.strip(), tree)
 
 
+@keep_id_provenance
 def mark_dates(tree, rhs, config):
     """Adds the ``DATE_MARKER`` to any entry who's key ends with ``"date"``"""
     if not tree[-1]:
@@ -2360,6 +2363,7 @@ def mark_dates(tree, rhs, config):
     return entry
 
 
+@keep_id_provenance
 def marked_date_to_date_object(tree, rhs, config):
     """Transforms a marked date string into a Date object"""
     if not tree[-1]:
@@ -2390,6 +2394,7 @@ def marked_date_to_date_object(tree, rhs, config):
     return entry
 
 
+@keep_id_provenance
 def unmark_dates(tree, rhs, config):
     """Removes the ``DATE_MARKER`` to any entry who's entry contains the ``DATE_MARKER``."""
     if not tree[-1]:
@@ -2401,6 +2406,7 @@ def unmark_dates(tree, rhs, config):
     return entry
 
 
+@keep_id_provenance
 def perform_actions(tree, rhs, config):
     if not tree[-1]:
         tree = tree[:-1]
@@ -2451,6 +2457,7 @@ def perform_actions(tree, rhs, config):
     return entry
 
 
+@keep_id_provenance
 def purify_booleans(tree, rhs, config):
     if not tree[-1]:
         tree = tree[:-1]
--- a/src/esm_parser/esm_parser.py
+++ b/src/esm_parser/esm_parser.py
@@ -87,6 +87,7 @@ import yaml
 
 # functions reading in dict from file
 from .yaml_to_dict import *
+from .provenance import *
 
 # Date class
 from esm_calendar import Date
@@ -591,7 +592,7 @@ def new_dict_merge(dct, merge_dct, winner="to_be_included"):
             new_dict_merge(dct[k], merge_dct[k], winner)
         else:
             if not (winner == "receiving" and k in dct):
-                dct[k] = merge_dct[k]
+                assign_key_val_with_provenance(dct, merge_dct, k)
 
 
 def new_deep_update(
@@ -680,9 +681,9 @@ def dict_merge(dct, merge_dct, resolve_nested_adds=False, **kwargs):
             # is empty and protection is requested
             if k in dct:
                 if dct[k] and not merge_dct[k] and dont_overwrite_with_empty_value:
-                    merge_dct[k] = dct[k]
+                    assign_key_val_with_provenance(merge_dct, dct, k)
 
-            dct[k] = merge_dct[k]
+            assign_key_val_with_provenance(dct, merge_dct, k)
 
 
 def deep_update(chapter, entries, config, blackdict={}):
@@ -837,7 +838,7 @@ def add_entries_from_chapter(config, add_chapter, add_entries):
         if type(config[add_chapter]) == list:
             for entry in my_entries:
                 config[add_chapter].append(entry)
-        elif type(config[add_chapter]) == dict:
+        elif isinstance(config[add_chapter], dict):
             # MA: I'm not supper happy about the resolve_nested_adds implementation
             dict_merge(
                 config[add_chapter],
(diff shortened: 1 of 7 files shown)


mandresm:
By the way, the docstrings are way deprecated, please don't pay attention to them...

src/esm_parser/esm_parser.py
```diff
@@ -1133,7 +1134,7 @@ def merge_dicts(*dict_args):
     -------
     A merged dictionary (shallow).
     """
-    result = {}
+    result = {} #DictWithProvenance({}, {"provenance": None})
```
pgierz, line 1137:
```suggestion
    result = {}
```

pgierz: Keep it clean ;-) You can always go back into the git history if you need that again. Basically this code does not change...

src/esm_parser/esm_parser.py
```diff
@@ -87,6 +87,7 @@
 
 # functions reading in dict from file
 from .yaml_to_dict import *
+from .provenance import *
```
pgierz, line 90: Just a style comment, you can leave this as is: I don't particularly like wildcard imports, it makes it more difficult to see where the code is coming from.

src/esm_parser/esm_parser.py
```diff
@@ -837,7 +838,7 @@ def add_entries_from_chapter(config, add_chapter, add_entries):
         if type(config[add_chapter]) == list:
             for entry in my_entries:
                 config[add_chapter].append(entry)
-        elif type(config[add_chapter]) == dict:
+        elif isinstance(config[add_chapter], dict):
```
pgierz, line 841: Thanks, that's much nicer 👍🏻

src/esm_parser/esm_parser.py
```diff
@@ -1881,6 +1882,7 @@ def determine_regex_list_match(test_str, regex_list):
     return any(result)
 
 
+@keep_id_provenance
```
pgierz, line 1885: If I see it correctly, you add this decorator to every possible `recursive_run_function` function, yes?

I prefer the decorator, since it's right next to the function definition, but you could also do in `recursive_run_function`:

```python
def recursive_run_function(..., func, ...):
    # ... some other code
    func = keep_id_provenance(func)
    # ... rest of body
```

src/esm_parser/provenance.py
```diff
@@ -63,6 +65,8 @@ class DictWithProvenance(dict):
         provenance_dict = dict_with_provenance.get_provenance()
     """
 
+    leaf_id_provenance = {}
+
```
pgierz, line 69: A doc string would useful here. A mini suggestion, but of course you can probably word this more accurately than I can:
```suggestion
    leaf_id_provenance = {}
    """
    dict : 
        This is a class-level variable to hold IDs of objects in memory. It is
        at the class level because this information needs to be passed around
        between various instances
    """
```

src/esm_parser/provenance.py
```diff
@@ -130,18 +134,27 @@ def get_provenance(self, dictionary=None):
...
             dictionary = self
 
         for key, val in dictionary.items():
             if isinstance(val, dict):
                 provenance_dict[key] = self.get_provenance(val)
+            elif hasattr(dictionary, "provenance"):
+                provenance_dict[key] = dictionary.provenance.get(key)
             else:
-                provenance_dict[key] = dictionary.provenance[key]
+                provenance_dict[key] = None
```
pgierz, line 146: I'm not sure I understand this case. A comment would be helpful for future maintenance

src/esm_parser/provenance.py
```diff
@@ -183,14 +196,62 @@ def __setitem__(self, key, val):
...
+            val_id = id(val)
+            if isinstance(dictionary, DictWithProvenance) and not isinstance(val, dict):
+                self.provenance[key] = dictionary.provenance.get(key,
+                    DictWithProvenance.leaf_id_provenance.get(val_id, None)
+                )
+            elif isinstance(dictionary, dict) and not isinstance(val, dict):
+                self.provenance[key] = DictWithProvenance.leaf_id_provenance.get(val_id, None)
+
+
+    def set_leaf_id_provenance(self, key):
```
pgierz, line 234: Please check if the doctoring is correct as suggested:
```suggestion
    def set_leaf_id_provenance(self, key):
        """
        Stores the last-leaf provenance information in the class level 
        variable ``leaf_id_provenance``.
        
        This method gets the ``id`` value (unique Python object counter), which
        is used as a key in the `leaf_id_provenance`. The value becomes the
        provenance of that key, or defaults to `None`
        
        Parameters
        ----------
        key : Any
           The key of the "inner-most" leaf to store provenance information for
        """
```

Question: Why `None`, and not `KeyError`? Shouldn't **everything** have a provenance?

src/esm_parser/provenance.py
```diff
@@ -183,14 +196,62 @@ def __setitem__(self, key, val):
...
+            val_id = id(super().__getitem__(key))
+            # Stores the provenance in a class variable, under an id key
+            DictWithProvenance.leaf_id_provenance[val_id] = self.provenance.get(key, None)
+
+
+def keep_id_provenance(func):
+    """
+    This is a decorator for recursive functions in esm_parser to preserve
+    provenanve. Better docstrings will come later...
+    """
```
pgierz, line 246:
```suggestion
    """
    Decorator for recursive functions in esm_parser to preserve
    provenance.
    
    Recursive run functions in ``esm_parser`` are generally called on the innermost
    leaf. Here, we still run the function, but additionally store the output of the
    function into the `leaf_id_provenance` container so that provenance can be added
    to the result of the function call.
    
    Parameters
    ----------
    func : Callable
        The function to decorate
    """
```

src/esm_parser/provenance.py
```diff
@@ -183,14 +196,62 @@ def __setitem__(self, key, val):
...
+
+    def set_leaf_id_provenance(self, key):
+        # If it's a leaf
+        if not isinstance(super().__getitem__(key), DictWithProvenance):
+            val_id = id(super().__getitem__(key))
+            # Stores the provenance in a class variable, under an id key
+            DictWithProvenance.leaf_id_provenance[val_id] = self.provenance.get(key, None)
+
+
+def keep_id_provenance(func):
```
pgierz, line 242: I would find it cool if this also had some sort of info attached to it: "Modified by <FUNC NAME>" or something like that. Not sure we will ever need it, but it couldn't hurt to have...

Note: low priority

src/esm_parser/yaml_to_dict.py
```diff
@@ -199,9 +201,11 @@ def yaml_file_to_dict(filepath):
                     yaml_load["computer"]["runtime_environment_changes"][
                         "add_export_vars"
                     ] = add_export_vars
-                # Check for empty components/models
-                # check_for_empty_components(yaml_load, filepath + extension)
-                return yaml_load
+
+            yaml_load = DictWithProvenance(yaml_load, {"filepath": filepath})
```
pgierz, line 205: This looks suspiciously "too easy"

src/esm_runscripts/last_minute.py
```diff
@@ -96,7 +96,7 @@ def _modify_config_with_settings(config, settings):
             selected_config = config
             for k2 in path_to_key:
                 selected_config = selected_config[k2]
-            if type(selected_config) == dict:
+            if isinstance(selected_config, dict):
```
pgierz, line 99: 👍🏻

src/esm_runscripts/oasis.py
```diff
@@ -144,7 +144,7 @@ def add_coupling(
             allpres = [allpres]
         for pre in allpres:
             detail_line = ""
-            if type(pre) == dict:
+            if isinstance(pre, dict):
```
pgierz, line 147: 👍🏻

tests/test_esm_parser/test_provenance.py
```diff
@@ -120,3 +120,22 @@ def test_get_provenance_6():
...
+            "type": "a_string",
+            "files": {
+                "greenhouse": {"kind": "a_new_string", "path_in_computer": "a_string"}
+            },
+        },
+        "fesom": {"asd": 2, "model": 2, "update_test": "new_provenance"},
+        "computer": None,
+        True: None,
+    }
+    assert config.get_provenance() == check_provenance
```
pgierz, line 141: Do I see it correctly that updates now only work if you are updating a `DictWithProvenance` from another `DictWithProvenance`? What happens if you update a `DictWithProvenance` with a regular boring `dict`?

pgierz (commented):
I'm not going to officially push the green button yet or officially request changes, have a look at my comments first. I may have been out of the loop with this sprint for too long to have a real feel for the code.

What is here looks good, there is not too much dark magic happen 😉 

Before this gets released to users, I would really like a nice top-level module doc-string (probably just so I can use the thing correctly later on)

Overall, excellent work! 😀

mandresm:
Thanks for your feedback @pgierz! We'll cleanup for sure before anything goes into release :)

src/esm_parser/esm_parser.py
mandresm: I thought about that but there is the recursive function `list_to_multikey` that won't work like that because it actually mutates and transforms the content of a dictionary (it's or looping syntax). So for now I thought is better to have the decorator for the more simple recursive functions.

src/esm_parser/yaml_to_dict.py
mandresm: This is just a test, in Nadine's work is a bit more complicated, but in conceptually is that easy ;)

tests/test_esm_parser/test_provenance.py
mandresm: https://github.com/esm-tools/esm_tools/blob/9bfbd32e4e9ba0c73c9aa9d734dd095de02038f8/src/esm_parser/provenance.py#L221-L231

No provenance is assigned.

mandresm:
Thanks @pgierz! Really appreciated the detailed review, you save me some writing of docstrings!

I have added your suggestions to another branch that also includes the changes by @nwieters, and I will close this PR now.

pgierz:
(Shrug) It's Python, and not `chmod` ;-) I have the nice gut feeling to know what to look for. 

Overall great work! I'll keep an eye on the sprint PRs over the next days.

PR #994, opened 2023-07-07
fix/oasis branchoff awicm3

mandresm:
fix oasis restart for awicm3 when branching off an experiment but loading the oasis fluxes from the pool. It also fixes it for a branchoff experiment which only has 1 run

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.21.21
+current_version = 6.21.22
 commit = True
 tag = True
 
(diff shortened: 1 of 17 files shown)


mandresm:
@JanStreffing, this is actually a good solution and general considering how ESM-Tools creates those files :) way easier than expected and it works. Qiyun just tried it. We need it on release asap so that he can start a simulation with venv.

JanStreffing (approved):

mandresm:
#bump!

PR #996, opened 2023-07-11
fix/oasis_branchoff_and_restarts

mandresm:
fix multiple bugs in the oasis coupling field restarts

**TODO**
- [x] Extensive testing

@JanStreffing

### DIFF
--- a/configs/setups/awicm3/awicm3.yaml
+++ b/configs/setups/awicm3/awicm3.yaml
@@ -706,9 +706,9 @@ oasis3mct:
                 true:
                         lag: "${time_step}"
                         # If ini_restart_dir is not specified in the runscript and, therefore,
-                        # it is not a branched off experiment add the /fesom.nproc/ to the
+                        # it is not a branched off experiment adds the /fesom.nproc/ to the
                         # sources when the run_number is 1
-                        load_restart_from_pool: "$(( '${input_dir}' == '${ini_restart_dir}' and ${general.run_number} == 1))"
+                        load_restart_from_pool: "$(( os.path.realpath('${input_dir}') == os.path.realpath('${ini_restart_dir}') and ${general.run_number} == 1 ))"
                         choose_load_restart_from_pool:
                                 true:
                                     add_restart_in_sources:
--- a/esm_tools/motd/motd.yaml
+++ b/esm_tools/motd/motd.yaml
@@ -134,9 +134,9 @@ release6.7:
         action: DELAY(1)
         announcement: true
 
-oasis_branchoff6.21.17:
+oasis_branchoff6.21.23:
         package: "esm_tools"
-        versions: "<6.21.17"
+        versions: "<6.21.23"
         message: "\x1b[1m\x1b[96mOASIS BRANCH-OFF RESTARTS FIXED!\x1b[0m\n
 
             The bug on the copying of the wrong flux files from OASIS \x1b[96mwhen running\n
--- a/motd/motd.yaml
+++ b/motd/motd.yaml
@@ -134,9 +134,9 @@ release6.7:
         action: DELAY(1)
         announcement: true
 
-oasis_branchoff6.21.17:
+oasis_branchoff6.21.23:
         package: "esm_tools"
-        versions: "<6.21.17"
+        versions: "<6.21.23"
         message: "\x1b[1m\x1b[96mOASIS BRANCH-OFF RESTARTS FIXED!\x1b[0m\n
 
             The bug on the copying of the wrong flux files from OASIS \x1b[96mwhen running\n
--- a/src/esm_runscripts/oasis.py
+++ b/src/esm_runscripts/oasis.py
@@ -491,7 +491,7 @@ class oasis:
             # If the restart file path is not defined set it to be the same as the
             # ini_restart_dir
             if not restart_file_path:
-                restart_file_path = config["ini_restart_dir"]
+                restart_file_path = f"{config['ini_restart_dir']}/{restart_file}"
             # If set in config (oasis):
             if "ini_restart_dir" in config and "ini_restart_date" in config:
                 # check if restart file with ini_restart_date in filename is in the restart
--- a/src/esm_runscripts/oasis.py
+++ b/src/esm_runscripts/oasis.py
@@ -489,7 +489,7 @@ class oasis:
             if "ini_parent_dir" in config and "ini_restart_dir" not in config:
                 config["ini_restart_dir"] = config["ini_parent_dir"]
             # If the restart file path is not defined, or it's not an absolute path to
-            # the file set it to be the same as the ini_restart_dir
+            # the file, set it to be the same as the ini_restart_dir
             if not restart_file_path or restart_file_path==restart_file:
                 restart_file_path = f"{config['ini_restart_dir']}/{restart_file}"
             # If set in config (oasis):
--- a/src/esm_runscripts/oasis.py
+++ b/src/esm_runscripts/oasis.py
@@ -411,7 +411,7 @@ class oasis:
         Cases
         -----
         1. If this run is a restart but not a branch-off experiment, set the source to
-            be the same as defined by the user in ``restart_in_in_sources`` or the same
+            be the same as defined by the user in ``restart_in_sources`` or the same
             as the name coming from ``coupling_<target/input>_fields``, if the first is
             missing (the normal case).
         2. Same as case 1 but with the time stamp added to the name of the restart file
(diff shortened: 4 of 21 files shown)


configs/setups/awicm3/awicm3.yaml
```diff
@@ -706,9 +706,9 @@ oasis3mct:
                 true:
                         lag: "${time_step}"
                         # If ini_restart_dir is not specified in the runscript and, therefore,
-                        # it is not a branched off experiment add the /fesom.nproc/ to the
+                        # it is not a branched off experiment adds the /fesom.nproc/ to the
                         # sources when the run_number is 1
-                        load_restart_from_pool: "$(( '${input_dir}' == '${ini_restart_dir}' and ${general.run_number} == 1))"
+                        load_restart_from_pool: "$(( os.path.realpath('${input_dir}') == os.path.realpath('${ini_restart_dir}') and ${general.run_number} == 1 ))"
```
pgierz, line 711: Wait wait wait, we are allowing people to run completely arbitrary Python directly from the YAML file? That is a very very bad idea... How long has this been a thing?

configs/setups/awicm3/awicm3.yaml
```diff
@@ -706,9 +706,9 @@ oasis3mct:
                 true:
                         lag: "${time_step}"
                         # If ini_restart_dir is not specified in the runscript and, therefore,
-                        # it is not a branched off experiment add the /fesom.nproc/ to the
+                        # it is not a branched off experiment adds the /fesom.nproc/ to the
```
pgierz, line 709: 👍🏻

esm_tools/motd/motd.yaml
```diff
@@ -134,9 +134,9 @@ release6.7:
         action: DELAY(1)
         announcement: true
 
-oasis_branchoff6.21.17:
+oasis_branchoff6.21.23:
         package: "esm_tools"
-        versions: "<6.21.17"
+        versions: "<6.21.23"
```
pgierz, line 139: 👍🏻

motd/motd.yaml
```diff
@@ -134,9 +134,9 @@ release6.7:
         action: DELAY(1)
         announcement: true
 
-oasis_branchoff6.21.17:
+oasis_branchoff6.21.23:
         package: "esm_tools"
-        versions: "<6.21.17"
+        versions: "<6.21.23"
```
pgierz, line 139: 👍🏻

JanStreffing:
Perhaps @jian-shi72 can do a check run of the original runscript that we found the issue on.

mandresm:
I've tested everything I could imagine with AWICM3 and AWICM1:
- initial simulation
- continuing an initial simulation in the same folder by increasing the final_date
- standard branchoff (getting the file from the parent directory)
- standard branchoff with 2 latest dates matching (that triggers the question to the user about which one to use)
- setting `lresume: True` for Oasis and making sure it takes it from the pool and uses the correct name as extracted from `oasis3mct.restart_in_sources`, defining the `ini_restart_` variables, and also without defining them.

I think it should be bullet proof now... but let's see.

@pgierz can you review? :)

src/esm_runscripts/oasis.py
```diff
@@ -488,9 +488,9 @@ def add_restart_files(self, restart_file_label, fconfig):
                 config["ini_restart_date"] = config["ini_parent_date"]
             if "ini_parent_dir" in config and "ini_restart_dir" not in config:
                 config["ini_restart_dir"] = config["ini_parent_dir"]
-            # If the restart file path is not defined set it to be the same as the
-            # ini_restart_dir
-            if not restart_file_path:
+            # If the restart file path is not defined, or it's not an absolute path to
+            # the file set it to be the same as the ini_restart_dir
```
mandresm:
```suggestion
            # the file, set it to be the same as the ini_restart_dir
```

src/esm_runscripts/oasis.py
```diff
@@ -404,11 +404,35 @@ def add_output_file(self, lefts, rights, leftmodel, rightmodel, config):
...
 
-    def add_restart_files(self, restart_file, fconfig):
+    def add_restart_files(self, restart_file_label, fconfig):
+        """
+        Handles the special restart case of the coupling fields.
+
+        Cases
+        -----
+        1. If this run is a restart but not a branch-off experiment, set the source to
+            be the same as defined by the user in ``restart_in_in_sources`` or the same
```
mandresm:
```suggestion
            be the same as defined by the user in ``restart_in_sources`` or the same
```

src/esm_runscripts/oasis.py
```diff
@@ -431,73 +455,89 @@ def add_restart_files(self, restart_file, fconfig):
...
         if gconfig["run_number"] == 1 and config["lresume"] and gconfig["jobtype"] == "prepcompute":
             # If they do not exist, define ``ini_restart_date`` and ``ini_restart_dir``
             # based on ``ini_parent_date`` and ``ini_parent_dir``
             if "ini_parent_date" in config and "ini_restart_date" not in config:
                 config["ini_restart_date"] = config["ini_parent_date"]
             if "ini_parent_dir" in config and "ini_restart_dir" not in config:
                 config["ini_restart_dir"] = config["ini_parent_dir"]
+            # If the restart file path is not defined, or it's not an absolute path to
+            # the file, set it to be the same as the ini_restart_dir
+            if not restart_file_path or restart_file_path==restart_file:
```
pgierz: Minor style fix:

```suggestion
            if not restart_file_path or restart_file_path == restart_file:
```

src/esm_runscripts/oasis.py
```diff
@@ -431,73 +455,89 @@ def add_restart_files(self, restart_file, fconfig):
...
+                        message = (
+                            "More than one OASIS restart file was found for "
+                            "your branchoff experiment that matches the "
+                            "ini_restart_date you selected. Please select "
+                            "one of the following OASIS restart files:"
+                        )
+                        answers = questionary.form(
+                            restarts = questionary.select(message, choices=glob_restart_file)
+                        ).ask()
+                        restart_file = answers["restarts"]
```
pgierz, line 536: I like this a lot 😄

pgierz (approved):

configs/setups/awicm3/awicm3.yaml
mandresm: Since I have memory. Is it a bad idea? Potentially. Would removing this capability kill many features from AWICM3 and make us have to implement even more syntax into the yaml? for sure

src/esm_runscripts/oasis.py
mandresm: @nwieters work, not just one indentation level up, for a more generalized case

configs/setups/awicm3/awicm3.yaml
pgierz: Not thumbs down to how you did it, thumbs down to...just....mmmmpf. I probably should have told Dirk this is dangerous from the beginning before it was so established.

mandresm:
#bump

configs/setups/awicm3/awicm3.yaml
JanStreffing: Is it? All of our users can just edit the backend anyway if they like.

pgierz: Well, maintenance cost is the factor here: If users NEED to change the backend by hand, then they can do whatever they want. Most people are afraid of that -- it's the backend, we have often enough said: you look at this only if you know exactly what you are doing -- so it's unlikely to happen. If users could just change one little yaml and then (given clever yaml skills) erase their entire folder, and then come to ask us why its all gone, that is a less happy scenario (again: this will never actually happen)

JanStreffing: If the user wants to delete their work dir we can't stop them. Life finds a way ;-)

mandresm: Honestly, I see the danger on it, but I cannot live without it. I wouldn't expect that anyone writes a long script here, mostly because it needs to be a one-liner and you cannot create variables in the backend outside of the config via this method. That is IMO safe enough. On the other hand, this allows us to do complex boolean comparisons which is excellent for `choose_` blocks, without complicating choose blocks or having to invent an if syntax. In summary I love it.

And one funny note, `$(( ))` syntax of ESM-Tools was implemented by you guys thinking to allow math operations. I think you did not imagine that someone will find that anything you put inside the parenthesis would run within an `eval()`. Sorry, is partly my fault for finding it ;)

PR #997, opened 2023-07-13
Feat/awicm3.1.2 

JanStreffing:
Minor version of  awicm3 where the orbital forcing double radiation call bug has been fixed. As a result turning on the orbital forcing at 1850 orbital values no longer results in warming compared to 1850 with orbital forcing off.


mandresm (approved):
You're on fire @JanStreffing ;)

mandresm:
#approve-changes

mandresm:
#bump

PR #1002, opened 2023-07-24
Catch wrong syntax in add_namelist_changes block

### LINKED ISSUE #1001: Catch when users does add_namelist_changes without giving namelist section
A few weeks ago I had a case where I made an operator error in namelist changes. I did:

```
    add_namelist_changes:
        fort.4:
            LHVOLCA: false
            LCMIP6: true
            NAERANT_SCALE: 0
            CMIP6DATADIR: '/work/ab0246/a270092/input/oifs-43r3/cmip6-data/'
            NCMIPFIXYR: 1850
            SSPNAME: 'historical'
```

instead of
```
    add_namelist_changes:
        fort.4:
            NAERAD:
                LHVOLCA: false
                LCMIP6: true
                NAERANT_SCALE: 0
                CMIP6DATADIR: '/work/ab0246/a270092/input/oifs-43r3/cmip6-data/'
                NCMIPFIXYR: 1850
                SSPNAME: 'historical'
```

I got a cryptic python error message about fitting a boolean into a float or so. Today @Mccino had a similar error. In both cases I only caught it with breakpoints in namelists.py and looking at the specific namelist change that crashes the code.

I think we should have a check in place that verifies the number of levels in a namelist change. I.e. that ensures, that for a namelist change we had 1) the namelist, 2) the section, and 3) the variable.

If we don't have three levels, we should crash and give a clear message saying, the namelist change in line x does not follow FORTRAN syntax.

nwieters:
Closes #1001

### DIFF
--- a/src/esm_runscripts/namelists.py
+++ b/src/esm_runscripts/namelists.py
@@ -147,7 +147,7 @@ class Namelist:
                     if not isinstance(change_entries, dict):
                         nml_syntax_error = True
                         this_is_wrong = (
-                            f"There is a syntax error in the following lines:\n\n[add_]namelist_changes:\n    {namelist}\n        {change_chapter}: {change_entries}\n...")
+                            f"There is a syntax error in the following lines:\n\n[add_]namelist_changes:\n    '{namelist}':\n        '{change_chapter}: {change_entries}'\n...")
                         break
         if nml_syntax_error:
             user_error(
--- a/src/esm_runscripts/namelists.py
+++ b/src/esm_runscripts/namelists.py
@@ -215,7 +215,6 @@ class Namelist:
                         # `key` is the processed variable from f90nml module and
                         # is lowercase.
                         remove_original_key = False
-    
                         # traverse the namelist chapter and see if a mixed case
                         # variable is also found
                         for key2 in namelist_changes[namelist][change_chapter]:
--- a/src/esm_runscripts/namelists.py
+++ b/src/esm_runscripts/namelists.py
@@ -103,7 +103,8 @@ class Namelist:
         ----------
         namelist_changes : nested dict
         
-        If the syntax is correct, namelist_changes should be a nested dict of the following form:
+        If the syntax is correct, namelist_changes should be a nested dict of the following form::
+
         {'namelist1.nml': {'namelist1': {'variable1': 'value1', 'variable2': 'value2', 'variable3': 'value3'}, 'namelist2': {'variable1': value1}}}
         
         Returns
--- a/src/esm_runscripts/namelists.py
+++ b/src/esm_runscripts/namelists.py
@@ -91,6 +91,74 @@ class Namelist:
         return mconfig
 
     @staticmethod
+    def nmls_check_changes(namelist_changes):
+        """
+        Checks if namelist changes are given in correct syntax.
+        If not, a user_error will be raised and stopp the execution.
+        
+        Programmer Information
+        ----------------------
+        
+        Parameters
+        ----------
+        namelist_changes : nested dict
+        
+        If the syntax is correct, namelist_changes should be a nested dict of the following form:
+        {'namelist1.nml': {'namelist1': {'variable1': 'value1', 'variable2': 'value2', 'variable3': 'value3'}, 'namelist2': {'variable1': value1}}}
+        
+        Returns
+        -------
+        None
+        
+        Calls to other methods:
+        ----------------------
+        """
+
+        error_message = (
+            "There is a syntax error, probably in your runscript (but potentially in other yaml files), "
+            "regarding namelist changes (e.g. in a 'add_namelist_changes' block). "
+            "It seems that either 'namelist_file' or "
+            "'namelist_group' or 'both' are missing.\n"
+            "Please make sure that namelist changes are specified in the correct syntax (see example below)"
+            " and rerun your runscript.\n"
+        )
+        example = (
+            "\nExample for [add_]namelist_changes:\n"
+            "\t [add_]namelist_changes:\n"
+            "\t    '<namelist_file>':\n"
+            "\t        '<namelist_group>':\n"
+            "\t            <variable>: <value>"
+        )
+
+        nml_syntax_error = False
+
+        for namelist in list(namelist_changes):
+            changes = namelist_changes[namelist]
+            # Check if namelist_changes are specified in correct syntax (e.g. in runscript)
+            # If correct syntax, changes is always a dict.
+            if not isinstance(changes, dict):
+                nml_syntax_error = True
+                this_is_wrong = (f"There is a syntax error in the following lines:\n\n[add_]namelist_changes:\n    {namelist}: {changes}\n...")
+#                this_is_wrong = [namelist, changes]
+                break
+            else:
+                for change_chapter in list(changes):
+                    change_entries = changes[change_chapter]
+                    # Check if namelist_changes are specified in correct syntax (e.g. in runscript)
+                    # If correct syntax, change_entries is always a dict.
+                    if not isinstance(change_entries, dict):
+                        nml_syntax_error = True
+                        this_is_wrong = (
+                            f"There is a syntax error in the following lines:\n\n[add_]namelist_changes:\n    {namelist}\n        {change_chapter}: {change_entries}\n...")
+                        break
+        if nml_syntax_error:
+            user_error(
+                "Syntax error in namelist changes",
+                f"{error_message}\n{this_is_wrong}\n{example}",
+                dsymbols=["``", "'"],
+            )
+
+    @staticmethod
     def nmls_remove(mconfig):
         """
         Removes an element from a namelist chapter.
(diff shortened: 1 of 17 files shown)


nwieters:
The following possible syntax errors in a add_namelist_changes block are now cached:

- Missing namelist group:

```
add_namelist_changes:
    fort.4:
        LHVOLCA: false
        LCMIP6: true
        CMIP6DATADIR: '/work/ab0246/a270092/input/oifs-43r3/cmip6-data/'
        NCMIPFIXYR: 1850
        SSPNAME: 'historical'
```
- Missing namelist filename:
```
add_namelist_changes:
    NAERAD:
        LHVOLCA: false
        LCMIP6: true
        CMIP6DATADIR: '/work/ab0246/a270092/input/oifs-43r3/cmip6-data/'
        NCMIPFIXYR: 1850
        SSPNAME: 'historical'
```
- Missing namelist filename and group:
```
add_namelist_changes:
    LHVOLCA: false
    LCMIP6: true
    CMIP6DATADIR: '/work/ab0246/a270092/input/oifs-43r3/cmip6-data/'
    NCMIPFIXYR: 1850
    SSPNAME: 'historical'
```

These syntax errors will cause an `user_error` and stop the execution with an error message that hints to the cause of the error.

JanStreffing (approved):
Tested and works well, Thank you, @nwieters!

src/esm_runscripts/namelists.py
```diff
@@ -126,46 +126,77 @@ def nmls_remove(mconfig):
 
         namelist_changes = mconfig.get("namelist_changes", {})
         namelist_removes = []
+        error_message = (
+            "There is a syntax error, probably in your runscript, regarding namelist "
```
mandresm:
```suggestion
            "There is a syntax error, probably in your runscript (but potentially in other yamls), regarding namelist "
```

src/esm_runscripts/namelists.py
```diff
@@ -126,46 +126,77 @@ def nmls_remove(mconfig):
...
-                        for key2 in namelist_changes[namelist][change_chapter]:
-                            # take care of the MiXeD FORTRAN CaSeS
-                            if key2.lower() == key.lower() and key2 != key:
-                                original_key = key2
-                                remove_original_key = True
-                                namelist_removes.append(
-                                    (namelist, change_chapter, original_key)
+            # Check if namelist_changes are specified in correct syntac (e.g. in runscript)
+            # If correct syntax, changes is always a dict.
+            if isinstance(changes, dict):
```
mandresm: If here you write:
```python
            if not isinstance(changes, dict):
                user_error(
                    "Syntax error in namelist changes",
                    error_message,
                    dsymbols=["``", "'"],
                )
```
is better because:
1. you don't need a new level of indentation in the following lines
2. because you don't need an extra level of indentation the git diffs actually speak only about the actual changes to the functioning of the code
3. You have one line less (no need for the `else` line)

In any case, no action is required here as my final suggestion is to make an independent `nmls_changes_check` function.

src/esm_runscripts/namelists.py
```diff
@@ -126,46 +126,77 @@ def nmls_remove(mconfig):
...
+                logging.debug("Determining remove entires for %s", namelist)
+                logging.debug("All changes: %s", changes)
+                for change_chapter in list(changes):
+                    change_entries = changes[change_chapter]
+                    # Check if namelist_changes are specified in correct syntac (e.g. in runscript)
+                    # If correct syntax, change_entries is always a dict.
+                    if isinstance(change_entries, dict):
```
mandresm: Same as above

src/esm_runscripts/namelists.py
```diff
@@ -126,46 +126,77 @@ def nmls_remove(mconfig):
 
         namelist_changes = mconfig.get("namelist_changes", {})
         namelist_removes = []
+        error_message = (
+            "There is a syntax error, probably in your runscript, regarding namelist "
+            "changes (e.g. in a 'add_namelist_changes' block). "
+            "It seems that either 'namelist_file' or "
+            "'namelist_group' or 'both' are missing.\n"
```
mandresm: I've looked at the code and it's actually possible to give a hint to what is exactly wrong. For instance, because you have the variable `namelist`, `changes` and `change_chapter` (only on the second if), you can actually point to there is something wrong with that. For example
```
This is wrong:
    [add_]namelist_changes:
        fort.4:
            LHVOLCA: false
Make sure your namelist changes have a syntax that follows this nesting rule:
     [add_]namelist_changes:
            <namelist_file>:
                    <namelist_group>:
                            <variable>: <value>
```

src/esm_runscripts/namelists.py
```diff
@@ -126,46 +126,77 @@ def nmls_remove(mconfig):
...
+            "Please make sure that namelist changes are specified in the correct syntax (see example below)"
+            " and rerun your runscript.\n"
+            "Example for add_namelist_changes:\n"
+            "\t add_namelist_changes:\n"
+            "\t\t'<namelist_file>':\n"
+            "\t\t\t'<namelist_group>':\n"
+            "\t\t\t\t<variable>: <value>"
```
mandresm: The convention we are trying to impose is 4 spaces as indentation rule. Can you change that?

mandresm (changes requested):
Thanks for taking care of this @nwieters! I have some minor comments and a major one, but all the functionality is there.

The minor comments you can find them in the suggestions.

The major comment: I think this development belongs to it's own function, for example `nmls_changes_check` that is called at the beginning of `nmls_remove` and `nmls_modify`. You probably want to make this function a `@staticmethod` as well so that is callable from outside the class. The insides of that function you have them already, basically copying the loop and if structure of what you already have.

The reason why this is probably better is because, although it adds more looping, it's clearer and a method contains only one functionality, in this case removing namelists and checking namelist changes should have each their methods.

src/esm_runscripts/namelists.py
nwieters: I thought so at first. But I learned that you should always put the most likely case in the beginning of an if statement. That is probably outdated code design.

mandresm: I think in terms of performance, what you are saying still holds. So here we are walking the line of performance/readability. However, for this particular case the performance gain would probably be insignificant? (we are only changing a `isinstance()` by a `not isinstance()`, so it's almost the same operation and the code interpreted in the ifs lines is minimal).

nwieters:
Thank you @mandresm for the review.
I placed the check for the namelist_changes syntax into its own method and added a hint to the error message where the syntax error might be caused.

src/esm_runscripts/namelists.py
```diff
@@ -136,7 +207,7 @@ def nmls_remove(mconfig):
                     value = change_entries[key]
                     if value == "remove_from_namelist":
                         namelist_removes.append((namelist, change_chapter, key))
-
+
```
mandresm: Apologize my OCD for pointing out this trailing space... on my defense I will say that having less diffs means it's easier to browse through the git history.
```suggestion

```

By the way, one can add the following to `.vimrc` to trailing spaces (if you are that kind of person):
```
highlight ExtraWhitespace ctermbg=red guibg=red
match ExtraWhitespace /\s\+$/
```

src/esm_runscripts/namelists.py
```diff
@@ -145,7 +216,7 @@ def nmls_remove(mconfig):
                         # `key` is the processed variable from f90nml module and
                         # is lowercase.
                         remove_original_key = False
-
+
```
mandresm:
```suggestion

```

src/esm_runscripts/namelists.py
```diff
@@ -156,12 +227,12 @@ def nmls_remove(mconfig):
                                 namelist_removes.append(
                                     (namelist, change_chapter, original_key)
                                 )
-
+
```
mandresm:
```suggestion

```

src/esm_runscripts/namelists.py
```diff
@@ -156,12 +227,12 @@ def nmls_remove(mconfig):
...
                                     (namelist, change_chapter, original_key)
                                 )
-
+    
                         # remove both lowercase and mixed case variables
                         del namelist_changes[namelist][change_chapter][key]
                         if remove_original_key:
                             del namelist_changes[namelist][change_chapter][original_key]
-
+
```
mandresm:
```suggestion

```

src/esm_runscripts/namelists.py
```diff
@@ -90,6 +90,71 @@ def nmls_load(mconfig):
                 mconfig["namelists"][nml].uppercase = True
         return mconfig
 
+    @staticmethod
+    def nmls_check_changes(namelist_changes):
+        """
+        Checks if namelist changes are given in correct syntax.
+        If not, a user_error will be raised and stopp the execution.
```
mandresm:
```suggestion
        If not, a user_error will be raised and stop the execution.
```

src/esm_runscripts/namelists.py
```diff
@@ -90,6 +90,71 @@ def nmls_load(mconfig):
...
+        
+        Programmer Information
+        ----------------------
+        
+        Parameters
+        ----------
+        namelist_changes : nested dict
+        
+        If the syntax is correct, namelist_changes should be a nested dict of the following form:
+        {'namelist1.nml': {'namelist1': {'variable1': 'value1', 'variable2': 'value2', 'variable3': 'value3'}, 'namelist2': {'variable1': value1}}}
```
mandresm:
```suggestion
        If the syntax is correct, namelist_changes should be a nested dict of the following form::

        {'namelist1.nml': {'namelist1': {'variable1': 'value1', 'variable2': 'value2', 'variable3': 'value3'}, 'namelist2': {'variable1': value1}}}
```
This should make sphinx generate a code box for the namelist dictionary, when building the documentation

src/esm_runscripts/namelists.py
```diff
@@ -90,6 +90,71 @@ def nmls_load(mconfig):
...
+        error_message = (
+            "There is a syntax error, probably in your runscript (but potentially in other yaml files), "
+            "regarding namelist changes (e.g. in a 'add_namelist_changes' block). "
+            "It seems that either 'namelist_file' or "
+            "'namelist_group' or 'both' are missing.\n"
+            "Please make sure that namelist changes are specified in the correct syntax (see example below)"
+            " and rerun your runscript.\n"
+        )
+        example = (
+            "\nExample for [add_]namelist_changes:\n"
```
mandresm:
```suggestion
            "\nExample of a ``correct syntax`` for [add_]namelist_changes:\n"
```

src/esm_runscripts/namelists.py
```diff
@@ -90,6 +90,71 @@ def nmls_load(mconfig):
...
+
+        nml_syntax_error = False
+
+        for namelist in list(namelist_changes):
+            changes = namelist_changes[namelist]
+            # Check if namelist_changes are specified in correct syntax (e.g. in runscript)
+            # If correct syntax, changes is always a dict.
+            if not isinstance(changes, dict):
+                nml_syntax_error = True
+                this_is_wrong = (f"There is a syntax error in the following lines:\n\n[add_]namelist_changes:\n    {namelist}: {changes}\n...")
```
mandresm:
```suggestion
                this_is_wrong = (f"There is a syntax error in the following lines:\n\n[add_]namelist_changes:\n    '{namelist}: {changes}'\n...")
```

src/esm_runscripts/namelists.py
```diff
@@ -90,6 +90,71 @@ def nmls_load(mconfig):
...
+                break
+            else:
+                for change_chapter in list(changes):
+                    change_entries = changes[change_chapter]
+                    # Check if namelist_changes are specified in correct syntax (e.g. in runscript)
+                    # If correct syntax, change_entries is always a dict.
+                    if not isinstance(change_entries, dict):
+                        nml_syntax_error = True
+                        this_is_wrong = (
+                            f"There is a syntax error in the following lines:\n\n[add_]namelist_changes:\n    {namelist}\n        {change_chapter}: {change_entries}\n...")
```
mandresm:
```suggestion
                            f"There is a syntax error in the following lines:\n\n[add_]namelist_changes:\n    '{namelist}':\n        '{change_chapter}: {change_entries}'\n...")
```

mandresm (approved):
Thanks for this nice error handling work @nwieters!

Just some minor suggestions. Feel free to bump and merge after comming those.

nwieters:
#bump

PR #1010, opened 2023-07-31
Feat/provenance variables with prov

mandresm:
So this is the new proposed approach for the provenance. So far it can do a successful run with AWICM3 and stores all the provenance information correctly for all the types I could imagine.

**TODO**
- [x] Fix a problem with `get_provenance` for lists
- [x] Docstrings - To be done after merging the provenance history
- [x] Unit tests


src/esm_parser/provenance.py
```diff
@@ -1,5 +1,119 @@
...
+
+            def __init__(self, value, provenance=None):
+                self.provenance = provenance
+
+        # Instantiate the subclass with the given value and provenance
+        return WrapperWithProvenance(value, provenance)
+
+
 class DictWithProvenance(dict):
     """
```
mandresm, line 118: From this point the docstrings are completele out of date, so no point to check the docstrings...

nwieters (approved):

PR #972, opened 2023-06-06
Filelist-log file to yaml and with checksum, to be used with esm_tests for the old filelists.py

mandresm:
**Note**: this PR is to `release` because we want to be able to create this file with the old `filelists.py` and then compare it to our new `filedicts.py` when we are ready to merge our new super feature.

This PR includes the following:
- The filelists log file now lands in the `log` folders, not in `config`
- The filelists log is now a yaml file with the following nesting:
   ```yaml
    <component>:
        <file_tag>:
            source: foo
            intermediate: foo
            target: foo
            checksum: foo
   ```
- checksum is `false` if the file is not found
- deleted `.backup` folder in `src/esm_runscripts`

**TODO**
- [x] Actual tests in levante
- [x] Add the new file log to `esm_tests` last tested info
- [x] Add `compute_and_log_file_checksums` to all `prepcompute` recipes
- [x] Clean up 

Example of the new file:
```yaml
fesom:
  config:
    checksum: 2aa73d9f1a98c7ecb1e53605dc6962f2
    intermediate: /Users/mandresm/Work/file_log/run_20000101-20000131/config/fesom/namelist.config
    source: /Users/mandresm/Codes/esm_tools/namelists//fesom2/2.0/awicm3/v3.1//namelist.config
    target: /Users/mandresm/Work/file_log/run_20000101-20000131/work/namelist.config
    type: config
  fesom.clock:
    checksum: 5291ed1d422d040fcbd37b90a7d52292
    intermediate: /Users/mandresm/Work/file_log/run_20000101-20000131/config/fesom/fesom.clock
    source: /Users/mandresm/Work//file_log/run_20000101-20000131/config/fesom//fesom.clock
    target: /Users/mandresm/Work/file_log/run_20000101-20000131/work/fesom.clock
    type: config
  forcing:
    checksum: f3545eaac138f8efe24de0ed9598a768
    intermediate: /Users/mandresm/Work/file_log/run_20000101-20000131/config/fesom/namelist.forcing
    source: /Users/mandresm/Codes/esm_tools/namelists//fesom2/2.0/awicm3/v3.1//namelist.forcing
    target: /Users/mandresm/Work/file_log/run_20000101-20000131/work/namelist.forcing
    type: config
  ice:
    checksum: 71435aca9a8fd094447d010f28c59686
    intermediate: /Users/mandresm/Work/file_log/run_20000101-20000131/config/fesom/namelist.ice
[...]

PR #1013, opened 2023-08-01
File Checksums

pgierz:
First steps for file checksums. See https://app.clickup.com/t/861n41hpy

### DIFF
--- a/src/esm_runscripts/compute.py
+++ b/src/esm_runscripts/compute.py
@@ -1,29 +1,50 @@
+import copy
+import hashlib
 import os
+import pathlib
 import shutil
 import subprocess
 import sys
-import copy
-import pathlib
 
 import f90nml
 import questionary
 import yaml
-from colorama import Fore, Back, Style, init
+from colorama import Back, Fore, Style, init
+from loguru import logger
 
-import esm_tools
 import esm_calendar
 import esm_parser
 import esm_runscripts
+import esm_tools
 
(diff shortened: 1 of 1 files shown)


src/esm_runscripts/compute.py
```diff
@@ -1,31 +1,52 @@
...
 import esm_calendar
 import esm_parser
 import esm_runscripts
+import esm_tools
 
 from .batch_system import batch_system
 from .filelists import copy_files, log_used_files
 from .helpers import end_it_all, evaluate, write_to_log
 from .namelists import Namelist
-from loguru import logger
```
pgierz, line 22: Only "new" thing is `hashlib`, the rest is just `isort` doing its thing. At some point I really should show the team [`pre-commit`](https://pre-commit.com)....

src/esm_runscripts/compute.py
```diff
@@ -300,6 +321,7 @@ def _write_finalized_config(config):
     ----------
     config : esm-tools config object
     """
+
```
pgierz, line 324: All changes below this line are just `black` running in the background.

mandresm:
Isn't this duplicated here #972?

pgierz:
....yes it is. Sorry, forgot about that one. 

We should start merging, there are too many half finished things floating around. I'm starting to lose track. We dont want to enter the fesom refactoring world.

PR #1020, opened 2023-08-10
remove pickling from esm_master that breaks the provenance

mandresm:


src/esm_master/compile_info.py
```diff
@@ -337,16 +337,7 @@ def get_relevant_info(relevant_entries, raw_config, merge_into_this_config=None)
 class setup_and_model_infos:
     def __init__(self, vcs, general, parsed_args):
 
-        if not os.path.isfile(ESM_MASTER_PICKLE):
```
pgierz, line 340: Is this variable also gone?

mandresm: good point!

nwieters (approved):

PR #1021, opened 2023-08-11
Provenance is a list of provenances with the value's history

mandresm:
- Provenance is now a subclass of a list
- Each element can contain any value, but typically it will be a dictionary with file, col, line...
- Each element on that list corresponds to a step in the history of the provenance of that value

**TODO**

- [x] docstrings
- [x] adapt unit tests


mandresm:
This is ready to go, so... here we goooo!

PR #1025, opened 2023-08-21
Create stale.yml

nwieters:
This pull request adds a github-actions workflow to set inactive issues (older than 365 days, this is for a first test and can be decreased later) to 'Stale'. The following steps are done:

- the issue will be labelled with 'Stale'
- a comment will be added to the issue
- and if no further changes will be done to the issue, it will be closed after 30 days.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,31 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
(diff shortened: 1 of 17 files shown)


pgierz (approved):
This looks exactly like what we talked about, excellent!! Does it also work on things that are already online retroactively? That would unfortunately still be a step to be done "by hand", but we can live with that. 

There is no really nice way to test that particular problem unless we open something ourselves and then just wait, and I think we all have better uses for our time ;-)

nwieters:
Thanks @pgierz ,
let's see what happens tomorrow at 14:00. I hope that it will do this retroactive. If not I will reduce the time until it is marked as stale to something shorter.

nwieters:
#bump

mandresm:
Interesting... What about enhancement issues? Those should stay until we have time to take care of them.

PR #1030, opened 2023-08-22
fix(install.sh): small annoyances fixed for install script

pgierz:


pgierz:
Don't close, this can be a nice practice merge for someone (e.g. the viz hi-wi?) during on-boarding.

PR #1032, opened 2023-08-23
Update stale.yml

nwieters:
Update the stale action configuration file. Inactive PR are set to stale but will not be closed after 30 days. Added the stale message for PR accordingly.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -22,7 +22,7 @@ jobs:
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         days-before-issue-stale: 365
-        stale-before-pr-stale: 365
+        days-before-pr-stale: 365
         days-before-issue-close: 30
         days-before-pr-close: -1
         stale-issue-message: 'This issue has been inactive for the last 365 days. It will now be marked as stale and closed after 30 days of further inactivity. Please add a comment to reset this automatic closing of this issue or close it if solved.'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -21,11 +21,11 @@ jobs:
     - uses: actions/stale@v5
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-stale: 365
-        stale-before-pr-stale: -1
-        days-before-close: 30
+        days-before-issue-stale: 365
+        stale-before-pr-stale: 365
+        days-before-issue-close: 30
         days-before-pr-close: -1
         stale-issue-message: 'This issue has been inactive for the last 365 days. It will now be marked as stale and closed after 30 days of further inactivity. Please add a comment to reset this automatic closing of this issue or close it if solved.'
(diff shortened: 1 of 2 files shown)


.github/workflows/stale.yml
```diff
@@ -21,11 +21,11 @@ jobs:
     - uses: actions/stale@v5
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-stale: 365
-        stale-before-pr-stale: -1
-        days-before-close: 30
+        days-before-issue-stale: 365
+        stale-before-pr-stale: 365
```
pgierz: I do not even a little bit like the name of this option. `stale-before-pr-stale` is just....stupid..... But if that's the thing that works, OK.

.github/workflows/stale.yml
```diff
@@ -21,11 +21,11 @@ jobs:
...
+        days-before-issue-close: 30
         days-before-pr-close: -1
         stale-issue-message: 'This issue has been inactive for the last 365 days. It will now be marked as stale and closed after 30 days of further inactivity. Please add a comment to reset this automatic closing of this issue or close it if solved.'
-        stale-pr-message: 'This pull request has been incative for the last days.'
+        stale-pr-message: 'This PR has been inactive for the last 365 days. It will now be marked as stale. Please close this PR if no longer needed.'
```
pgierz, line 29: 👍🏻

.github/workflows/stale.yml
```diff
@@ -21,11 +21,11 @@ jobs:
...
         stale-issue-label: 'Stale'
-        stale-pr-label: 'no-pr-activity'
+        stale-pr-label: 'Stale'
```
pgierz: 👍🏻

.github/workflows/stale.yml
```diff
@@ -21,11 +21,11 @@ jobs:
     - uses: actions/stale@v5
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-stale: 365
-        stale-before-pr-stale: -1
-        days-before-close: 30
+        days-before-issue-stale: 365
+        stale-before-pr-stale: 365
+        days-before-issue-close: 30
```
pgierz, line 26: 👍🏻

pgierz (approved):
I understand most of this, except for the option names that rub me in the wrong way. But go ahead and put it in, we can fix anything else that might go wrong later on ;-)

pgierz:
Wait wait, this cannot go in at least in it's current state. It's breaking almost all of the auto-tests, because of some version conflict. I don't want to give people a release version which does not work. It might be the case that my version pinning stuff needs to go in first, or that we need to modify the CI.

That might need to be a second fix and then we re-try this with a fixed release branch.

.github/workflows/stale.yml
nwieters: No, this is a typo. It should be days-before-pr-stale

nwieters:
Yes, I also saw that the tests were failing. I will wait and also fix the typo first. It is not so urgent to merge in.

PR #1033, opened 2023-08-23
Update esm_tools_actions.yml

pgierz:
Allows to test against various Python versions (if the robot did the right thing)


pgierz:
This at least now tests against many various Pythons. Tests are still broken.

PR #1034, opened 2023-08-23
Update install.sh

pgierz:
Removes language stuff to see if that helps with auto-tests


pgierz:
Not stale, and there is no useful reason to force people to use American English settings in their shell. Someone should look at this again or just merge it.

PR #1035, opened 2023-08-23
Test to fix the python depency error during tests.

nwieters:
This seems to fix for the failing tests due to version incompatibility.


nwieters:
With this I can install esm-tools, the tests all run through, and pytest passes all tests as well.

Test fail for albedo and blogin.

PR #1038, opened 2023-08-24
Fix/test in container error

nwieters:
Test to fix the github-actions error for test in container.

I set the  python-version that are used in esm_tools_actions.yaml to only 3.8.10. This is the python version available in the docker container esm_test.
My understanding is now, that the previous tests with docker container failed, since there is no other python version installed.

### DIFF
--- a/.github/workflows/esm_tools_actions.yml
+++ b/.github/workflows/esm_tools_actions.yml
@@ -21,7 +21,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Setup Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/esm_tools_actions.yml
+++ b/.github/workflows/esm_tools_actions.yml
@@ -11,7 +11,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.8.10']
 
     container:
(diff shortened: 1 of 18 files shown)


.github/workflows/esm_tools_actions.yml
```diff
@@ -1,62 +1,64 @@
...
     runs-on: ubuntu-latest
-    # Docker Hub image that `container-job` executes in
     env:
       LC_ALL: en_US.UTF-8
       LANG: en_US.UTF-8
+
+    strategy:
+      matrix:
+        python-version: ['3.8.10']
+
```
mandresm: If we are bound to the python version of the container we can simply remove this completely. This test is not for testing python versions, we have other tests for that.
```suggestion
```

mandresm (changes requested):
Thanks for the fix and for adding the stale feature for enhancement labels and feature requests!

.github/workflows/esm_tools_actions.yml
nwieters: I removed the strategy as suggested. @pgierz can you live with it?

nwieters:
This PR includes the changes in PR #1032 and #1036.

.github/workflows/esm_tools_actions.yml
pgierz: (grumble grumble something about beauty and elegance) Yes I can.

mandresm: sorry Paul, we have the same tests without container with all the versions. The advantage of this is that if someone changes the container to a new python version it will still work.

mandresm (approved):

mandresm:
#bump

PR #1040, opened 2023-08-30
fix orbital forcing for echam when the radctl namelist is empty

mandresm:
Related to #1031 

Fixes a bug discovered by @jian-shi72 and @JanStreffing in which the changes in the `namelist.echam` for using transient orbital forcing where not happening if `radctl` namelist was not existing in the `namelist.echam`.

This should not affect Lu Niu's dual hemisphere simulations because she has `radctl` in the `namelist.echam`, and I would guess that's the case for all echam simulations.

The reason why they find this bug is because they where using an equivalent algorithm for OpenIFS, where the equivalent namelist (`NAMORB`) is missing in `fort.4` most of the time.

The problem comes from this line:
https://github.com/esm-tools/esm_tools/blob/7c48cea1045ca0f45d8b8c781d464e81c2af7c57/src/esm_runscripts/namelists.py#L265

If `radctl` exists the returned value is a mapping linked to `nml` and `config`, so changing `radctl` key values will change those in `nml` and `config`. In contrast if `radctl` does not exist `.get` does not create a key in `nml`. Instead is simply setting the "unlinked" value `f90nml.namelist.Namelist()` to `radctl`.


nwieters (approved):

mandresm:
#bump

PR #1041, opened 2023-09-04
Sprint/workflow_manager

### LINKED ISSUE #1047: Workflow Manager: Improve exception handling and error messages.
- Replace `exit()` by `user_error` or `user_note`

nwieters:
Refactoring workflow manager

Can be tested by calling 
`esm_runscripts <runscript.yaml> -i workflow`

Examples:
https://esm-tools.readthedocs.io/en/sprint-workflow_manager_documentation/workflow.html#examples-for-the-definition-of-new-workflow-phases

Closes #1047


src/esm_plugin_manager/esm_plugin_manager.py
```diff
@@ -18,11 +18,26 @@ def read_recipe(recipe, additional_dict, needs_parse=True):
...
+    Arguments:
+        plugins_bare -- disctionary as it is read in by function 'read_recipe'
+        recipe -- dictionary of all workitems of an recipe
+        needs_parse -- True (default) or False
+
+    Returns:
+        plugins - dictionary that has information for each workitem of recipe:
+                    module: e.g. esm_runscripts
+                    submodule: e.g. prepare (this is the Python file where the workitem function is defined.
+                    type: e.g. core
```
mandresm:
```suggestion
    Reads in plugin information from the pluginfile = esm_plugins.yaml

    Parameters
    ----------
    plugins_bare : dict
        Dictionary as it is read in by function ``read_recipe``
    recipe : dict
        Dictionary of all workitems of a recipe
    needs_parse : bool
        True (default) or False

    Returns
    -------
    plugins : dict
        Dictionary that has information for each workitem of the recipe:
        - module: e.g. esm_runscripts
        - submodule: e.g. prepare (this is the Python file where the workitem function is defined.
        - type: e.g. core
```

src/esm_plugin_manager/esm_plugin_manager.py
```diff
@@ -117,11 +142,23 @@ def check_plugin_availability(plugins):
...
+    Works through the esm_runscripts recipes and plugin recipes.
+
+    Arguments:
+        recipe -- dictionary            # What is in these two dictionaries? Where do the entries are comming from?
+        plugins -- dictionary
+        config -- dictionary
+
+    Returns:
+        config
+    """
```
mandresm:
```suggestion
    """
    Works through the esm_runscripts recipes and plugin recipes.

    Parameters
    ----------
        recipe : dict            # What is in these two dictionaries? Where do the entries are comming from?
        plugins : dict
        config : dict

    Returns
    -------
        config : dict
    """
```

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
+    last_task_in_queue = ""
+    next_run_triggered_by = ""
+
+    def __init__(self, phases, always_run_with=[]):
+        """
+        Create a new workflow.
+
+        Arguments:
+            phases -- List of workflow phases
+            always_run_with -- List of phases that precedes each phase
```
mandresm: The syntax for the input parameters and outputs in the docstrings is as in the suggestions above, so that sphinx can parse and formate them nicely. The text is treated as `rst`. That means that to highlight code, variables, etc you need to use `` around the <code commands> instead of the single quote from markdown.

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
+        """
+        # TODO: NW call here the phase object ???
+        self.phases = phases
+        self.always_run_with = always_run_with
+
```
mandresm, line 159:
```suggestion

    @property
```
Decorating this as a property means that you can use `self.num_phases_in_workflow` instead of `self.num_phases_in_workflow()`. I would also suggest that you change the name to `num_phases` since you are already in the workflow class/object.

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
+        """
+        Return the number of phases in workflow.
+        """
+        return len(self.phases)
+
+    def write_to_config(self, config):
+        """
+        Write to config.
+        """
+        # NW: It is assumed here, that there are no workflows in config["general"]
```
mandresm:
```suggestion
        # It is assumed here, that there are no workflows in config["general"]
```
We can see how wrote the comment with git blame, so at least Paul and myself had until now the agreement that no person is to be named in the comments unless we are doing something risky. Totally up to you whether you decide to remove it or not, just so you are aware of what our practices are.

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
+
+    def check_user_workflow_dependency(self):
+        """
+        Check whether the user defined workflow phases are independent from eachother or not.
+        """
+        independent = False
+        user_phases_names = [phase.name for phase in self.user_phases]
+        run_after_list = [phase.run_after for phase in self.user_phases]
+        run_before_list = [phase.run_before for phase in self.user_phases]
+        if not set(user_phases_names).intersection(set(run_after_list).union(set(run_before_list))):
```
mandresm, line 403: Can we have a comment here explaining this if? Something on the lines:
```
# If no ``user_phases_names`` are found in the union of ``run_before_list`` and ``run_after_list``
```

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
+
+        return independent
+
+    def check_unknown_phases(self):
+        """
+        Check if any user phase addresses an unknown workflow phase.
+        """
+        unknown_user_phase = True
+        phases_names = [phase.name for phase in self.phases]
+        user_phases_names = [phase.name for phase in self.user_phases]
```
mandresm: This is the second time you use this two lines, should we make a workflow property out of them?

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
+        # Filter out all falsy items (e.g. [], "", None)
+        run_after_list = list(filter(None, [phase.run_after for phase in self.user_phases]))
+        run_before_list = list(filter(None, [phase.run_before for phase in self.user_phases]))
+
```
mandresm, line 414: Can you add a comment here, please?

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
+class UserWorkflowPhase(WorkflowPhase):
+    """A user workflow phase class."""
+    script = None
+    script_dir = None
+    call_function = None
+    env_preparation = None
+
+    def __init__(self, phase_name):
+        self.name = phase_name
+        batch_or_shell = "batch"
```
mandresm: Why is this exclusive from `UserWorkflowPhase`?

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
+    submit_to_batch_system = True
+    run_on_queue = None
+    cluster = None
+    next_submit = []
+    called_from = None
+    batch_or_shell = "SimulationSetup"
+    order_in_cluster = "sequential"
+    run_only = None
+    skip_chunk_number = None
+    skip_run_number = None
```
mandresm: This variable are class variables. This is what ChatGPT has to say about the topic (and was also my understanding):

> The class attributes like name, nproc, and others are not intended to be used as default values for instances of the class. If you want to provide default values, they should be specified in the __init__ method. Class attributes are shared among all instances, which can lead to unexpected behavior.

This is a nice example I found online:
```
class Shark:

    # Class variables
    animal_type = "fish"
    location = "ocean"

    # Constructor method with instance variables name and age
    def __init__(self, name, age):
        self.name = name
        self.age = age
```
A shark will always be a fish, that's why is a class variable. Then its name, and age should be instance variables because they are associated to each individual shark. I would argue that the location should also be an instance variable because who has not seen Sharknado?

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
-    #
-    config = init_total_workflow(config)
-    config = collect_all_workflow_information(config)
-    config = complete_clusters(config)
-    config = order_clusters(config)
-    config = prepend_newrun_job(config)
+    Arguments:
+        config -- dictionary
+    Returns:
+        config
```
mandresm: same as above

src/esm_runscripts/workflow.py
```diff
@@ -1,49 +1,189 @@
...
+    # initialize the default workflow as Workflow object
+    # TODO: NW where are these default phases defined???
+    # 2. Initialize default workflow phases
+    workflow = init_default_workflow(workflow, config)
+    # 3. Read in workflows from runscript and config files
+    workflow = collect_all_user_workflows(workflow, config)
+
+    #config = collect_all_workflow_information(config)
+
+# Why do I need to do the following function call?
```
mandresm: I could only review until here today. I'll try a bit more tomorrow.

mandresm (changes requested):
Great job so far @nwieters! I'm half way through the review, I'll continue tomorrow. It's way more clear, however, I don't yet understand the whole architecture of the new code, but so far it makes sense to me. Maybe we need to consider changing some of the methods below to be moved into the Workflow class, but that's a discussion for later.

configs/esm_software/esm_runscripts/defaults.yaml
```diff
@@ -1,11 +1,17 @@
...
-
+    file_movements:
+        default:
+            all_directions: copy
+        bin:
+            init_to_exp:    copy
+            exp_to_run:     copy
+            run_to_work:    copy
+            work_to_run:    copy
+default_workflow_phases:
```
mandresm: I would call it`workflow`: you're already in the `defaults.yaml` file and the workflow has the `phases` property below.

src/esm_runscripts/workflow.py
nwieters: Thank you for the explanation.

configs/esm_software/esm_runscripts/defaults.yaml
nwieters: I agree. Because it will be in the config dict section `config["general"]["defaults.yaml"]["workflow"]...`. Otherwise it would conflict with the `workflow` section in `config["general"]`.

src/esm_runscripts/workflow.py
nwieters: Thank you. Good idea. I made a more general method out of it, to return a list of any attribute for all phases (default or user phases).

src/esm_runscripts/workflow.py
nwieters: Hm, I haven't. And I thought I had already seen everything. And there's even Sharknado 2.

src/esm_runscripts/workflow.py
nwieters: I thought of some attributes that the default workflow phases don't need to have. Like 'script'. 
But maybe it is not necessary to have two classes for this. Maybe I merge it to one class.

src/esm_runscripts/workflow.py
nwieters: yes of course

src/esm_runscripts/workflow.py
nwieters: Great. Thank you very much for your review so far.

src/esm_plugin_manager/esm_plugin_manager.py
```diff
@@ -18,11 +18,32 @@ def read_recipe(recipe, additional_dict, needs_parse=True):
 
 
 def read_plugin_information(plugins_bare, recipe, needs_parse=True):
```
mandresm, line 22: I don't get this function at all haha. You seem to have understood it @nwieters, maybe you can clarify (again) to me the need of `esm_plugins.yaml`, one of the mysteries of ESM-Tools

src/esm_plugin_manager/esm_plugin_manager.py
```diff
@@ -117,11 +148,25 @@ def check_plugin_availability(plugins):
 
 
 def work_through_recipe(recipe, plugins, config):
+    """
+    Works through the esm_runscripts recipes and plugin recipes.
+
+    Parameters
+    ----------
+        recipe : dict            # What is in these two dictionaries? Where do the entries are comming from?
```
mandresm, line 158: It all comes from here:

https://github.com/esm-tools/esm_tools/blob/e6bada5540b473523a783715f6a93e5a9beb91fc/src/esm_runscripts/helpers.py#L62-L71

src/esm_runscripts/batch_system.py
```diff
@@ -374,6 +374,7 @@ def get_run_commands(config, subjob, batch_or_shell):  # here or in compute.py?
```
mandresm, line 374: We should change `subjob` to `phase` (line above) if that's the new term we are using everywhere in the code, but I would leave these type of changes for when we are cleaning up.

src/esm_runscripts/batch_system.py
```diff
@@ -516,7 +518,7 @@ def write_simple_runscript(config, cluster, batch_or_shell="batch"):
             #    dummy = 0
             else:  # "normal" case
                 dummy = 0
-
+# was macht das hier? wo/wie wird submits_abother_job definiert?
```
mandresm: This was the exception we were talking about yesterday: each phase calls the next phase. If the current phase is running a script or a program the current phase is going to be actually a `sbatch` script:

```bash
#SBATCH --job=my_current_phase
#SBATCH my other sbatch settings

./my_script with input &
#or if it is a parallel program
srun <arguments> <parallel_program> &

# Gets the PID of the process launch right above so that it can be passed to esm_runscripts. Observe phase is going to be looking at the PID and if it doesn't exist anymore it means the process above is finished
process=$!

esm_runscripts <runscript>.yaml -e <expid> -t observe_<cluster_name> -p <PID> -s <date?> -r <run_number?> -v --last-jobtype <previous_job?phase?>
```

esm_runscripts command will run then concurrently the phase `observe`. Observe checks logs for errors and takes care of killing the slurm job if something goes wrong. It also takes care of printing the logs separately and merging them all at the end. Ultimately observe launches the next phase. The next phase will run in the current slurm job, unless it is a "slurm" phase in which case it creates a new sbatch script and submits it.

src/esm_runscripts/inspect.py
```diff
@@ -17,8 +17,8 @@ def run_job(config):
 
 def inspect_workflow(config):
     if config["general"]["inspect"] == "workflow":
-
-        config = workflow.display_nicely(config)
+        config = workflow.display_workflow(config)
+#        config = workflow.display_nicely(config)
```
mandresm:
```suggestion
        config = workflow.display_workflow(config)
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+        phases : list
+            List of workflow phases names
+        always_run_with : list
+            List of phases that precedes each phase in phases
+
+        Returns
+        -------
+        none
+        """
+        self.user_phases = []
```
mandresm: I don't see the need for `user_phases`. Maybe I'll see it below, but this is against the design principle of ESM-Tools: load a default and next file in the hierarchy overrides the default.

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+        self.first_task_in_queue = None                     # needed
+        self.last_task_in_queue = None
+        self.next_run_triggered_by = None
+        # TODO: Call here the phase object ???
+        self.phases = phases
+        self.always_run_with = always_run_with
+        self.phases_to_submit_to_batch_system = phases_to_submit_to_batch_system
```
mandresm: Is this necessary? This is information that belong to the particular phase itself, so for example:

```yaml
workflow:
    phases:
        compute:
            submit_to_batch_system: True
            [ ... ]
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+            self : class Workflow
+            phase_name : str (name of the phase to be returned
+
+        Returns
+        -------
+            phase : class phase or user_phase
+        """
+        for phase in self.phases + self.user_phases:
+            if phase.name == phase_name:
+                return phase
```
mandresm: If we unify `phases` and `user_phases` this function is not needed:
```python
phases[phase_name]
``` 
would do the job.

If we do not unify them, then the trick to do this is:
```python
all_phases = self.phases + self.user_phases
all_phases["phase_name"]
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+
+    def get_phases_attribs_list(self, phase_type, attrib):
+        """
+        Returns a certain attribute for all phases as a list.
+
+        Parameters
+        ----------
+            self: class Workflow
```
mandresm:
```suggestion
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+            phase_type: str (default or user)
```
mandresm:
```suggestion
            phase_type : str
                ``default`` or ``user``
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+            attrib: str
```
mandresm:
```suggestion
            attrib : str
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+        return phases_attribs
+
+    def init_default_workflow(self, config):
+        """
+        Add workflow for precompute, compute, and tidy phases
+        etc information already here!
+
+        Parameters
+        ----------
+            self : Workflow object
```
mandresm:
```suggestion
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+            config : dict
+
+        Returns
+        -------
+            self : Workflow object
+        """
+
+        workflow_phases = self.phases
+
```
mandresm:
```suggestion
```

Not used

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+        # Calculating the number of tasks for each component/model
```
mandresm:
```suggestion
        # Calculating the number of mpi tasks for each component/model/script
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+        # needed for phase compute
```
mandresm:
```suggestion
```

src/esm_runscripts/workflow.py
```diff
@@ -36,253 +616,154 @@ def skip_cluster(cluster, config):
...
+        self.run_on_queue = None
+        self.cluster = None
+        self.next_submit = []                       # needed
+        self.called_from = None                     # needed
+        self.batch_or_shell = "SimulationSetup"     # needed
+        self.order_in_cluster = "sequential"        # needed ???
+        self.run_only = None
+        self.skip_chunk_number = None
+        self.skip_run_number = None
+        self.name = phase_name
```
mandresm: I would make this a subclass of a dict. In this way you can initialize it simply with `WorkflowPhase("my_dict_that_comes_from_a_yaml")`

```python
class WorkflowPhase:
    """A workflow phase class."""

    def __init__(self, phase_name):
        self["name] = None
        self["nproc"] = 1                              # needed
        self["run_before"] = None
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+        """
+
+        workflow_phases = self.phases
+
+        # Calculating the number of tasks for each component/model
+        # needed for phase compute
+        tasks = calc_number_of_tasks(config)
+        # Initiate/create default workflow phase objects
+        # and reset/append to Workflow.phases variable
+        self.phases = []
```
mandresm:
```suggestion
```

I would remove this. What if I want to initialize it, do something with it and then it gets here removed, when I expected to be appended? The initialization of this already happened in `__init__`

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+        for ind, phase in enumerate(workflow_phases):
```
mandresm:
```suggestion
        for ind, phase in enumerate(self.phases):
```

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+                phase.run_before = self.phases[0].name
+                phase.next_submit.append(self.phases[0].name)
+                phase.run_after = self.phases[ind-1].name
+
+            phase.cluster = phase.name
+            if phase.name in self.phases_to_submit_to_batch_system:
+                phase.batch_or_shell = 'batch'
+                phase.submit_to_batch_system = True
+                phase.run_on_queue = config["computer"]["partitions"]["compute"]["name"]
+                phase.nproc = tasks
```
mandresm: I suggest to delete this thing here and added as a dictionary in `configs/esm_software/esm_runscripts/defaults.yaml`.

IMO there is no reason to define the default workflow differently than the "user" workflow, it's better do define the default workflow somewhere else. There might be no need to make the individual phases be an object other than a dictionary. If you decide to make them an object (because for example you need to check certain things when you initialize the phase) then I'd say make it a subclass of dictionary.

src/esm_runscripts/workflow.py
```diff
@@ -1,8 +1,588 @@
...
+
+        Returns
+        -------
+            None
+        """
+
+        if type(getattr(self, attrib)).__name__ == "list":
+            self.__dict__[attrib].append(value)
+        else:
+            self.__setattr__(attrib, value)
```
mandresm: It can all be defined in `configs/esm_software/esm_runscripts/defaults.yaml` phase by phase as a dict.

One of the things that could be a useful check here is to see if the default workflow provided in the default contains this variables. If not trigger an error.

```suggestion
```

src/esm_runscripts/workflow.py
mandresm: This class will only be needed if you use it to check the "mandatory" variables within, so that you can add a method that whenever the dictionary is modified it checks whether all the "mandatory" are available. Please, don't add this method yet. Let's leave it for later (by not resolving this comment), because it can take a bit of time to add such a method

src/esm_runscripts/workflow.py
```diff
@@ -36,253 +616,154 @@ def skip_cluster(cluster, config):
...
+
+        Returns
+        -------
+            true or false
+        """
+
+        if hasattr(self, keyword):
+            return True
+        else:
+            return False
```
mandresm: This could be done in one line:
```python
return hasattr(self, keyword)
```

src/esm_runscripts/workflow.py
```diff
@@ -36,253 +616,154 @@ def skip_cluster(cluster, config):
...
+    def __init__(self, phase_name):
+
+        WorkflowPhase.__init__(self, phase_name)
+
+        self.script = None
+        self.script_dir = None
+        self.call_function = None
+        self.env_preparation = None
+        self.batch_or_shell = "shell"
+        self.submit_to_batch_system = False
```
mandresm: IMO we don't need a UserWorkflowPhase, the user should be able to simply append to our default.

src/esm_runscripts/workflow.py
mandresm: When `WorkflowPhase` is a dictionary subclass this is not needed anymore. You will be able to do at any time:
```python
if "env_preparation" in phase
```

mandresm (changes requested):
Hi @nwieters,

Thanks for your efforts. At this point I have reviewed quite a lot already. I'll stop for now as I am suggesting a major change in the way the default workflow is described -> in a yaml file

src/esm_runscripts/batch_system.py
nwieters: Oops. Sorry for the german comment.

src/esm_runscripts/batch_system.py
nwieters: Yes. I have this already as a task in the project.

src/esm_runscripts/workflow.py
nwieters: No. Only if `self.phases` and `self.user_phases` were lists of dictionaries (that have keys with `<phase_name>` in it). But for the code you reviewed, this was not the case. These were lists of phase objects with certain attributes (e.g. name).
`all_phases[phase_name]`

src/esm_runscripts/workflow.py
nwieters: This is to separate the default phases (prepcompute, compute, tidy) from any other phase defined in any other yaml file.
Phases defined in setup/components/runscripts yamls can still overwrite each other. But they are not allowed to overwrite the default phases (prepcompute, compute, tidy).

src/esm_runscripts/workflow.py
nwieters: I put a check of valid keys in the workflow phase initialization. No method yet.

nwieters:
I hope that these were the last conceptual changes to the code. I will now start to do a cleanup and do some code optimizations.

To run only the tests for examples, run
`pytest test_workflow.py -m example --disable-warnings`

PR #1044, opened 2023-09-06
allow more flexibility in oasis restart file names

seb-wahl:
Hi all
as we use different naming conventions in FOCI for oasis restart files (to be compatible with our old runtime environment):
```yaml
    restart_in_sources:
         sstocean: sstocean_${parent_expid}_${echam.parent_date!syear!smonth!sday}
```
the current implementation in `oasis.py` fails for us. I suggest to use my more flexible approach which still enforces a restart file ending on the end date of the previous leg.

### DIFF
--- a/src/esm_runscripts/oasis.py
+++ b/src/esm_runscripts/oasis.py
@@ -497,7 +497,7 @@ class oasis:
                 # check if restart file with ini_restart_date in filename is in the restart
                 # folder of the parent experiment to be branched off from:
                 glob_search_file = (
-                    f"{restart_file_path}_????????-"
+                    f"{restart_file_path}*"
                     f"{config['ini_restart_date'].year}"
                     f"{config['ini_restart_date'].month:02}"
                     f"{config['ini_restart_date'].day:02}"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.24.1
+current_version = 6.24.2
 commit = True
 tag = True
 
--- a/setup.py
+++ b/setup.py
@@ -103,6 +103,6 @@ setup(
     test_suite="tests",
     tests_require=test_requirements,
     url="https://github.com/esm-tools/esm_tools",
-    version="6.24.1",
+    version="6.24.2",
     zip_safe=False,
 )
--- a/src/esm_archiving/__init__.py
+++ b/src/esm_archiving/__init__.py
@@ -4,7 +4,7 @@
 
 __author__ = """Paul Gierz"""
 __email__ = "pgierz@awi.de"
-__version__ = "6.24.1"
+__version__ = "6.24.2"
 
 from .esm_archiving import (archive_mistral, check_tar_lists,
                             delete_original_data, determine_datestamp_location,
--- a/src/esm_calendar/__init__.py
+++ b/src/esm_calendar/__init__.py
@@ -2,6 +2,6 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = "dirk.barbi@awi.de"
-__version__ = "6.24.1"
+__version__ = "6.24.2"
 
 from .esm_calendar import *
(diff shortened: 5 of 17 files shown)


src/esm_runscripts/oasis.py
```diff
@@ -497,7 +497,7 @@ def add_restart_files(self, restart_file_label, fconfig):
                 # check if restart file with ini_restart_date in filename is in the restart
                 # folder of the parent experiment to be branched off from:
                 glob_search_file = (
-                    f"{restart_file_path}_????????-"
+                    f"{restart_file_path}*"
```
mandresm, line 500: What about the following? Then we are enforcing 2 dates format.
```suggestion
                    f"{restart_file_path}*_????????-"
```

joakimkjellsson: Yes that would work, but I'm assuming the last minus sign is a typo? 
```python
f"{restart_file_path}*_????????"
```
rather than
```python
f"{restart_file_path}*_????????-"
```

seb-wahl: Yes, the minus and the end must be a type :-)
And yes enforcing an 8 digit (YYYYMMDD) end date is probably better, but it assumes that people use YYYMMDD as date format.

nwieters: Hi,
the minus is not a typo, because the whole filename of the restart file is composed like this:
```
glob_search_file = (
    f"{restart_file_path}_????????-"
    f"{config['ini_restart_date'].year}"
    f"{config['ini_restart_date'].month:02}"
    f"{config['ini_restart_date'].day:02}"
)
```
It searches for a file with this pattern: `<restart_file_path>_????????-20000101`
For example:
`rstas.nc_20000101-20000131`

This pattern can be made more general for the restart file name as mentioned by @mandresm and @seb-wahl 
```
f"{restart_file_path}*_????????-"
f"{config['ini_restart_date'].year}"
f"{config['ini_restart_date'].month:02}"
f"{config['ini_restart_date'].day:02}"
```

This pattern is taken to search for the right restart file. If it finds more then one, the user will be asked to select one.

mandresm:
I've just pushed my suggested change. I'll test a couple of things in DKRZ with this change and it works I'll merge :)

mandresm:
Closes #1049

seb-wahl:
The suggested fix by @mandresm doesn't work for us as our restart file pattern is 
```yaml
restart_in_sources:
         sstocean: sstocean_${parent_expid}_${echam.parent_date!syear!smonth!sday}
```
so the `-` add the end will make sure it fails for us :-)
Can't we use `f"{restart_file_path}*_????????"` or `f"{restart_file_path}*_????????*"` instead of `f"{restart_file_path}*_????????-"`

nwieters:
But you have to look at the whole file pattern, that is looked for
```
glob_search_file = (
    f"{restart_file_path}*_????????-"
    f"{config['ini_restart_date'].year}"
    f"{config['ini_restart_date'].month:02}"
    f"{config['ini_restart_date'].day:02}"
)
```

This will result in something like restartfile_20000101-20000201
Replacing `f"{restart_file_path}*_????????-"`  by `f"{restart_file_path}*_????????"` will not solve your issue.

The idea here is to look for a certain restart file with a specific date at the end of the filename when you do a branchoff experiment. Thats why there are the lines 
```
f"{config['ini_restart_date'].year}"
f"{config['ini_restart_date'].month:02}"
f"{config['ini_restart_date'].day:02}"
```
You probably need something like 
```
glob_search_file = (
    f"{restart_file_path}*"
    f"{config['ini_restart_date'].year}"
    f"{config['ini_restart_date'].month:02}"
    f"{config['ini_restart_date'].day:02}"
)
```

mandresm:
Sorry @seb-wahl, you're right. I'm reverting the changes and testing AWICM3 in Levante.

mandresm:
#bump

seb-wahl:
thanks for taking care of this. Now I'm happy :-)

mandresm (approved):

mandresm:
> thanks for taking care of this. Now I'm happy :-)

So easy to make you happy! :D

PR #1059, opened 2023-09-26
fix building error when using esm_master with generic.yaml machine file

mandresm:


nwieters (approved):
I am sure this is working, but I can't reproduce neither the error nor the solution. When I try to run esm_master on my local comupter I get a warning from the parser:
```
WARNING:root:The yaml file for this computer (machinexyz) could not be determined!
WARNING:root:Continuing with generic settings...
```
and then it will break trying to compile it.

Just for me to follow this, can you describe how the error looked like and how this solves it? Thank you.

mandresm:
Sorry for the no description: esm_master will break if you try esm_master get, or use the -c flag.

mandresm:
Sorry, I fogot about bumping but I did it manually in release...

PR #1056, opened 2023-09-20
Feat/awicm3.2

JanStreffing:
This merge brings AWI-CM3 v3.2 (https://awi-cm3-documentation.readthedocs.io/en/latest/releases.html#awi-cm3-v3-2) to the esm_tools release branch

PR #1063, opened 2023-09-30
fix/recursive resolution of early variables and fix find_key

mandresm:
This PR fixes the following problem affecting the version switching of FESOM in the PR by @JanStreffing #1056:

In *awicm3.yaml*:

```yaml
fesom:
    version: "2.0"
    choose_general.major_version:
        "3.2":
            version: "2.5"
```

When `general.major_version: "3.2"` the `version` was still `"2.0"` because of several bugs in `esm_parser`.

@seb-wahl  I think you have encounter this problem on the past.

To solve the bugs the following changes have been carried out:

- add recursion to resolve_choose_with_var so that variables that depend on other choose_ blocks can be resolved in early stages if needed.
- fix an issue with find_key to always continue searching into dictionaries to avoid returning only the first match within a path.
- add the early resolution of choose blocks containing component versions in early stages so that the correct yaml file of the component can be loaded.
- improve and clean the code for filtering of found paths to search for chooses containing the desired variable.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.24.3
+current_version = 6.24.4
 commit = True
 tag = True
 
--- a/setup.py
+++ b/setup.py
@@ -103,6 +103,6 @@ setup(
     test_suite="tests",
     tests_require=test_requirements,
     url="https://github.com/esm-tools/esm_tools",
-    version="6.24.3",
+    version="6.24.4",
     zip_safe=False,
 )
--- a/src/esm_archiving/__init__.py
+++ b/src/esm_archiving/__init__.py
@@ -4,7 +4,7 @@
 
 __author__ = """Paul Gierz"""
 __email__ = "pgierz@awi.de"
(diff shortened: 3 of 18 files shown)


JanStreffing:
Thank you very much @mandresm. I'll get around to testing this in 2 weeks. If someone else wants to give it ago before then, by all means: go :)

nwieters:
I can't really do a full test. When I try to create this situation by adding a fesom version 2.5 into awicm3 v3.1 choose block (in awicm3.yaml), I get an error because of missing coupling fields.
But I debugged it, and I see the fesom version 2.5 in the config dict. I also think that I would not get this error message if the 'correct' version had not been taken. (in this case 2.5).

nwieters (approved):

mandresm:
Differences with the older tests are minimal and should not create any problem, therefore, #approve-changes

mandresm:
#bump

mandresm:
#bump

mandresm:
#bump

PR #1075, opened 2023-10-19
Implement tar_binary_restarts plugin into awicm3 and fesom-2.5.

### LINKED ISSUE #1072: Packaging of folders before copying/moving/linking
### Discussed in https://github.com/esm-tools/esm_tools/discussions/1062

<div type='discussions-op-text'>

<sup>Originally posted by **JanStreffing** September 28, 2023</sup>
I recently started an AWI-CM3 production run on aleph without netcdf FESOM2 restarts. Only binary. Normally we move the binary restart files from the old to the new run folder at the end of a run. It's not viable to copy them out first, because it's too many files. (because time to copy is long and number of files goes towards file limit).

I tried a parallel packing like this: `tar cf - np1280 | pigz -p 30 > archive.tar.gz`. This was quite fast. We could then move this archive out and move the unpacked files to the new work folder.

How would this best be done? plugin script, modification of python backend?</div>

nwieters:
Closes #1072

### DIFF
--- a/configs/components/fesom/fesom-2.5.yaml
+++ b/configs/components/fesom/fesom-2.5.yaml
@@ -169,8 +169,6 @@ restart_out_files:
         fesom_raw_restart: fesom_raw_restart
         fesom_bin_restart_info: fesom_bin_restart_info
         fesom_bin_restart: fesom_bin_restart
-        fesom_bin_restart_tar: fesom_bin_restart_tar
-        fesom_raw_restart_tar: fesom_raw_restart_tar
 
 restart_out_in_work:
         par_oce_restart: fesom.${end_date!syear}.oce.restart/*.nc
@@ -292,6 +290,8 @@ file_movements:
                 all_directions: move
         fesom_raw_restart_info_out:
                 all_directions: move
+        fesom_raw_restart_tar_out:
+                all_directions: move
         fesom_bin_restart_in:
                 all_directions: move
         fesom_bin_restart_out:
@@ -300,6 +300,13 @@ file_movements:
                 all_directions: move
         fesom_bin_restart_info_out:
                 all_directions: move
+        fesom_bin_restart_tar_out:
+                all_directions: move
+
+choose_tar_binary_restarts:
+    True:
+        add_restart_out_files:
+            fesom_raw_restart_tar: fesom_raw_restart_tar
 
 # Is it a branchoff experiment?
 branchoff: "$(( ${lresume} and ${general.run_number}==1 ))"
@@ -565,6 +572,9 @@ grids:
 further_reading:
         - fesom/fesom.forcing.yaml
 
+# ------------------------
+# RECIPE
+# ------------------------
 tidy_recipe:
 - "tidy_coupler"
 - "tar_binary_restarts"
--- a/configs/components/fesom/fesom-2.5.yaml
+++ b/configs/components/fesom/fesom-2.5.yaml
@@ -19,6 +19,9 @@ icb_code: False
 with_wiso: false
 wiso_code: false
 
+required_plugins:
+- "git+https://github.com/esm-tools-plugins/tar_binary_restarts"
+
 choose_version:
   '2.5':
     branch: "2.5"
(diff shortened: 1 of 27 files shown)


JanStreffing:
Hello @nwieters, thanks for the feature. I would suggest merging into release instead after #1056 is merged.

configs/components/fesom/fesom-2.5.yaml
```diff
@@ -554,3 +557,20 @@ grids:
...
+- "throw_away_some_infiles"
+- "copy_stuff_back_from_work"
+- "copy_all_results_to_exp"
+#- "_update_chunk_date_file"
+- "clean_run_dir"
+#- "start_post_job"
+#- "signal_tidy_completion"
+#- "_increment_date_and_run_number"
+#- "_write_date_file"
+#- "maybe_resubmit"
```
mandresm:
```suggestion
- "clean_run_dir"
```
This are remains of an older version of ESM-Tools

mandresm (changes requested):
The tests are broken. My guess is that all versions of fesom2 yamls need the same changes as fesom-2.5.yaml. Otherwise when awicm3 versions <3.2 tries to call the `fesom.tidy_recipe` it won't be found

configs/setups/awicm3/awicm3.yaml
```diff
@@ -105,6 +105,7 @@ general:
 
         # Load compute recipe from oifs.yaml
         prepcompute_recipe: ${oifs.prepcompute_recipe}
+        tidy_recipe: ${fesom.tidy_recipe}
```
nwieters, line 108: Wouldn't it be possible to integrate the changes of a recipe here in a choose block? So that the changed recipe is actually only taken if necessary?
In other words: If I don't want to use the plugin `tar_binary_restarts` that will be called in the changed recipe, I don't need a changed recipe here.
As I understand it now, everybody using AWICM3 needs/will install this plugin because of a changed recipe, but may never use it. Or do I have a thinking error here?  

Something like:
```
tar_bin = fesom.tar_binary_restart
choose_tar_bin:
    "true":
        tidy_recipe: ${fesom.tidy_recipe}
```

mandresm: >  As I understand it now, everybody using AWICM3 needs/will install this plugin because of a changed recipe, but may never use it. Or do I have a thinking error here?

With these lines
https://github.com/esm-tools/esm_tools/pull/1075/files#diff-455b0ceb6e69ed7e3a2b286473f7b9b0a0c99a59892b0d5feee4f87f6d60d53cR12-R14
you are saying "everyone that installs AWICM3 will also install this plugin"

That should remain, cause you don't have a way to tell esm_master that you need or not tar bin plugin.

The following is another way of doing things, you either put the switch within the yamls, or in the plugin. It's the same essentially:
```diff
-tar_bin = fesom.tar_binary_restart
-choose_tar_bin:
+choose_tar_binary_restarts:
    "true":
        tidy_recipe: ${fesom.tidy_recipe}
```

mandresm: That being said, I would leave it as it currently is. It's cleaner and the if is in the plugging itself

runscripts/awicm3/v3.2/awicm3-v3.2-levante-TCO95L91-CORE2.yaml
```diff
@@ -5,10 +5,10 @@ general:
     account: "ab0246"
     compute_time: "00:20:00"
     initial_date: "2000-01-01"
-    final_date: "2000-02-01"
+    final_date: "2000-01-02"
```
mandresm, line 8: Hey @JanStreffing, are you sure you want all this changes in your runscript? Daily instead of monthly?

nwieters: Hi @JanStreffing , @mandresm ,
are you still doing changes on this PR or is it still in review? Or can we merge it?

mandresm:
@nwieters, I've added some changes to it to allow for automatically installing missing plugins while using esm_runscripts. This is commonly encountered by people using an old compiled model with an updated version of ESM-Tools that requires a new plugin. This happens because in the past only `esm_master` was able to automatically install missing plugins, but now both can do so.

Nothing is required from your side, unless you want to give a review and a look at what I did :)

nwieters:
> @nwieters, I've added some changes to it to allow for automatically installing missing plugins while using esm_runscripts. This is commonly encountered by people using an old compiled model with an updated version of ESM-Tools that requires a new plugin. This happens because in the past only `esm_master` was able to automatically install missing plugins, but now both can do so.
> 
> Nothing is required from your side, unless you want to give a review and a look at what I did :)

Thank you @mandresm . I had a look at your last commits. From my side it looks good. I would approve. I just realized I can not approve my own PR. :-)

mandresm (approved):

mandresm:
#approve-changes

mandresm:
#bump

PR #1078, opened 2023-10-20
fix/Provide notification flags for PBS as list.

### LINKED ISSUE #1077: email notification on PBS
esm tools will join together mail1 and mail2 in pbs.yaml containing the mail type and mail address into a single notification_flag. This is ok for slurm, where the email notification flag is a one-liner. But for PBS it needs to go into two separate lines, like so:

```
#PBS -m abe
#PBS -M my_email@umn.edu
```
Thus we need the two separate variables to make it into batch_system.py. At least in the case that we use PBS.

Currently we get:
```
#PBS -m abe -M my_email@umn.edu
```

nwieters:
For job scheduler PBS, the notification flags provided like this
```
#PBS -m abe
#PBS -M my_email@umn.edu
```
In this PR the notification_flags will be set as list.

closes #1077

### DIFF
--- a/configs/other_software/batch_system/pbs.yaml
+++ b/configs/other_software/batch_system/pbs.yaml
@@ -33,7 +33,10 @@ choose_general.mail_user:
         "*":
                 mail2:  "-M ${general.mail_user}"
 
-notification_flag: "${mail1} ${mail2}"
+notification_flag: 
+    - "${mail1}"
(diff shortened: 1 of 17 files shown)


mandresm (approved):

mandresm:
#approve-changes

nwieters:
#bump

PR #1092, opened 2023-11-07
Fix/early further reading and env changes

JanStreffing:
small fix to evaluation two important variables early


JanStreffing:
what the... this did not go according to plan

PR #1101, opened 2023-11-20
Feature/oasis from cerfacs

joakimkjellsson:
I think I'm done with adding the CERFACS branch of MCT5 now. 

Main changes: 
* New oasis3mct version: "5.0-geomar" (we can certainly rename it in the future)
* New method to compile and clean, using `make -f TopMakefileOasis3`
* Small changes to `src/esm_runscripts/oasis.py` since `NLOGPRT` has different format

Works for FOCI-OpenIFS on glogin. 

Awaiting your approval!
/J

### DIFF
--- a/configs/components/nemo/nemo.yaml
+++ b/configs/components/nemo/nemo.yaml
@@ -1451,8 +1451,12 @@ choose_generation:
                             namdyn_hpg:
                                ln_hpg_sco: '.true.'
         "3.6":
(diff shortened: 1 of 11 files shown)


mandresm (approved):
All none FOCI/NEMO related changes look good, I don't see anything it would break other models or OASIS

mandresm:
Hi @joakimkjellsson, you can go ahead and merge whenever you need to if @seb-wahl is okay with that.

PR #1105, opened 2023-12-14
fix/perl levante

mandresm:
use the perl installed by spack in levante

Due to some changes in the perl libraries from levante, OpenIFS is not compiling anymore.

This PR will solbe the problem once DKRZ installs all the packages needed for perl in their spack version of perl.


mandresm:
Hi @joakimkjellsson and @seb-wahl, if you are wondering why suddenly openIFS is not compiling in Levante, @JanStreffing and myself found the culprit.

mandresm:
DKRZ solved the problem with the default perl library. This PR is not needed anymore.

PR #1106, opened 2023-12-14
feat/add jobid to name of logfiles

### LINKED ISSUE #842: all logfiles should contain the jobid in their filename
In an `awicm-1.0-recom` simulation, 5 logfiles are created in the `log/` directory per run:
```
<expid>_awicm_compute_<from>-<to>_<jobid>.log
<expid>_awicm_observe_compute_<from>-<to>.log
<expid>_awicm_prepcompute_<from>-<to>.log
<expid>_awicm_tidy_<from>-<to>.log
<expid>_esm_runscripts_<from>-<to>.log
```
Only one of them contains the jobid. If a specific date is computed more than once, the logfiles without the jobid in their filenames will be overwritten. 

All logfiles should contain the jobid in their filename.

Thanks and cheers,
Chris

nwieters:
This PR will add the jobtype and jobid to the filename of all logfiles.

closes #842

### DIFF
--- a/src/esm_runscripts/compute.py
+++ b/src/esm_runscripts/compute.py
@@ -282,11 +282,13 @@ def initialize_experiment_logfile(config):
     if "trace_sink" in dir(logger):
         experiment_dir = config["general"]["experiment_dir"]
         expid = config["general"]["expid"]
+        jobid = config["general"].get("jobid", None)
+        jobid = f"_{jobid}" if jobid else ""
         it_coupled_model_name = config["general"]["iterative_coupled_model"]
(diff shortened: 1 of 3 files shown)


nwieters:
I think the function `initialize_experiment_logfile` in `compute.py` is not used at all.

nwieters:
There will be now a `run_test_issue_842_esm_runscripts_<cluster>_<drom_to>_<jobid>.log` for each cluster.
I think before, they were always overwritten by the logfile creation (open file with `w` instead of `a`).

mandresm:
**TODO**
- [ ] Decide if we really want this feature

PR #1107, opened 2023-12-15
Fix/list in user note

mandresm:
If you provide a `note_text` to `user_note` that is a list, that breaks `user_note`. This PR fixes the problem.


nwieters (approved):

mandresm:
#bump

PR #1108, opened 2023-12-15
fix/1053_esm_runscripts_argument-start_date

### LINKED ISSUE #1053: esm_runscripts argument -s (--start_date) is not evaluated
**Describe the bug**
The argument `-s` (or `--start_date`) given to `esm_runscripts` is not evaluated.

**To Reproduce**
Run
`esm_runscripts awicm3-v3.1-levante-TCO95L91-CORE2_initial_test_WFM.yaml -e run_test_WFM_005 -t observe_compute -s 20000101 -r 1 -v  --last-jobtype prepcompute --open-run`

The argument `-s` is ignored and the date is evaluated from file `<expid>_<setup>.date` in `scripts` folder of the experiment.

**Expected behavior**
If this argument is given, the `current_date` should be set/overwritten by this argument and not from the `<expid>_<setup>.date` file.

**Screenshots**

 - Levante / Any
 - Version 6.24.1

**Additional context**

nwieters:
This PR will read in the current date from command line argument -s (--start_date) if given instead of reading it from the date file.

closes #1053 

New functionality:
- Will read the `current_date` and `run_number` from the argument given by `-s` and `-r`
- Will read `*.date` file if no argument for `current_date` (`-s`) is given
- If `current_date` is given by argument `-s` and is different compared to the date in `*.date` file, it will ask the user to choose which date should be taken.

But:
- `current_date` date and `run_number` needs to be set correctly by user if given as arguments `-s` and `-r`
    - [ ] Check if there is a log file logging the date together with run_number
- When is `*.date` file written? In `maybe_resubmit`
- Can a single jobtype be executed on its wown? E.g. only tidy? Or will it continue the simulation because of current resubmit mechanism?

### DIFF
--- a/src/esm_runscripts/prepare.py
+++ b/src/esm_runscripts/prepare.py
@@ -90,17 +90,25 @@ def _read_date_file(config):
                 f"Please confirm if you want to continue:"
             )
             esm_parser.user_note("Detached experiment:", msg)
-            proceed = True
+            proceed = ""
             if isresubmitted:
-                proceed = questionary.confirm(
-                    "Do you want to continue?"
-                ).ask()
-            if proceed:
-                date = str(date_c)
-                run_number = run_number_c
-            else:
-                esm_parser.user_note("The experiment will be cancelled:", f"You cancelled the experiment due to date discrepancies.")
-                sys.exit(1)
+                proceed = questionary.select(
+                    "Do you want to continue?",
+                    choices=[
+                        f"Yes, with date from command line argument: {str(config['general']['current_date'])}",
+                        f"Yes, with date from date file: {date}",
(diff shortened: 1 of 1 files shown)


nwieters:
This is not ready yet, because I want to let the user confirm that the current date should be taken from the argument instead of read in from date file. 
The problem I see here is that the experiment will be in some kine of detached state, when using a different date. But on the other hand, if you give -s as argument, you should know what you do. But I think a user note and maybe also a confirmation is good to have.

mandresm:
Work in progress, to be continued/reviewed by @mandresm

**TODO**
- [ ] File-registry mapping dates to runs
- [ ] Run number supported as well

PR #1109, opened 2023-12-18
fix/direnv get_real_path_from_file

mandresm:
fix a bug on _get_real_dir_from_pth_file not getting the correct paths for direnv

The problem was that even when you run within a direnv, if you had installed a general esm_tools, the binaries used would be the ones from the direnv, but the configuration files would be the ones from the general esm_tools installation.

This is fixed now

### DIFF
--- a/src/esm_tools/__init__.py
+++ b/src/esm_tools/__init__.py
@@ -96,6 +96,18 @@ def _get_real_dir_from_pth_file(subfolder):
         [],
     )
     logger.debug(site_packages_dirs)
+
+    # Check if the user is running with direnv and removes all other sites
+    direnv = os.getenv("VIRTUAL_ENV")
+    if direnv:
+        logger.debug(f"User running with direnv, removing other sites")
+        site_package_dirs_new = []
+        for site_package_dir in site_packages_dirs:
+            if direnv in site_package_dir:
+                site_package_dirs_new.append(site_package_dir)
+        site_packages_dirs = site_package_dirs_new
+        logger.debug(site_packages_dirs)
+
     for site_package_dir in site_packages_dirs:
         logger.debug(f"Working on {site_package_dir}")
         # Read the pth file:
@@ -374,6 +386,7 @@ def get_config_filepath(config=""):
         cpath = _get_config_filepath_editable_install(config)
     else:
         cpath = _get_config_filepath_standard_install(config)
+    print(cpath)
     return cpath
 
 
--- a/src/esm_tools/__init__.py
+++ b/src/esm_tools/__init__.py
@@ -23,7 +23,7 @@ so it's just the dictionary representation of the YAML.
 
 __author__ = """Dirk Barbi, Paul Gierz"""
 __email__ = "dirk.barbi@awi.de"
-__version__ = "6.25.7"
+__version__ = "6.25.8"
 
 import functools
 import inspect
--- a/src/esm_tools/__init__.py
+++ b/src/esm_tools/__init__.py
@@ -386,7 +386,7 @@ def get_config_filepath(config=""):
         cpath = _get_config_filepath_editable_install(config)
     else:
         cpath = _get_config_filepath_standard_install(config)
-    print(cpath)
+    logger.debug(cpath)
     return cpath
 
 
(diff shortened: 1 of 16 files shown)


nwieters (approved):
On Albedo I also had to add the following line 
eval "$(direnv hook bash)"
into the .bash_profile (instead or in addition to .bashrc) in order to get the right environment described in .envrc and the right esm_tools installation.
I added this to the documentation, see PR #1112

nwieters:
One question to direnv: I tested to run a simulation within a direnv environment and with --contained-run mode and in editable mode. But it seems not to work. There is no esm-tools installation in the experiment path. When I test this outside a direnv it works.

Should an experiment in a virtualenv from direnv work?

pgierz:
Emmm...it should work, I think. I can have a look at this later this morning.

Note though that this is a very obscure use-case. The venv and direnv effectively solve the same problem.

nwieters:
> Note though that this is a very obscure use-case. The venv and direnv effectively solve the same problem.

But there is the questionaire asking whether it should be installed in venv or not.

I was wondering which esm_tools version is installed if you run it from a direnv and want to also install it in a venv in the experiment. Shouldn't this be the version you used in the direnv environment?

Maybe I get it wrong: But I thought you use direnv to have a very specific esm_tools version and venv to have this version also installed within your experiment. Is that right?

mandresm:
Hey @pgierz! I'm taking care of checking that out since there is a user I promised I'll be merging this by 12:00

mandresm:
> I was wondering which esm_tools version is installed if you run it from a direnv and want to also install it in a venv in the experiment. Shouldn't this be the version you used in the direnv environment?

`venv` always installs release in the virtual environment unless you add to the runscript `install_esm_tools_branch: branch/tag`. The documentation for that is here: https://esm-tools.readthedocs.io/en/docs-direnv/esm_runscripts.html#running-an-experiment-with-a-virtual-environment

Note that you can also install in editable mode so that the source code is in the experiment and accessible, and the changes you make in that source code have an effect on the simulation.

> Maybe I get it wrong: But I thought you use direnv to have a very specific esm_tools version and venv to have this version also installed within your experiment. Is that right?

In essence yes, but at the same time, as Paul mentioned, they are two different ways of solving the same problem. The biggest difference is that direnv is more similar to your local installation, where you can make modifications on the fly and they make it into your simulation. venv in contrast is a more encapsulated approach, where you cannot modify easily the yamls and source code used (unless you add the `install_esm_tools_editable` variable into your runscript) and more difficult to use something that is not `release`. I think venv is more adequate for production as everything is contained in the runscript, while direnv is more error prone for production as it requires that the user is very careful with which folder she/he is working from, make sure they are not making changes during the simulation... etc. Unfortunatelly, some users prefer direnv for production using different ESM-Tools versions/branches because it is more intuitive to them.

pgierz:
> Hey @pgierz! I'm taking care of checking that out since there is a user I promised I'll be merging this by 12:00

OK, hands off from my side then 😉 

To the discussion which to choose regarding direnv/venv: I think this is a nice case for user education. @mandresm very nicely summarised it: they solve the same problem, but differently, and here are my two cents:

* If you want to have the **exact same** `esm-tools` every single time *and* have a record of this in your experiment, use `venv`. This case is probably what the researchers want most of the time.

* If you are actively developing and want to ensure your *global user copy* (whatever that might be, e.g. in `~/.local/bin/esm_runscripts`) is **not** used but instead a *local* version tied to collection of experiments (e.g. `/some/supercomputer/path/user/project/experiments/[A, B, C]` then, use `direnv`.

I hope that made sense...

src/esm_tools/__init__.py
```diff
@@ -374,6 +386,7 @@ def get_config_filepath(config=""):
         cpath = _get_config_filepath_editable_install(config)
     else:
         cpath = _get_config_filepath_standard_install(config)
+    print(cpath)
```
pgierz:
```suggestion
    logger.debug(cpath)
```

pgierz (changes requested):
Mini change: just use the `logger`.


> `# NOTE(PG): there is probably a better way of doing that than with assert.`


...I also love reading comments where it was obvious that I thought there was probably a better way of doing things, but instead just say "ehhh...ok, screw it, we are doing it somehow sloppily and hope that no one actually notices."

mandresm:
#bump

PR #1112, opened 2024-01-09
docs/direnv_pip

nwieters:
Describing the usage of direnv

Changes to uninstallation and pip installation.

You can view the version here:
https://esm-tools.readthedocs.io/en/docs-direnv/installation.html


docs/installation.rst
```diff
@@ -34,57 +34,79 @@ A possible (default) output can be ``~/.local/bin/esm_tools``.
...
+- Update ``pip`` and ``wheel`` python packages::
+    
+    $ pip install -U pip wheel
+    
+- Install `ESM-Tools`::
+    
+    $ pip install esm-tools 
+
+.. note::
+   The version that is available via ``pip`` is not the most recent version. We strongly recommend to install the most recent version of `ESM-Tools` using the install script described above (see also :ref:`get_esm-tools:Get ESM-Tools` and :ref:`installation:Installing in local environment`).
```
mandresm: I think this version doesn't work, or if it works, one cannot easily find the configuration files. Can you please check that that's the case? If yaml files are difficult to find, maybe we should consider removing this.

docs/installation.rst
```diff
@@ -118,21 +140,37 @@ If you installed in editable mode as described above, you can update `ESM-Tools`
...
+- Remove the installed Python packages::
+
+  $ rm -ri ~/.local/lib/python3.<version>/site-packages/esm*
+
+Note that you may have a different Python version, so the second command might need to be adapted.
+
+Using pip
+^^^^^^^^^
+
+Make sure you have the most recent version of
```
mandresm: Is there an specific reason why you are adding all of this? On the past we needed the `pip freeze | grep esm | xargs pip uninstall -y` because we had many independent esm packages but now uninstalling esm_tools it uninstalls everything: I've just tried `pip uninstall esm_tools` and it worked perfectly. The advantage of this is that is only one command and that is independent on whether you are using a conda environment, local environment or direnv.

mandresm (changes requested):
Thanks! See suggestions :)

docs/installation.rst
nwieters: No, I only restructured the section. The pip freeze | grep esm | xargs pip uninstall -y was already in the documentation. Can we than remove it?

mandresm: Maybe you can keep it but only for versions below 6.0.0. For versions >= 6.0.0 `pip uninstall esm_tools` should do.

nwieters: I added here a line for which version of esm-tools the pip uninstall command is appropriate for.

docs/installation.rst
nwieters: I haven't tested it so far but I will comment this section out for now. I will have a look at this version that is available via pip later.

mandresm (approved):
Hi @nwieters,

I'm approving this, however, can you please put the pip uninstallating before the manual one?

You can go ahead and merge afterwards, we don't bump the versions for documentation PRs.

PR #1115, opened 2024-01-12
generalization of the early resolution of chooses with early_choose_v…

mandresm:
This PR generalizes the early resolution of chooses with `early_choose_vars` and also the `further_reading` attachments to solve `further_reading` problems encounter by @JanStreffing while having `further_readings` in the `general` section of the coupled setup `awicm3`, under `choose_` blocks. This PR should also improve the overall functioning of the early_choose logic so now any variable included in `early_choose_vars` will actually be resolved early, without needing extra workarounds (unless it needs to be for real resolved very very early, like `include_models` or `version`).

### DIFF
--- a/src/esm_parser/esm_parser.py
+++ b/src/esm_parser/esm_parser.py
@@ -423,6 +423,14 @@ def attach_to_config_and_reduce_keyword(
                 config_to_read_from[reduced_keyword] = config_to_read_from[full_keyword]
         # FIXME: Does this only need to work for lists?
         if isinstance(config_to_read_from[full_keyword], list):
+
+            # Deduplicate items
+            new_conf_reduced_keyword = []
+            for elem in config_to_read_from[reduced_keyword]:
+                if elem not in new_conf_reduced_keyword:
+                    new_conf_reduced_keyword.append(elem)
+            config_to_read_from[reduced_keyword] = new_conf_reduced_keyword
+
             for item in config_to_read_from[full_keyword]:
                 model = item
                 if "-" in item:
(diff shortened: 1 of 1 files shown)


pgierz (approved):

src/esm_parser/esm_parser.py
```diff
@@ -2881,6 +2882,17 @@ def _config_init(self, user_config):
...
+    def is_coupled_setup(self):
+
+        if (
+            "general" in self.config
+            and isinstance(self.config["general"], dict)
+            and "coupled_setup" in self.config["general"]
+        ):
+            return True
+        else:
+            return False
```
pgierz, line 2894:
```suggestion
    @property
    def is_coupled_setup(self):

        return (
            "general" in self.config
            and isinstance(self.config["general"], dict)
            and "coupled_setup" in self.config["general"]
        )
```

src/esm_parser/esm_parser.py
```diff
@@ -2933,7 +2945,7 @@ def _config_init(self, user_config):
 
         # distribute self.config into setup_config
 
-        if "general" in self.config and "coupled_setup" in self.config["general"]:
+        if self.is_coupled_setup():
```
pgierz, line 2948:
```suggestion
        if self.is_coupled_setup:
```

pgierz (commented):
Some suggestions

pgierz:
The bot marked this as stale. @mandresm, is this still needed? It seems to have not been needed for a year, which surprises me...

PR #1118, opened 2024-01-15
fix/oasis_restarts_non_default_names

mandresm:
I was trying to fix FOCI automatic tests and I found that the problem was a bug on the OASIS restarts.

The most important change in this PR is the substitution of `restart_file_label` by `restart_file` in the definition of `restart_file` within the in. The rest is just making sure that if the user has defined their own names in the source or work directory those names are not used.

**TODO**

@seb-wahl 
- [ ] can you check that a test run of FOCI works for you?

@ESM-Tools team
- [ ] run standard tests and restarts
- [ ] run coldstarts with `oasis3mct.lresume: True` with AWICM3 (and diff files to make sure all is correct!)
- [ ] run AWICM3 with Tido's restart logic (in his runscripts) (and diff files to make sure all is correct!)

### DIFF
--- a/src/esm_runscripts/oasis.py
+++ b/src/esm_runscripts/oasis.py
@@ -461,23 +461,43 @@ class oasis:
         # get the basename of that path, otherwise assign the file label also as name
         # of the file (coming from ``coupling_<target/input>_fields``)
         if restart_file_path:
-            restart_file = os.path.basename(restart_file_label)
+            restart_file = os.path.basename(restart_file_path)
         else:
             restart_file = restart_file_label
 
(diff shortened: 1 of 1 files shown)


mandresm:
@seb-wahl, I'm getting the following while running FOCI:

```
3879  0: FATAL ERROR in InputFileOpen: File: population_density.nc does not exist and defaults are not available for all file variables
3880  0:  FINISH called from PE: 1
```

My test is based in one of your runscripts, that you provided some time ago. I'm testing in `blogin`. If you run a FOCI test please let me know about the outcome.

PR #1122, opened 2024-01-19
WIP adding oasis3mct6 with yac interpolation

JanStreffing:
So far only compile time works. We might need some new options in the python backend to support the YAC grammar for namcouple find.

e.g. this old namcouple:
```
LOCTRANS SCRIPR
INSTANT
CONSERV U SCALAR LATITUDE 500 FRACNNEI FIRST 1.45 -2.0
####
```
would be replaced by:
```
LOCTRANS YAC
INSTANT
GC GC 2 YAC_CONSERV_FRACNNEI 4
CONSERV FIRST FRACAREA
NNN AVG 1
#######
```

### DIFF
--- a/configs/components/oasis3mct/oasis3mct.yaml
+++ b/configs/components/oasis3mct/oasis3mct.yaml
@@ -33,6 +33,7 @@ available_versions:
 - '4.0-awicm-3.0'
 - '4.0-awicm-3.1'
 - '5.0-smhi'
+- '6.0-YAC'
 choose_version:
   '2.8-paleodyn':
     branch: '2.8mct-awiesm-2.1'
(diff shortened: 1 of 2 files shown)


joakimkjellsson (approved):
I think this would be a great addition to ESM-Tools! Nicely done! 

But we will need to figure out how to best ship the make.ESMTOOLS file as well. Maybe CERFACS would add it if we asked nicely? 

/J

JanStreffing:
> I think this would be a great addition to ESM-Tools! Nicely done!
> 
> But we will need to figure out how to best ship the make.ESMTOOLS file as well. Maybe CERFACS would add it if we asked nicely?
> 
> /J

I agree. And yes I'm positive we can make that happen. I just need to make a PR from here: https://gitlab.com/awi-esm/oasis3-mct/-/tree/feature/yac_remap_esm-tools?ref_type=heads back into the parent of the fork.

I'm currently trying to compile the first setup with this oasis. Afterwards I try to make a namcouple by hand that works. Finally, I try to make the namcouple with esm_tools. That's the plan at least.

We could of course merge this first, and the setup (awicm3-v3.2.1) separately.

JanStreffing:
Superseeded by: https://github.com/esm-tools/esm_tools/pull/1127

PR #1125, opened 2024-01-23
Small bugfix to concatenate the string for the date file

nwieters:
…, if needed variable in runscript is given as float.

I stumbled across this small bug when I did a Fesom2 run on Albedo. I gave the Fesom Version in my runscript as a number/Float (not in quotes) and I got a concatenation error in the function that resolves the file name for the date file.

### DIFF
--- a/src/esm_runscripts/prepare.py
+++ b/src/esm_runscripts/prepare.py
@@ -41,7 +41,7 @@ def mini_resolve_variable_date_file(date_file, config):
                         )
                         print(f"date_file = {date_file}")
                         sys.exit(1)
-        date_file = str(pre) + str(answer) + str(post)
+        date_file = f"{pre}{answer}{post}"
     return date_file
 
 
--- a/src/esm_runscripts/prepare.py
+++ b/src/esm_runscripts/prepare.py
@@ -41,7 +41,7 @@ def mini_resolve_variable_date_file(date_file, config):
                         )
                         print(f"date_file = {date_file}")
                         sys.exit(1)
-        date_file = pre + answer + post
+        date_file = str(pre) + str(answer) + str(post)
     return date_file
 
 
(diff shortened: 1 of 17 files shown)


src/esm_runscripts/prepare.py
```diff
@@ -41,7 +41,7 @@ def mini_resolve_variable_date_file(date_file, config):
                         )
                         print(f"date_file = {date_file}")
                         sys.exit(1)
-        date_file = pre + answer + post
+        date_file = str(pre) + str(answer) + str(post)
```
mandresm:
```suggestion
        date_file = f"{pre}{answer}{post}"
```
Deniz always told us to stop using + for strings (string concatenation) and use the formatting instead. Formatting is a better practice because is more readable, less error prone, in theory more performant and more flexible. I have followed this recommendation since Deniz pointed that out, and I have implemented this approach in any line having a string concatenation that I modified. I suggest we adopt this practice if none of us has nothing against it.

@pgierz any extra insights about the benefits of using formatting?

mandresm (approved):
Thanks for the fix!

src/esm_runscripts/prepare.py
nwieters: Thanks @mandresm for this tip. Actually this is a good example to pay more attention to how you fix bugs. Not just make it work, but think about if it can be done a little better, even if it is just one line of code. I will try to remember that.

pgierz: Yes, one more insight: the so-called `f-strings` also allow you to encapsulate expressions, so you can do stuff like this:

```python
>>> my_date = f"{1+2000}_date"
>>> my_date
'2001_date'
```

That is (sometimes) easier to read than pre-defining variables and then needing to remember to look in two different places. However, I would recommend to not overuse this, you can just as easily make it too complicated.

nwieters:
#bump

PR #1139, opened 2024-02-26
Sprint/refactor resubmit

nwieters:


mandresm:
All tests are broken, even the ones that usually don't break. @nwieters do you know what's going on?

mandresm:
After a fast look it seems that some user error is triggered in the pytest. Is it because the pytests are not yet finished?

https://github.com/esm-tools/esm_tools/actions/runs/8047430844/job/21976640630?pr=1139#step:8:419

nwieters:
Yes. Because I changed the cluster of the default phases now all the older tests for the workflow are broken. And for a newer test I need to commit a test config file.
I still need to do some work on this PR.

mandresm:
Ah okay, sorry, my bad, I understood it was ready to go. Let me know when I can trigger some real tests in Levante.

mandresm:
@nwieters will look at this and see if can be merge with any of the other branches of the workflow manager

PR #1142, opened 2024-02-29
at some point tag FESOM2 2.5 instead of AWI-CM3_v3.2 was selected. Th…

JanStreffing:
…is tag does not run, as a crutial MPI init with OpenIFS fix was missing

### DIFF
--- a/configs/components/fesom/fesom-2.5.yaml
+++ b/configs/components/fesom/fesom-2.5.yaml
@@ -25,6 +25,8 @@ required_plugins:
 choose_version:
   '2.5':
     branch: "2.5"
+  '2.5-awicm-3.2':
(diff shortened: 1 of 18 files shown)


mandresm:
It was working for me, but probably not working well. What is really weird is that this PR is making no changes in any of the tests... I suspect that's because of the following:
https://github.com/esm-tools/esm_tools/blob/3b81b69006a9c7d3f717dab453685a0b31707b4c/configs/setups/awicm3/awicm3.yaml#L402-L403

Since you select version "2.5" for fesom nothing is really changes. Changes occur only for esm_master and that's not reflected on the tests because the comp tests only check the comp files

mandresm (approved):

mandresm:
#bump #patch

PR #1143, opened 2024-03-06
Correcting version names for FOCI-OpenIFS

joakimkjellsson:
This branch contains a bunch of work that went in to rename FOCI-OpenIFS versions so that the hierarchy is clearer and that names make more sense. 

Main changes to FOCI-OpenIFS: 
* v3.0 is v2.2 but with OASIS-MCT5 (not MCT4). 
* v3.0.1 is v3.0 but no vvl and no smagorinsky, and is therefore a direct twin to AGRIF configurations. 
* v4.0 uses NEMO 4.2.2

Also: 
* moved a bunch of common settings from runscripts into `oifs.yaml` and `focioifs.yaml` to ensure they are default. 
* use -O3 when compiling OASIS and OpenIFS on glogin. 30% faster model. 
* remove what I think is a superfluous choose statement in `xios.yaml`. (https://github.com/esm-tools/esm_tools/commit/59779d076060cb603e9f1206d6c40ab0c63d1591) 

Tests: 
v2.2, v3.0, v3.0-agrif, v4.0 compile and run on glogin. 

The changes should have no impact on non-FOCIOpenIFS models, except for any run with OpenIFS Tco95 where I've now set the default time step to 1800s. 
From what we have found, 3600s is faster but causes unreasonable biases. 1800s is definitely worth the 2x compute cost. 

If @seb-wahl and @JanStreffing are ok, then I'll go ahead and merge this. 

Cheers
Joakim


JanStreffing (approved):
I see no way this would impact AWI-CM3. From that perspective, I approve. Other tests are on you ;-)

PR #1151, opened 2024-03-22
Bugfix for setting use_venv in runscript having no effect.

### LINKED ISSUE #1150: Setting use_venv in runscipt has no effect
**Describe the bug**
Setting the variable `use_venv` in runscript -> general section has no effect, since it will be overwritten by `None` from command_line_config.
This causes esm_runscripts to always asking whether an venv should be generated. Even if set to `False` in the runscript.

**To Reproduce**
Call `esm_runscripts` without `--open-run`

**Expected behavior**
Runscript variable for `use_venv` should be taken when no `--open-run` is given as argument.

**Screenshots**

 - Supercomputer: [which machine you are on]
 - Version [Please run `esm_tools --version`]

esm_tools, version 6.29.0

**Additional context**

nwieters:
Closes #1150

### DIFF
--- a/src/esm_runscripts/config_initialization.py
+++ b/src/esm_runscripts/config_initialization.py
@@ -165,15 +165,13 @@ def get_user_config_from_command_line(command_line_config):
             f"``{command_line_config['runscript_abspath']}`` from the command line.",
         )
 
-    # Update all **except** for use_venv if it was supplied in the
-    # runscript:
-    deupdate_use_venv = False
-    if "use_venv" in user_config["general"]:
-        user_use_venv = user_config["general"]["use_venv"]
-        deupdate_use_venv = True
+    # If the use_env variable defined by the command line is None deleted so that
+    # the definition from the runscript prevails after the user_config.update
+    # If it's not None (True or False) the definition of the command line wins
+    # over the runscript as is the case for all the other variables
+    if command_line_config["use_venv"] is None:
+        del command_line_config["use_venv"]
     user_config["general"].update(command_line_config)
-    if deupdate_use_venv:
-        user_config["general"]["use_venv"] = user_use_venv
 
     return user_config
 
--- a/src/esm_runscripts/config_initialization.py
+++ b/src/esm_runscripts/config_initialization.py
@@ -165,7 +165,15 @@ def get_user_config_from_command_line(command_line_config):
             f"``{command_line_config['runscript_abspath']}`` from the command line.",
         )
 
+    # Update all **except** for use_venv if it was supplied in the
+    # runscript:
+    deupdate_use_venv = False
+    if "use_venv" in user_config["general"]:
+        user_use_venv = user_config["general"]["use_venv"]
+        deupdate_use_venv = True
     user_config["general"].update(command_line_config)
+    if deupdate_use_venv:
+        user_config["general"]["use_venv"] = user_use_venv
 
     return user_config
 
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -46,3 +46,4 @@ documentation issue on `our GitHub repository <https://github.com/esm-tools/esm_
 .. include:: recipes/exclude_forcing_input.rst
 .. include:: recipes/use_own_namelist.rst
 .. include:: recipes/branchoff_from_old_spinups.rst
+.. include:: recipes/batch_mail.rst
--- a/docs/esm_variables.rst
+++ b/docs/esm_variables.rst
@@ -59,31 +59,33 @@ Installation variables
 Runtime variables
 -----------------
 .. csv-table::
-   :header: Key, Description
-   :widths: 15, 85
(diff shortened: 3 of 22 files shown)


nwieters:
Need to do more testing since --contained-run is now/still breaking.

nwieters:
I reactivated the lines that take care of the `use_venv` setting from the runscript.

nwieters:
Hi, I rerun the failing tests and they are now passing. Please feel free to review and also merge this into release.

nwieters:
Hi @mandresm , as I mentioned already, these changes let the setting in the runscript always win or give an error if runscript setting and argument are not campatible. You may want to change this behaviour.
When I started to fix this bug I had a different approach. Maybe you can have a look at that again, if this is useful for you. It is in the commit a269640a0292127e4193467958c7511fc3b26606

pgierz:
Just so I understand this correctly: my current mental model of how this works is as follows:

* Command line arguments win, always. They are recycled run to run (e.g. the next run resubmits itself with the same flags as were given the first time)
* User specifications in the runscript are next. Changing the submission script between runs (assuming it is stored in a different location) has no influence on the run, since it will recycle the on in `${EXP_BASE}/scripts` and not `/some/user/location/scripts/my_run.yaml`
* Changing `${EXP_BASE/scripts/my_run.yaml` **will** change what happens, but if *and only if* the options are not overridden by a command line flag
* Internal (e.g. component/setup/etc) configuration files come last.

Is that more or less correct? Or did I misunderstand the overwrite order?

mandresm:
> Just so I understand this correctly: my current mental model of how this works is as follows:
> 
>     * Command line arguments win, always. They are recycled run to run (e.g. the next run resubmits itself with the same flags as were given the first time)
> 
>     * User specifications in the runscript are next. Changing the submission script between runs (assuming it is stored in a different location) has no influence on the run, since it will recycle the on in `${EXP_BASE}/scripts` and not `/some/user/location/scripts/my_run.yaml`
> 
>     * Changing `${EXP_BASE/scripts/my_run.yaml` **will** change what happens, but if _and only if_ the options are not overridden by a command line flag
> 
>     * Internal (e.g. component/setup/etc) configuration files come last.
> 
> 
> Is that more or less correct? Or did I misunderstand the overwrite order?

That is correct. Somehow we had it the other way around only for `use_venv`, and I don't remember why, but to me this is nonsense (even if I implemented that approach myself) so I'm bringing things back to "normal"

mandresm:
All tests passing so... #bump #patch

mandresm (approved):

src/esm_runscripts/config_initialization.py
```diff
@@ -165,6 +165,12 @@ def get_user_config_from_command_line(command_line_config):
             f"``{command_line_config['runscript_abspath']}`` from the command line.",
         )
 
+    # If the use_env variable defined by the command line is None deleted so that
+    # the definition from the runscript prevails after the user_config.update
+    # If it's not None (True or False) the definition of the command line wins
+    # over the runscript as is the case for all the other variables
+    if command_line_config["use_venv"] is None:
+        del command_line_config["use_venv"]
```
pgierz, line 173: I'm not sure I fully understand this. Does the `use_venv` check for the *presence* of the flag? Because if it only checks if the value is somehow something that evaluates to `True`, this is not really needed. `None` is a `False` thing (well.....technically, not really, they are showing different programming things, but an `if` check will show you the same answer)


```python
>>> my_false_var = False
>>> my_true_var = True
>>> my_none_var = None
>>> if my_false_var:
...     print("woohoo")
...
>>> if my_true_var:
...     print("woohoo2")
...
woohoo2
>>> if my_none_var:
...     print("woohoo3")
...
>>> 
```

pgierz:
> > Just so I understand this correctly: my current mental model of how this works is as follows:
> > ```
> > * Command line arguments win, always. They are recycled run to run (e.g. the next run resubmits itself with the same flags as were given the first time)
> > 
> > * User specifications in the runscript are next. Changing the submission script between runs (assuming it is stored in a different location) has no influence on the run, since it will recycle the on in `${EXP_BASE}/scripts` and not `/some/user/location/scripts/my_run.yaml`
> > 
> > * Changing `${EXP_BASE/scripts/my_run.yaml` **will** change what happens, but if _and only if_ the options are not overridden by a command line flag
> > 
> > * Internal (e.g. component/setup/etc) configuration files come last.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is that more or less correct? Or did I misunderstand the overwrite order?
> 
> That is correct. Somehow we had it the other way around only for `use_venv`, and I don't remember why, but to me this is nonsense (even if I implemented that approach myself) so I'm bringing things back to "normal"

There was probably a good reason at the time, but I can't remember it either ;-) This is cleaner anyway. So, all green!

PR #1152, opened 2024-03-27
fix aero forcing for HIST in echam.yaml

ackerlar:
The aero forcing for the HIST scenario in echam.yaml was linked to the PI forcing. This gave me this error in the second year of simulation: `FATAL ERROR in aero_read_opt: file aero_coarse_1852.nc does not exist`
After changing the echam.yaml to 
```
                        aerocoarse: histaerocoarse
                        aerofarir: histaerofarir
                        aerofin: histaerofin
```
the file is found and no error arises.

The rest is only clean up in the PISM coupling scripts


configs/components/echam/echam.yaml
```diff
@@ -411,20 +411,14 @@ choose_scenario:
                 dataset: r0008
                 greenhouse_dir: "${input_dir}"
                 forcing_files:
-                        # aerocoarse: piaerocoarse
-                        # aerofarir: piaerofarir
-                        # aerofin: histaerofin
+                        #aerocoarse: histaerocoarse
+                        #aerofarir: histaerofarir
+                        #aerofin: histaerofin
```
nwieters: Can these lines be removed?

couplings/fesom/env_fesom.py
```diff
@@ -15,7 +15,7 @@ def prepare_environment(config):
             "EXP_ID": config["general"]["command_line_config"]["expid"],
             "iter_coup_regrid_method_ice2oce": "INTERPOLATE",
             "fesom_use_icebergs": config["fesom"].get("use_icebergs", ""), 
-            "BASIN_FILE": config["fesom"].get("basin_file"),
+            #"BASIN_FILE": config["fesom"].get("basin_file"),
```
nwieters, line 18: Can this line be removed?

nwieters (commented):
Just some minor comments. Looks good to me. But I can't give any comment in detail. Maybe @mandresm can also have a look? Thank you.

configs/components/echam/echam.yaml
ackerlar: Yes, these lines can be removed

couplings/fesom/env_fesom.py
ackerlar: Yes, this line can be removed. I made some more clean up

jannitzbon:
Hi @ackerlar @mandresm  @pgierz 
I was just having the same issue as Lars a year ago. The fix worked for me as well. However, it seems that it has never made it's way into the release branch as some tests failed. I guess I cannot help with resolving the conflicts, but it would be nice if this fix was incorporated into the release branch.

pgierz:
I'll have a look at this later this week, thanks for reminding us @jannitzbon!

configs/components/echam/echam.yaml
pgierz:
```suggestion
```

pgierz (commented):
This looks good, still need to look into the failing tests.

configs/components/echam/echam.yaml
pgierz: done

pgierz:
Fixed merge conflicts, @ackerlar, it seems a commented-out `BASIN_FILE` made its way back and is now in there. I guess that isn't needed anymore?

ackerlar:
> Fixed merge conflicts, @ackerlar, it seems a commented-out `BASIN_FILE` made its way back and is now in there. I guess that isn't needed anymore?

Yes, you are right. It is not needed anymore

PR #1155, opened 2024-04-08
Bugfix for correct awicm3 general version in choose block

### LINKED ISSUE #1154: Set the 'version' variable including other variables defined in choose_ blocks
version: "2.0${subversion}"

Problem: will take the wrong component yaml file for e.g. fesom.yaml instead of fesom-2.0.yaml

'subversion' should be defined in another choose_ block

**Describe alternatives you've considered**
Defining 'version' in different choose_ block does not work.

**Additional context**
In the finished_config file the right version is set, but not when the related component config file name is gathered.

nwieters:
This PR corrects a small bugfix in awicm3.yaml for setting the right choose in a `choose_general.version` block.

### DIFF
--- a/configs/setups/awicm3/awicm3.yaml
+++ b/configs/setups/awicm3/awicm3.yaml
@@ -368,7 +368,7 @@ fesom:
                         nproc: 144
                 TL159_CORE2:
                         nproc: 72
-        
+
         comp_command: mkdir -p build; cd build; cmake -DOIFS_COUPLED=ON -DFESOM_COUPLED=ON ..;   make install -j `nproc --all`
 
         opbnd_dir: ""
@@ -409,13 +409,13 @@ fesom:
                                 namelist.oce:
                                         oce_tra:
                                                 surf_relax_s: "${surf_relax_s}"
-                        
+
                         # Remove ICMCL_INIT forcing, which is only required for standalone simulations (form v3.2 onwards contains seasonal leaf area index, veg albedo etc.)
                         remove_forcing_files:
                                 - ICMCL_INIT
-                        
+
                         choose_general.version:
-                                "2.0-awicm-3.1-recom":
+                                "v3.1.2_recom":
                                         subversion: "-recom"
                                         comp_command: mkdir -p build; cd build; cmake -DOIFS_COUPLED=ON -DFESOM_COUPLED=ON -DRECOM_COUPLED=ON ..;   make install -j `nproc --all`
                                 "*":
@@ -888,7 +888,7 @@ oasis3mct:
                 rstos.nc: rstos.nc
         remove_config_files:
                 - cf
-        
+
 
 
 computer:
--- a/configs/setups/awicm3/awicm3.yaml
+++ b/configs/setups/awicm3/awicm3.yaml
@@ -21,7 +21,7 @@ general:
         - 'v3.1'
         - 'v3.1.1'
         - 'v3.1.2'
-        - 'v3.1.2_recom'
+        - 'v3.1.2-recom'
         - 'v3.2'
         - 'master'
         - 'frontiers-xios'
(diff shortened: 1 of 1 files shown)


nwieters:
A short description about what still fails in this branch and in branch `awicm3_recom`:

- In `awicm3.yaml` the variable `tidy_recipe: ${fesom.tidy_recipe}` is set.
- Because the fesom version is set with a variable `subversion`, the wrong `fesom.yaml` is taken (`fesom.yaml` instead of `fesom-2.0.yaml`)
  You can test this by masking the actual error by copying fesom-2.0.yaml -> fesom.yaml and look into the finished config: it says 
```
fesom:
    ...
    debug_info:
        loaded_from_file: /work/ab0995/a270089/esm_tools/configs//components/fesom/fesom.yaml
```
Now it finds the variable `tidy_recipe`, but it takes the **wrong** fesom yaml config file.
- In `fesom.yaml` there is no variable `tidy.recipe` set
- This leads to an infinit loop searching for the variable tidy.recipe which gives the error 
`RecursionError: maximum recursion depth exceeded while calling a Python object`


I think the cause of this error is in the `esm_parser` function `look_for_file` line 234. 

In the case of the fesom version set with a subversion variable (as in `awicm3.yaml` of branch `awicm3_recom`), at the time the actual fesom config file name is generated, the fesom version variable looks still like this `fesom-2.0${subversion}`. Without the variable `${subversion}` evaluated. In line 234 of `esm_parser.py` it will cut off everything after and including the minus sign. So the fesom config file name will be `fesom.yaml`  which leads to the above described error.

mandresm (approved):
It should be working all now. @nwieters, can you give it another try?

This type of implementation was already done by Ying, @mbutzin and myself into AWIESM-2.1. One can find a lot of inspiration from the problems to come in AWICM3-REcoM in the `configs/setups/awiesm/awiesm.yaml` and in the `configs/components/fesom/fesom-2.1.yaml` specially if one searches for the `with_recom`.

@nwieters, merge whenever you consider is ready :)

nwieters:
Unfortunatly this doesn't solve the problem with the wrong fesom.yaml file taken.

nwieters:
Hi @chrisdane, @mandresm ,

I merged this PR into the branch awicm3-recom. It still does not solve the problem with the resolving the variable in the fesom version. But the workaround for this is to set the fesom version explicitly in the runscript

```
fesom:
    version: "2.0-recom"
```

With this I think @chrisdane can start to test awicm3-recom for now. I will fix the version issue then in release.

mandresm:
A very important remark for @chrisdane: once this is fix, one should not manually set the version of fesom (unless experimenting things) as specifying a particular version of AWICM3 should be enough, avoiding confusion and errors.

chrisdane:
Thanks a lot all. 

On 93409af the recom section now appears in the finished config of a check run. However, the namelist.recom is not in the run dir (the esm_runscripts call does not complain not to find namelist.recom). Any idea how to "activate" recom in this case?

Runscript:
```
/home/a/a270073/esm/runscripts/awicm3-v3.1.2_recom-levante.yaml
```
Expdir:
```
/work/ba1103/a270073/out/awicm3-v3.1.2-recom/test
```

Thanks a lot!
Chris

mandresm:
This is normal, the yaml file of `awicm3` is not including the `recom.yaml`, that means no information from the default recom configuration is being used.

The solution is to follow a similar logic as in `awiesm2` (use `awiesm.yaml` as a template) to include recom info into this coupled setup if `with_recom` is `True`:
https://github.com/esm-tools/esm_tools/blob/2d980df9a923e666f8797b9d96baa1ad3ff2a6bc/configs/setups/awiesm/awiesm.yaml#L110-L114

`add_include_models` will make sure that the info in `recom.yaml` is read.

https://github.com/esm-tools/esm_tools/blob/2d980df9a923e666f8797b9d96baa1ad3ff2a6bc/configs/setups/awiesm/awiesm.yaml#L434-L463

The box above covers more lines than it seems, so please scroll down on the box to see the whole thing. By adding a `recom` section you can modify in `awicm3.yaml` any `recom` information you need to modify without touching the default info in the `recom.yaml`. The choose is to make sure not to add recom information in simulations that don't run with recom. You probably don't need all of that, this is just so you have an idea how things could be done.

mandresm:
@chrisdane, let me know if we need to meet together with @nwieters to give you further advice on how to proceed, or if you think it would be good to discuss planning on support for adding this feature.

PR #1158, opened 2024-04-12
add a new function to recursively resolve early variables

mandresm:
Now early variables are not only resolved if the are in a choose block, but also their values are resolve if depending on another variable or math operation

**TODO**
- [ ] Docstrings
- [ ] Comments
- [ ] Fix tests
- [ ] Review and improve algorithm, specially the part where it is catching the error from `find_variable`
- [ ] Merge all these fixes above to Chris' so that his branch is up to date


nwieters:
closes #1154

PR #1159, opened 2024-04-12
Fix/resolve early vars

mandresm:
Hi @chrisdane (and @nwieters):

This is an unpolished fix that solves the fact that you were still required to define the `fesom.version` in your runscript and/or `fesom.with_recom` in your runscirpt. Unpolished means, it's working but is missing documentation and an esthetic cleanup, but you can merge it right away and start using it, just press "Merge".

With this fix you can now remove those variables from your runscript and the only thing you need to specify is your awicm3 version:

**your runscript**
```yaml
general:
    version: "v3.1.2-recom"
```

In the same way, this should allow you to do `esm_master get-awicm3-v3.1.2-recom` and you should get the branch of fesom `origin/awicm-3-recom` (after the esm_master command `cd awicm3-v3.1.2-recom/fesom-2.0` and `git status`).


src/esm_parser/esm_parser.py
```diff
@@ -1527,6 +1527,85 @@ def resolve_choose(model_with_choose, choose_key, config):
         logging.debug("key=%s", key)
 
 
+def resolve_early_var(
+    var, config, current_model=None, user_config={}, model_config={}, setup_config={}
+):
+
+    #import ipdb
+    if not current_model:
```
pgierz, line 1535: Not sure about this, but maybe you need something like this instead:
```suggestion
    if current_model is None:
```

As currently, it is you check if it is not a "False"-y thing.

src/esm_parser/esm_parser.py
```diff
@@ -1527,6 +1527,85 @@ def resolve_choose(model_with_choose, choose_key, config):
...
+    )
+
+    if var not in this_config[current_model]:
+        logger.debug(f"``{var}`` not found in this config (resolve_early_var)")
+        return
+
+    value = this_config[current_model].get(var)
+
+    #if value==['xios']:
+        #ipdb.set_trace()
```
pgierz, line 1564: Removed

```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -1527,6 +1527,85 @@ def resolve_choose(model_with_choose, choose_key, config):
...
+        var,
+        this_config,
+        current_model=current_model,
+        user_config=user_config,
+        model_config=model_config,
+        setup_config=setup_config,
+    )
+
+    if var not in this_config[current_model]:
+        logger.debug(f"``{var}`` not found in this config (resolve_early_var)")
```
pgierz, line 1558:
```suggestion
        logger.debug(f"``{var}`` not found in this config for {current_model} (resolve_early_var)")
```

src/esm_parser/esm_parser.py
```diff
@@ -1527,6 +1527,85 @@ def resolve_choose(model_with_choose, choose_key, config):
...
+            "Please define a ``current_model`` for ``resolve_early_var`` method"
+        )
+        sys.exit(1)
+    elif current_model in config:
+        this_config = config
+    elif current_model == config.get("model"):
+        this_config = {current_model: config}
+    else:
+        logger.error(f"``current_model: {current_model}`` and ``config`` do not correspond!")
+        return
```
pgierz, line 1546: `return`, or do we want to exit?

```suggestion
        sys.exit(1)
```

src/esm_parser/esm_parser.py
```diff
@@ -1527,6 +1527,85 @@ def resolve_choose(model_with_choose, choose_key, config):
...
+        this_config,
+        current_model=current_model,
+        user_config=user_config,
+        model_config=model_config,
+        setup_config=setup_config,
+    )
+
+    if var not in this_config[current_model]:
+        logger.debug(f"``{var}`` not found in this config (resolve_early_var)")
+        return
```
pgierz, line 1559: As above:

```suggestion
        sys.exit(1)
```

src/esm_parser/esm_parser.py
```diff
@@ -1527,6 +1527,85 @@ def resolve_choose(model_with_choose, choose_key, config):
...
+        subvars = re.findall(value_pattern, value)
+        for subvar in subvars:
+            path_to_var = subvar.split(".")
+            if len(path_to_var) > 1:
+                subvar_component = path_to_var[0]
+                subvar = ",".join(path_to_var[1:])
+            else:
+                subvar_component = current_model
+            #if var == "version" and var in config[current_model] and current_model=="fesom":
+                #ipdb.set_trace()
```
pgierz, line 1577:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -1527,6 +1527,85 @@ def resolve_choose(model_with_choose, choose_key, config):
...
+                config_to_search_into = setup_config
+            elif subvar in model_config.get(subvar_component, []):
+                config_to_search_into = model_config
+            else:
+                return
+
+            try:
+                new_value = find_variable([current_model], this_config[current_model][var], config_to_search_into, [], True)
+            except esm_parser.esm_parser.EsmParserError as error:
+                not_found_variable = str(error).split("was not resolved: ")[1].split(" not found")[0]
```
pgierz, line 1593: Interesting idea:

Can `Error` objects return things or have attributes?

Yes. See this example:
```python
class EsmParserError(Exception):
    """Base class for exceptions in the ESM parser."""
    def __init__(self, message="An error occurred in the ESM parser"):
        self.message = message
        super().__init__(self.message)

class EsmParserVariableNotFoundError(EsmParserError):
    """Exception raised when a variable is not found in the ESM parser."""
    def __init__(self, variable_name):
        self.variable_name = variable_name
        self.message = f"Variable '{self.variable_name}' not found in the ESM parser."
        super().__init__(self.message)

# Example usage
try:
    # Simulate code that may raise EsmParserVariableNotFoundError
    raise EsmParserVariableNotFoundError("my_variable")
except EsmParserVariableNotFoundError as e:
    # Extract the variable name from the exception and perform some logic with it
    variable_name = e.variable_name
    capitalized_variable_name = variable_name.capitalize()
    print(f"Original variable name: {variable_name}")
    print(f"Capitalized variable name: {capitalized_variable_name}")
```

This might be worth using in a future implementation...

src/esm_runscripts/config_initialization.py
```diff
@@ -165,6 +165,12 @@ def get_user_config_from_command_line(command_line_config):
             f"``{command_line_config['runscript_abspath']}`` from the command line.",
         )
 
+    # If the use_env variable defined by the command line is None deleted so that
+    # the definition from the runscript prevails after the user_config.update
+    # If it's not None (True or False) the definition of the command line wins
+    # over the runscript as is the case for all the other variables
+    if command_line_config["use_venv"] is None:
+        del command_line_config["use_venv"]
```
pgierz, line 173: Is this left over from a different PR? It seems to be unrelated...

pgierz (commented):
Looks good. See some comments I had.

I'm not going to "officially" give this a green light yet until we get a test from one of the RECOM guys.

PR #1161, opened 2024-04-18
Fix/paleo echam

ackerlar:
Changed back the ECHAM version from "6.3.05p2-awiesm-2.1-paleodyn-2.5" to "6.3.05p2-awiesm-2.1" for paleo setup after everything has been merged on the ECHAM side

### DIFF
--- a/configs/components/echam/echam.yaml
+++ b/configs/components/echam/echam.yaml
@@ -22,7 +22,7 @@ compile_infos:
         branch: ${echam.version}
         choose_version:
           6.3.05p2-awiesm-2.1:
-            branch: "6.3.05p2-awiesm-2.1-paleodyn-2.5" #LA: change back to 6.3.05p2-awiesm-2.1 once iceberg feature is merged https://gitlab.awi.de/paleodyn/Models/echam6/-/merge_requests/4
+            branch: "6.3.05p2-awiesm-2.1"
             clean_command: rm -rf src/echam/bin; rm -rf bin; make clean
             comp_command: ./config/createMakefiles.pl; autoreconf -i --force; mkdir -p src/.deps yaxt/src/.deps yaxt/tests/.deps; ./configure $configure_opts --with-fortran=intel INSTALL='/usr/bin/install -p'; make -j `nproc --all`; make install -j `nproc --all`; mkdir -p src/echam/bin; cp  bin/echam6 src/echam/bin/echam6
             git-repository: "https://gitlab.awi.de/paleodyn/Models/echam6.git"
@@ -411,9 +411,6 @@ choose_scenario:
                 dataset: r0008
                 greenhouse_dir: "${input_dir}"
                 forcing_files:
-                        #aerocoarse: histaerocoarse
-                        #aerofarir: histaerofarir
-                        #aerofin: histaerofin
                         aerocoarse: piaerocoarse
                         aerofarir: piaerofarir
                         aerofin: piaerofin
--- a/configs/components/echam/echam.yaml
+++ b/configs/components/echam/echam.yaml
@@ -411,9 +411,12 @@ choose_scenario:
                 dataset: r0008
                 greenhouse_dir: "${input_dir}"
                 forcing_files:
-                        aerocoarse: histaerocoarse
-                        aerofarir: histaerofarir
-                        aerofin: histaerofin
+                        #aerocoarse: histaerocoarse
+                        #aerofarir: histaerofarir
+                        #aerofin: histaerofin
+                        aerocoarse: piaerocoarse
+                        aerofarir: piaerofarir
+                        aerofin: piaerofin
                         sst: amipsst
                         sic: amipsic
                         ozone: ozone_hist_scenario
--- a/configs/components/echam/echam.yaml
+++ b/configs/components/echam/echam.yaml
@@ -411,20 +411,11 @@ choose_scenario:
                 dataset: r0008
                 greenhouse_dir: "${input_dir}"
                 forcing_files:
-                        # aerocoarse: piaerocoarse
-                        # aerofarir: piaerofarir
-                        # aerofin: histaerofin
+                        aerocoarse: histaerocoarse
+                        aerofarir: histaerofarir
+                        aerofin: histaerofin
                         sst: amipsst
                         sic: amipsic
-                        aerocoarse_1849: piaerocoarse
-                        aerocoarse_1850: piaerocoarse
-                        aerocoarse_1851: piaerocoarse
-                        aerofarir_1849: piaerofarir
-                        aerofarir_1850: piaerofarir
-                        aerofarir_1851: piaerofarir
-                        aerofin_1849: piaerofin
-                        aerofin_1850: piaerofin
-                        aerofin_1851: piaerofin
                         ozone: ozone_hist_scenario
                         greenhouse: greenhouse_hist_scenario
                         volcir: histvolcir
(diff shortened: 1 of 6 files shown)


pgierz (approved):

pgierz:
@mandresm: any reason to wait on this one? 

I just did a test compilation, and everything seems to be working. We can wait until a mini PI run goes through as well, but if that works, I would vote to put this into release.

Info for @christian-stepanek. These are the versions installed by `esm_master install-awiesm-2.1-paleodyn-2.5`:

```console
git clone -b 2.8mct-awiesm-2.1 https://gitlab.awi.de/paleodyn/Models/oasis3-mct.git oasis
git clone -b 6.3.05p2-awiesm-2.1 https://gitlab.awi.de/paleodyn/Models/echam6.git echam-6.3.05p2
git clone -b 2.5.1_paleo https://github.com/FESOM/fesom2.git fesom-2.5
```

pgierz:
After some additional fiddling, I have an almost working run, except for the location of a suitable `core2` mesh. I have been "out of the game" for long enough with active modelling to find one on the fly, but everything else loads correctly. See the latest commit for small changes to the configuration files.

I'd like to also take this opportunity to do some "cleaning" about the mentality surrounding version numbers, as this seems to be a topic that was for the `awiesm-2.x` world severely misimplemented. First some theory:

## How version numbers _should_ be defined:

Software normally follows one of two schemes for how to write down version. The simple form is based on some kind of calendar, which is how it is done e.g. in the `ubuntu` world. For `esm-tools`, we chose instead a `major.minor.fix` scheme. Major versions are updated when you introduce "breaking changes", such that using an old copy of your program with the newest version **will not work**. Minor versions are typically for new features, and fix numbers are for bug fixes.

## `AWI-ESM` version numbers (_in theory_):
This is also what I had in mind originally for versioning `awiesm`, and I will for simplicity leave out fix numbers here. It breaks down as follows:

`awiesm-2.x`. First off, the `major = 2` signifies that the model is composed of software based on top of `echam-6.XXX` and `fesom-2.XXX`. Then, the minor versions: 
* `awiesm-2.1` should be "bare bones" `echam-6.3.05p2` and `fesom-2.1`.
* `awiesm-2.2` should be "bare bones" plus `wiso` 
* `awiesm-2.3` includes `wiso` and `pism`. 
* `awiesm-2.4` should include `wiso`, `pism` and `icebergs`. 
* `awiesm-2.5` changed FESOM, so it should have been `echam-6.3.05p2`, `fesom-2.5`, `wiso`, `pism`, and `icebergs`.
* `awiesm-2.6` should have been `echam-6.3.05p2`, `fesom-2.5`, `wiso`, `pism`, `icebergs`, and `caverns` 

... and so on and so forth. However, the PI choices made to switch FESOM version somehow "on the fly" and then re-port features into this newer version made it all version confusing. Note that I have not touched anything about parameter settings in the describing the version numbers. In my opinion, which parameter settings you use for, e.g. mixing, is controlled via the namelist, so strictly speaking, the source code itself remains the same. Attaching a version number to a parameter setting would in my view be bad practice.

To make it worse, we started to add the FESOM version number as part of the version numbers, so we had:

## `AWI-ESM` version numbers (_in practice_):

* `awiesm-2.1`: bare-bones, as expected.
* `awiesm-2.1-wiso`: should have been `2.2`
* `awiesm-2.2`: bare bones, but with pism
* `awiesm-2.1-paleodyn-2.5`: Somę mixture of icebergs included, wiso available, and a different core version of FESOM.

Needless to say, it's a mess. 

@christian-stepanek, could you maybe briefly give me an idea where and in which group it would be best to discuss this? I know @JanStreffing is doing excellent work for the AWICM-3 world, and I really don't want to end up in this mess another time.

ackerlar:
Thanks @pgierz to bring this up! I think part of the problem is (or was) that there has been no linear development but rather parallel use of different "awiesm-2.1" versions (with wiso, iceberg, pism coupling) that were derived from one common version and never really merged back. I think, now we have really one main branch that everybody can use. Maybe we can just call it `awiesm-2.6` (or whatever) and start versioning from now on. On the other hand, I don't think there will be much development going on with this version int he future anyway.

christian-stepanek:
> I know @JanStreffing is doing excellent work for the AWICM-3 world, and I really don't want to end up in this mess another time.

You already gave the reason why you will not end up in this mess another time.

Indeed, the AWI-ESM2 development has been done in various parallel strains, lots of people working on their own business. At one point the need arose to put individual work back into one single model. This has been done in several steps that do not necessarily follow a logical order.

I think one of the problems has also been that there has not been a clear translation of the traditional version numbers into the AWI-ESM2 model realm. The decision was made that awi-esm2.1 means it does not do any interactive ice, while awi-esm2.2 means it does do the ice simulation. Implicitly this idea included the statement that everyone just adds to these two model versions whatever component floats their boat.

While AWI-ESM2 development is not yet at the end (Pengyang's cavities, flexible meshes, etc.), I agree with Lars that the major model development has likely ended. Most people will put their efforts into AWI-ESM3, where Jan keeps an eye on tidiness. We can, of course, define a new version number for the current "final" model as it is, incrementing the number later on for major additions when they happen. Yet, I do not know how we erase the impression from people's minds that there is more than 2.1 and 2.2, and that also other versions than 2.2 may simulate ice sheets.

mandresm:
> @mandresm: any reason to wait on this one?

Yes, the tests in levante > awicm-recom-historical.

We need @chrisdane opinion on that. The question is already on another PR

mandresm:
> Info for @christian-stepanek. These are the versions installed by `esm_master install-awiesm-2.1-paleodyn-2.5`:

I think is instaled by `esm_master install-awiesm-2.1` instead.

@christian-stepanek has a really good point: if there is little development more to be expected from awiesm-2 maybe is better to don't clean up too much, because it would be a waste of time

configs/components/echam/echam.yaml
```diff
@@ -411,20 +411,11 @@ choose_scenario:
...
                         sic: amipsic
-                        aerocoarse_1849: piaerocoarse
-                        aerocoarse_1850: piaerocoarse
-                        aerocoarse_1851: piaerocoarse
-                        aerofarir_1849: piaerofarir
-                        aerofarir_1850: piaerofarir
-                        aerofarir_1851: piaerofarir
-                        aerofin_1849: piaerofin
-                        aerofin_1850: piaerofin
-                        aerofin_1851: piaerofin
```
chrisdane, line 427: are you sure these are not needed? echam needs plus/minus 1 or 2 years for some forcing fields

chrisdane:
> > @mandresm: any reason to wait on this one?
> 
> Yes, the tests in levante > awicm-recom-historical.
> 
> We need @chrisdane opinion on that. The question is already on another PR

I am fine with this PR except the one comment above.

Regarding the versioning I'm fine with Pauls theory suggestion but I'm afraid in the end reality pushes us to choose the practice path :D

pgierz:
> if there is little development more to be expected from awiesm-2 maybe is better to don't clean up too much, because it would be a waste of time

I wasn't suggesting we clean it up; just a little nudge to everyone (actually, people involved in this thread know all of this already) of how it should have been done ;-)

ackerlar:
> > Info for @christian-stepanek. These are the versions installed by `esm_master install-awiesm-2.1-paleodyn-2.5`:
> 
> I think is instaled by `esm_master install-awiesm-2.1` instead.

The FESOM versions differ between 'awiesm-2.1' and 'awiesm-2.1-paleodyn-2.5'

PR #1162, opened 2024-04-19
Improvements on filemovements

mandresm:
This PR aims to:
- [x] Remove the intermediate copying by default (except for `config`, `input`, and `forcing`. The defaults can be overridden by `general.intermediate_movement = [list of filetypes that go to the intermediate folders]`
- [ ] [to be done in another PR] Parallelize the file movements (WIP)
- [x] Profiling

This is to reduce the copying times of @tsemmler05, @a270067 and @mjy1011.

@JanStreffing and @mjy1011, this branch is not in it's final stage, but it would be a great help if you can check that is working for an actual simulation. I've tested already a bit but not in depth.

### DIFF
--- a/src/esm_runscripts/filelists.py
+++ b/src/esm_runscripts/filelists.py
@@ -946,9 +946,11 @@ def copy_files(config, filetypes, source, target):
                 this_intermediate_movements = config[model].get(
                     "intermediate_movements", intermediate_movements
                 )
+                skip_intermediate = False
                 if filetype not in intermediate_movements:
                     if text_target == "intermediate":
                         this_text_target = "targets"
+                        skip_intermediate = True
                     elif text_source == "intermediate":
                         continue
                 sourceblock = config[model][filetype + "_" + text_source]
@@ -1007,8 +1009,16 @@ def copy_files(config, filetypes, source, target):
                                 "file_source": file_source,
                                 "file_target": file_target,
                             })
+
+                            # To avoid overwriting in general experiment folder
+                            if skip_intermediate == True:
+                                file_target = avoid_overwriting(
+                                    config, file_source, file_target
+                                )
+
+                            # Execute movement
                             movement_method(file_source, file_target)
-                            # shutil.copy2(file_source, file_target)
+
                             successful_files.append(file_source)
                         except IOError:
                             print(
@@ -1029,6 +1039,38 @@ def copy_files(config, filetypes, source, target):
                 helpers.print_datetime(config)
         config["general"]["files_missing_when_preparing_run"].update(missing_files)
     return config
+
+
+def avoid_overwriting(config, source, target):
+    if os.path.isfile(target):
+        if filecmp.cmp(source, target):
+            return target
+
+        date_stamped_target = f"{target}_{config['general']['run_datestamp']}"
+        if os.path.isfile(date_stamped_target):
+            esm_parser.user_note(
+                "File movement conflict",
+                f"The file ``{date_stamped_target}`` already exists. Skipping movement:\n"
+                f"{soucer} -> {date_stamped_target}"
+            )
+            return target
+
+        if os.path.islink(target):
+            os.remove(target)
+        else:
+            os.rename(target, f"{target}_{config['general']['last_run_datestamp']}")
+
+        os.symlink(date_stamped_target, target)
+        target = date_stamped_target
+
+    elif os.path.isdir(target):
+        esm_parser.user_error(
+            "File operation not supported",
+            f"The target ``{target}`` is a folder, and this should not be happening "
+            "here. Please, open an issue in www.github.com/esm-tools/esm_tools"
+        )
+
+    return target
 
 
 def filter_allowed_missing_files(config):
--- a/src/esm_runscripts/filelists.py
+++ b/src/esm_runscripts/filelists.py
@@ -1080,7 +1080,7 @@ def avoid_overwriting(config, source, target):
             esm_parser.user_note(
                 "File movement conflict",
                 f"The file ``{date_stamped_target}`` already exists. Skipping movement:\n"
-                f"{soucer} -> {date_stamped_target}"
+                f"{source} -> {date_stamped_target}"
             )
             return target
 
--- a/src/esm_runscripts/filelists.py
+++ b/src/esm_runscripts/filelists.py
@@ -1058,7 +1058,9 @@ def avoid_overwriting(config, source, target):
     and creates a link named ``target`` that points at the target with the current time
     stamp.
 
-    Note: This function does not execute the file movement.
+    Note
+    ---- 
+        This function does not execute the file movement.
 
     Parameters
     ----------
--- a/src/esm_runscripts/filelists.py
+++ b/src/esm_runscripts/filelists.py
@@ -1084,7 +1084,7 @@ def avoid_overwriting(config, source, target):
 
         date_stamped_target = f"{target}_{config['general']['run_datestamp']}"
         if os.path.isfile(date_stamped_target):
-            esm_parser.user_note(
+            esm_parser.user_error(
                 "File movement conflict",
                 f"The file ``{date_stamped_target}`` already exists. Skipping movement:\n"
                 f"{source} -> {date_stamped_target}"
--- a/src/esm_runscripts/filelists.py
+++ b/src/esm_runscripts/filelists.py
@@ -920,7 +920,7 @@ def copy_files(config, filetypes, source, target):
 
     intermediate_movements = config["general"].get(
         "intermediate_movements",
-        ["config", "input"], # DEFAULTS
+        [],
     )
 
     if source == "init":
--- a/src/esm_runscripts/filelists.py
+++ b/src/esm_runscripts/filelists.py
@@ -914,6 +914,13 @@ def copy_files(config, filetypes, source, target):
     of high resolution simulations. A better fix is not made because ``filelists`` are
     being entirely reworked, but the fix cannot wait.
 
+    Note
+    ----
+    Relevant variables in this function:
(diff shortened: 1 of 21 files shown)


JanStreffing:
This week I will be back. Hopefully be Wednesday or so, I get to give this a test.

mandresm:
I consider this finished, so feel free to give it a try. I will work on the parallelization on another PR, and that would take longer.

PD: Wednesday is bank holiday ;)

mjy1011:
Hi Miguel, thank you so much for the modification.
We tried on Aleph, but there is an error saying  'restart file of OIFS, especially rcf, waminfo, and rcf* files are not found '
They are searching these in EXP/run_*/restart/oifs directory while other BL* LA* files are searching in EXP/restart/oifs

mjy1011:
Hi Miguel, thank you very much for the modification.
On Aleph when we restart experiment with the updated codes, attached error occurs.
It seems oifs restart info files (rcf, waminfo)' path  are not detected correctly.
run_error

Do we need to modify any other files? for example,  /esm_tools/configs/esm_software/esm_runscripts/defaults.yaml ?

mandresm:
Hi @mjy1011, I'll be looking into this today. The problem is that the preprocessing scripts expect files in the intermediate folders.

mandresm:
Hi @mjy1011,

I've got done some progress today with the issue, but there are some weird errors still. I think I am close, but I'll have to continue tomorrow.

mjy1011:
> Hi @mjy1011,
> 
> I've got done some progress today with the issue, but there are some weird errors still. I think I am close, but I'll have to continue tomorrow.

That's nice. Thank you for the update.

mandresm:
Hi @mjy1011,

That was a real tough one... but I think it's working now. Can you please try?

@JanStreffing, can you try if the results are bit identical with `release` and this branch? I am getting non-bit identical results in some OpenIFS output files and restarts, and I am wondering if that's normal? (for example, maybe the results in the files are identical but there is something different in the headers)

mjy1011:
> Hi @mjy1011,
> 
> That was a real tough one... but I think it's working now. Can you please try?
> 
> @JanStreffing, can you try if the results are bit identical with `release` and this branch? I am getting non-bit identical results in some OpenIFS output files and restarts, and I am wondering if that's normal? (for example, maybe the results in the files are identical but there is something different in the headers)

Hi ~~
I am waiting for the pre-running job to be finished. I'll try and let you know in a few days. Nice work again!

src/esm_runscripts/filelists.py
```diff
@@ -901,11 +901,43 @@ def resolve_symlinks(config, file_source):
...
+    config : dict
+        The general configuration
+    filetypes : list
+        List of file types to be copied/linked/moved
+    source : str
+        Specifies the source type, to be chosen between ``init``, ``thisrun``,
+        ``work``.
+    target : str
+        Specifies the target type, to be chosen between ``init``, ``thisrun``,
+        ``work``.
```
pgierz: Yay for docstrings! 🥳

src/esm_runscripts/filelists.py
```diff
@@ -901,11 +901,43 @@ def resolve_symlinks(config, file_source):
...
+    or file type vie the ``file_movements``.
+
+    Note: when the ``target`` is ``thisrun`` (intermediate folders) check whether the
+    type of file is included in ``intermediate_movements``. If it's not, instead of
+    moving the file to the intermediate folder it moves it to ``work``. This is an
+    ugly fix to provide a fast solution to the problem that files are
+    copied/moved/linked twice unnecessarily, and this affects inmensely the performance
+    of high resolution simulations. A better fix is not made because ``filelists`` are
+    being entirely reworked, but the fix cannot wait.
+
```
pgierz, line 916:
```suggestion

    Note
    ----
    Relevant variables in this function:
    
    intermediate_movements : list
        List of file types that will be considered in the intermediate step (copy         from source to intermediate and then to work, rather than directly to work)

```

src/esm_runscripts/filelists.py
```diff
@@ -901,11 +901,43 @@ def resolve_symlinks(config, file_source):
...
     helpers.print_datetime(config)
+
     successful_files = []
     missing_files = {}
 
+    intermediate_movements = config["general"].get(
+        "intermediate_movements",
+        [],
+    )
+
```
pgierz, line 948: This is probably trivial, but is a bit confusing. If there is no defined intermediate movements by default, you get an empty list. Above, in `configs/defaults/general.yaml` is where the actual default list is defined. Maybe put it here instead? Or just add a comment so it is clearer:

```suggestion
    # See the default intermediate movements list in `configs/defaults/general.yaml`
    intermediate_movements = config["general"].get(
        "intermediate_movements",
        [],
    )

```

src/esm_runscripts/filelists.py
```diff
@@ -988,6 +1050,55 @@ def copy_files(config, filetypes, source, target):
...
 
 
+def avoid_overwriting(config, source, target):
+    """
+    Function that appends the date stamp to ``target`` if the target already exists.
+    Additionally, if the target exists, it renames it with the previous run time stamp,
+    and creates a link named ``target`` that points at the target with the current time
+    stamp.
+
+    Note: This function does not execute the file movement.
```
pgierz:
```suggestion
    Note
    ---- 
        This function does not execute the file movement.
```

src/esm_runscripts/filelists.py
```diff
@@ -988,6 +1050,55 @@ def copy_files(config, filetypes, source, target):
...
+    if os.path.isfile(target):
+        if filecmp.cmp(source, target):
+            return target
+
+        date_stamped_target = f"{target}_{config['general']['run_datestamp']}"
+        if os.path.isfile(date_stamped_target):
+            esm_parser.user_note(
+                "File movement conflict",
+                f"The file ``{date_stamped_target}`` already exists. Skipping movement:\n"
+                f"{soucer} -> {date_stamped_target}"
```
pgierz:
```suggestion
                f"{source} -> {date_stamped_target}"
```

src/esm_runscripts/filelists.py
```diff
@@ -988,6 +1050,55 @@ def copy_files(config, filetypes, source, target):
...
+        Path of the source of the file that will be copied/moved/linked
+    target : src
+        Path of the target of the file that will be copied/moved/linked
+    """
+    if os.path.isfile(target):
+        if filecmp.cmp(source, target):
+            return target
+
+        date_stamped_target = f"{target}_{config['general']['run_datestamp']}"
+        if os.path.isfile(date_stamped_target):
```
pgierz, line 1086: Should this throw an error and break? Or just a warning?

pgierz (changes requested):
Looks good!

Only a few minor changes (and one typo 😉)

src/esm_runscripts/filelists.py
mandresm: This is a really good question... In principle, I foresee see this been reach in any circumstances (neither in the actual tests I just ran in Levante), but if it's reached it would be very worrying because it means someone has externally paste a file that shouldn't be there or trigger a rerun in a very weird way. If we are talking about restarts being moved around, not overwriting here would be a big problem, but in other circumstances overwriting might be problematic for the user as well. In summary, there is not a clear behavior ESM-Tools should take under this circumstances, and consequently I'd rather break the simulation than doing something without the user knowing what's going on.

mandresm:
> Looks good!
> 
> Only a few minor changes (and one typo 😉)

Thanks for the review!

src/esm_runscripts/filelists.py
pgierz: I agree, if we just throw a warning, and let the run continue, this would be really hard to debug. In that case, I'd use `user_error`. If I remember correctly, that also triggers a program exit.

src/esm_runscripts/filelists.py
```diff
@@ -988,6 +1050,55 @@ def copy_files(config, filetypes, source, target):
...
+            esm_parser.user_note(
```
pgierz:
```suggestion
            esm_parser.user_error(
```

I did not check if you still need an import for this change.

pgierz (changes requested):
I agree with the error approach!

mjy1011:
Hi Miguel, is there an option in esm-tool to run " compute " only  in LSF batch system ?
Tidy and Prepcompute doesn't use multiple process, but very often killed in lsf batch job during the process in KMA machine. I think it would be much faster if I do compute only on LSF then I/O process out of the batch system.
Currently LSF node usage is almost 99.9% here, and whenever operation jobs are in, my job is killed automatically. During this process, outdata are lost and I have to rerun.

mjy1011:
> > Hi @mjy1011,
> > That was a real tough one... but I think it's working now. Can you please try?
> > @JanStreffing, can you try if the results are bit identical with `release` and this branch? I am getting non-bit identical results in some OpenIFS output files and restarts, and I am wondering if that's normal? (for example, maybe the results in the files are identical but there is something different in the headers)
> 
> Hi ~~ I am waiting for the pre-running job to be finished. I'll try and let you know in a few days. Nice work again!

Just confirmed it is working right! And most of all, saving more than 2 hours, amazing. Thank you very much!

src/esm_runscripts/filelists.py
mandresm: Yes, it does trigger a program exit :)

mandresm:
#approve-changes

mandresm:
> Just confirmed it is working right! And most of all, saving more than 2 hours, amazing. Thank you very much!

You are welcome, and thanks for your patience with this issue :)

mandresm:
#bump

PR #1167, opened 2024-04-25
Fix/wait for files pism

ackerlar:
add wait_for_iterative_coupling to pism recipe


mandresm:
@chrisdane, the tests for this PR seem to show something changing for your AWICM1-REcoM HIST simulations. Can you check if the changes make sense to you or do you think this will break or change your simulations in an undesired way:

https://github.com/esm-tools/esm_tools/actions/runs/8886379988/job/24399704920?pr=1167#step:8:542

PR #1178, opened 2024-05-07
Fix/small fesom cpl fix

ackerlar:
small fixes in awism pism coupling


pgierz:
Why was this one closed?

PR #1186, opened 2024-06-01
Feat/use threads for file movement

mandresm:
Parallel copying of files in one single node using threads.

**TODO**
- [x] Document
- [x] Check if it's logging node and if that's the case, use `cores_per_node/3` as threads
- [x] Testing


src/esm_runscripts/filelists.py
```diff
@@ -940,6 +940,18 @@ def copy_files(config, filetypes, source, target):
     successful_files = []
     missing_files = {}
 
+    # Initialization for parallelization with futures
+    futures = {}
+    if config["general"].get("parallel_file_movements", False) == "threads":
```
pgierz, line 1096: Are the allowed options for `parallel_file_movements` documented somewhere?

src/esm_runscripts/filelists.py
```diff
@@ -999,59 +1011,72 @@ def copy_files(config, filetypes, source, target):
...
+                        # To avoid overwriting in general experiment folder
+                        if skip_intermediate == True:
+                            file_target = avoid_overwriting(
+                                config, file_source, file_target
+                            )
 
-                            successful_files.append(file_source)
-                        except IOError:
-                            logger.error(
-                                f"Could not copy {file_source} to {file_target} for unknown reasons.",
```
pgierz, line 1193: I'm not sure I like that the `try/except` is totally removed. This will all go away when we do file dicts, but I would imagine that it tries to copy everything that it can (in parallel), gathers info about ones that fail, and then prints that info out at the end before an exit. Is that what is done in the implementation below (in `def movement`?)

src/esm_runscripts/filelists.py
```diff
@@ -1244,13 +1267,43 @@ def complete_one_file_movement(config, model, filetype, movement, movetype):
...
+
+
+def movement(func):
+    def inner(config, source_path, target_path):
+        try:
+            func(source_path, target_path)
+            return True
+        except IOError:
+            logger.error(
+                f"Could not copy {file_source} to {file_target} for unknown reasons.",
```
pgierz: The following is more explicit as an error message:
```suggestion
                f"Could not execute movement ({func.__name__}) {file_source} to {file_target} for unknown reasons.",
```

src/esm_runscripts/filelists.py
```diff
@@ -1244,13 +1267,43 @@ def complete_one_file_movement(config, model, filetype, movement, movetype):
...
+            )
+            helpers.print_datetime(config)
+            return False
```
pgierz: Why are you returning True and False here?

pgierz (commented):
Looks good overall, but I still had a few questions.

src/esm_runscripts/filelists.py
mandresm: Yes, this is what is done. All the `ifs` here should not result in a crash, and if they do we want to be able to catch it and not just ignore it with a `try`. The `try` is valuable around the actual movement and that's what us done now in the function `movement`

src/esm_runscripts/filelists.py
mandresm: So that it can be stored as a futures' results and then we can loop over the results and see what's missing (`False`). Then they are added to the list of missing files, so in that sense the behavior is not changed

src/esm_runscripts/filelists.py
mandresm: Just documented the new variable here: https://esm-tools.readthedocs.io/en/feat-use_threads_for_file_movement/esm_variables.html#runtime-variables (`parallel_file_movements`)

Note that `threads` is now the new default (defined in `configs/defaults/general.yaml`).

mandresm:
The real tests are not working yet in levante. I need to figure out why

src/esm_runscripts/filelists.py
```diff
@@ -998,59 +1010,72 @@ def copy_files(config, filetypes, source, target):
...
+                        else:
+                            futures[(file_source, file_target)] = movement_method(
+                                config,
+                                file_source,
+                                file_target,
                             )
-                            helpers.print_datetime(config)
-                            missing_files.update({file_target: file_source})
+
+    for (file_source, file_target), result in futures.items():
```
mandresm: I think this is wrong...

```suggestion
    for (file_source, file_target), future in futures.items():
```

pgierz: I think that looks better, yes. `client.submit` should give you back a `future`, not a resolved future `result`.

pgierz (approved):
Looks better, no suggestions from my side.

mandresm:
Blocked by #972

PR #1188, opened 2024-06-06
feat: transient forcing checker

pgierz:
Adds a checker on the top-level CLI for the transient forcing table. Use as:

```console
$ esm_tools check-transient-forcing [table_path] [year]
```

If not provided, user is prompted. Uses the same machinery as the echam namelist.

### DIFF
--- a/src/esm_runscripts/namelists.py
+++ b/src/esm_runscripts/namelists.py
@@ -13,9 +13,9 @@ import sys
 import warnings
 
 import f90nml
+from loguru import logger
 
 from esm_parser import user_error
-from loguru import logger
 
 
 class Namelist:
@@ -416,6 +416,8 @@ class Namelist:
                         logger.error("echam:")
                         logger.error("    transient_forcing_table: /path/to/your/table")
                         sys.exit(1)
+                    # TODO(PG): This is a bit harsh, but until we get a better idea, dump the error:
+                    logger.error(e)
                     sys.exit(1)
         return config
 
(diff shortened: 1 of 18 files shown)


src/esm_runscripts/namelists.py
```diff
@@ -416,6 +416,9 @@ def echam_transient_forcing(config):
                         logger.error("echam:")
                         logger.error("    transient_forcing_table: /path/to/your/table")
                         sys.exit(1)
+                    # TODO(PG): This is a bit harsh, but until we get a better idea, dump the error:
+                    logger.error(e)
+                    breakpoint()
```
mandresm: Should this go as it is in `release` or do you intend to change it @pgierz?

src/esm_tools/cli.py
```diff
@@ -58,5 +58,37 @@ def create_new_config(name, type):
     return 0
 
 
+# TODO(PG): This doesn't belong in the cli.py but whatever...
```
mandresm, line 61: This is too echam-specific to be part of the ESM-Tools CLI, isn't it? We can discuss it on the monday meeting.

mandresm (commented):
See comments below

src/esm_runscripts/namelists.py
pgierz:
```suggestion
```

pgierz (commented):
Removes breakpoint

mandresm (approved):

pgierz:
#bump

PR #1200, opened 2024-07-09
Fix/which ale

ackerlar:


configs/setups/awiesm/awiesm.yaml
```diff
@@ -358,6 +358,8 @@ fesom:
 
         namelist_changes:
                 namelist.config:
+                        ale_def:
+                                which_ale: "${ALE_scheme}"
```
mandresm, line 362: Is this something new to FESOM 2.5 or 2.6 or was it always in 2.0 and 2.1?

mandresm: Does the parameters defined below still hold?

```yaml
        # ALE_scheme switch
        # -----------------
        # Settings for Gent/McWilliams based upon the selected ALE scheme.
        choose_ALE_scheme:
            linfs:
                K_gm_max: "remove_from_namelist"
                K_hor: 5000.0
                K_gm: 5000.0
            zstar:
                K_gm_max: 2000.0
                K_hor: 3000.0
                K_gm: remove_from_namelist
            "*":
                is: invalid
```

mandresm (changes requested):
Hi @ackerlar, only some questions regarding the ALE scheme configuration.

configs/setups/awiesm/awiesm.yaml
ackerlar: Good point! `K_hor` should go into `namelist.tra/oce_dyn`, `K_gm_max` is fine I think, and `K_gm` is not needed anymore apparently. But I am unsure about the `K_gm_max` setting for `linfs`. Is it `3000` then?

Maybe @JanStreffing or @koldunovn can comment on that?

JanStreffing: Default GM value differs between FESOM2 core2 standalone (2000) and FESOM2 core2 in AWI-CM3 (3000). The retuning from 3000 to 2000 was found to be beneficial for the standalone model while Nikolay worked on the 2.1. release. I tested both for AWI-CM3.1 and found 3000 produced lower biases.

This holds true till at least FESOM 2.5. I am not aware that the parameter is not used anymore..

ackerlar: Thanks @JanStreffing ! You are talking about `K_gm_max`? Scanning through the code, I could not find a parameter `K_gm`. So my question would  be, if `K_gm` actually is `K_gm_max`.

JanStreffing: Right, sorry!
`    k_gm_max = 3000`

pgierz:
Is this still relevant?

PR #1202, opened 2024-07-10
esm_tools version in finished_config

mandresm:
Adds the `esm_tools_version` in the config so that each `finsihed_config` contains the version number of ESM-Tools that was used to be created.

@JanStreffing, this is not as good as the `vcs` feature from @pgierz (because we don't get the right hash of the commit id but still just the version that this branch used to depart from release) but unfortunatelly, vcs is not yet finished and it needs a second thought before finishing its implementation.

In the mean time, it would already be helpful to the NRT people to store the version of ESM-Tools they are using for each simulation (virtually 0 extra time added to ESM-Tools).

@antoniofis10000 @mathanase in effect what this implementation does is that when you look at your `finished_config` you can search for the following line:

```yaml
  esm_tools_version: 6.36.1 # no provenance info
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.36.1
+current_version = 6.37.0
 commit = True
 tag = True
 
--- a/setup.py
+++ b/setup.py
@@ -107,6 +107,6 @@ setup(
     test_suite="tests",
     tests_require=test_requirements,
     url="https://github.com/esm-tools/esm_tools",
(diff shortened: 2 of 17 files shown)


mandresm:
#bump

JanStreffing:
i guess the review is no longer needed. the branch is gone :)
Look good though.

pgierz:
I'll see about revisiting the `vcs` feature, it would be nice to have that actually functioning...

PR #1203, opened 2024-07-15
add some changes to vcs

mandresm:
VCS fixes and improvements.

**TODO**

- [ ] Parallelization of the gitdiff for thrads in a node)
- [x] Hashing of binaries to avoid doing a gitdiff all the time (only possible for codes that actually compile, in the future there might be some codes that have no binary, but instead they are python, that's why I think parallelization of git diff is still useful) -> see the chat in Chatgpt3: https://chatgpt.com/share/5386c188-2a7e-499c-a9ea-54c351e2af0d
- [x] Testing with the virtual environment activated, there were some problems in the past with it
- [ ] Clean the print statements I added

For this PR I will not implement the checking functionality, only the writing of the git status and diff file.

### DIFF
--- a/src/esm_runscripts/prepare.py
+++ b/src/esm_runscripts/prepare.py
@@ -722,8 +722,11 @@ def add_vcs_info(config):
         esm_parser.user_error(
             "esm_tools doesn't know where it's own install location is. Something is very seriously wrong."
         )
-    with open(exp_vcs_info_file, "w") as f:
-        yaml.dump(vcs_versions, f)
+
+    esm_parser.yaml_dump(vcs_versions, exp_vcs_info_file)
+
+    print(exp_vcs_info_file)
+
     return config
 
 
--- a/src/esm_runscripts/prepare.py
+++ b/src/esm_runscripts/prepare.py
@@ -93,7 +93,6 @@ def check_model_lresume(config):
                     model, user_lresume, config, [], []
                 )
             if isinstance(user_lresume, str):
-
                 if user_lresume == "0" or user_lresume.upper() == "FALSE":
                     user_lresume = False
                 elif user_lresume == "1" or user_lresume.upper() == "TRUE":
@@ -355,7 +354,6 @@ def set_leapyear(config):
 
 
 def set_overall_calendar(config):
-
     # set the overall calendar
     if config["general"]["leapyear"]:
         config["general"]["calendar"] = Calendar(1)
@@ -365,7 +363,6 @@ def set_overall_calendar(config):
 
 
 def find_last_prepared_run(config):
-
     calendar = config["general"]["calendar"]
     current_date = Date(config["general"]["current_date"], calendar)
     initial_date = Date(config["general"]["initial_date"], calendar)
@@ -409,7 +406,6 @@ def find_last_prepared_run(config):
 
 
 def set_most_dates(config):
-
     calendar = config["general"]["calendar"]
     if isinstance(config["general"]["current_date"], Date):
         current_date = config["general"]["current_date"]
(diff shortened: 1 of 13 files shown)


mandresm:
@pgierz, I'm putting this into clickup and assign it to you, if you don't have time for it we can reassign it to me in some point.

pgierz:
Hey Miguel, I have something that might be able to handle caching very easily. Have a look at the newest commit.

pgierz:
Cache seems to be working now as I would expect:

```
2024-08-07 10:10:03.938 | CRITICAL | esm_runscripts.prepare:add_vcs_info:700 - Experiment information is being stored for usage under:
2024-08-07 10:10:03.940 | CRITICAL | esm_runscripts.prepare:add_vcs_info:701 - >>> /albedo/work/user/pgierz/SciComp/Projects/esm-tools-vcs/experiments/test-007/run_18500201-18500228/log//test-007_vcs_info.yaml
2024-08-07 10:10:03.940 | CRITICAL | esm_runscripts.prepare:add_vcs_info:703 - >>> CACHE FILE EXISTS!
2024-08-07 10:10:03.940 | CRITICAL | esm_runscripts.prepare:add_vcs_info:704 - >>> Last modification time of cache: 1723017728.2125185
2024-08-07 10:10:03.941 | CRITICAL | esm_runscripts.prepare:add_vcs_info:713 - Locating echam
2024-08-07 10:10:03.952 | DEBUG    | esm_runscripts.helpers:is_older_than:477 - Checking modification times:
2024-08-07 10:10:03.952 | DEBUG    | esm_runscripts.helpers:is_older_than:478 - /albedo/work/user/pgierz/SciComp/Projects/esm-tools-vcs/experiments/test-007/log//test-007_vcs_info.yaml: 1723017728.2125185
2024-08-07 10:10:03.952 | DEBUG    | esm_runscripts.helpers:is_older_than:479 - /albedo/work/user/pgierz/SciComp/Projects/esm-tools-vcs/model_codes/awiesm-2.1/echam-6.3.05p2: 1722867496.347289
2024-08-07 10:10:03.952 | CRITICAL | esm_runscripts.prepare:add_vcs_info:722 - Using cached VCS info for echam
```

src/esm_runscripts/prepare.py
```diff
@@ -708,7 +709,12 @@ def add_vcs_info(config):
             vcs_versions[model] = f"Unable to locate model_dir for {model}."
             continue
         if helpers.is_git_repo(model_dir):
-            vcs_versions[model] = helpers.get_all_git_info(model_dir)
+            if esm_vcs_info_file_cache.is_younger_than(model_dir):
```
mandresm, line 747: We need to think about this, maybe we want to compare it with the age of the binary: is not until one compiles it and uses the binary that the diff needs to be refreshed?

At the same time we need to check whether the binary in the model dir and in the work folder are the same, because ESM-Tools does not update the bin in a simulation unless you run esm_runscripts with the option `--update-files=bin` (or some argument like that)

mandresm (commented):
Nice @pgierz! Really cool and elegant solution :) Just one comment on how we can improve it. Let me know when you are done with that, then I can start the real tests.

src/esm_runscripts/prepare.py
pgierz: Yeah, I wasn't entirely sure about this, I'll think more about it. At the moment, it checks the modification time *of the directory*. The naive assumption that "changing a file in a directory will change the modification time of the directory" is wrong, apparently:

```console
$ mkdir foo
$ gstat foo  # gstat because Darwin != Linux
  File: foo
  Size: 64        	Blocks: 0          IO Block: 4096   directory
Device: 1,14	Inode: 24969841    Links: 2
Access: (0755/drwxr-xr-x)  Uid: (952471260/  pgierz)   Gid: (    0/   wheel)
Access: 2024-08-09 15:19:30.833895558 +0200
Modify: 2024-08-09 15:19:30.833895558 +0200
Change: 2024-08-09 15:19:30.834023725 +0200
 Birth: 2024-08-09 15:19:30.833895558 +0200
$ sleep 10
$ touch foo/a_file
$ gstat foo
  File: foo
  Size: 96        	Blocks: 0          IO Block: 4096   directory
Device: 1,14	Inode: 24970749    Links: 3
Access: (0755/drwxr-xr-x)  Uid: (952471260/  pgierz)   Gid: (    0/   wheel)
Access: 2024-08-09 15:21:26.892522871 +0200
Modify: 2024-08-09 15:21:55.836772107 +0200
Change: 2024-08-09 15:21:55.836772107 +0200
 Birth: 2024-08-09 15:21:26.892522871 +0200
$ echo "my_change" > foo/a_file
$ sleep 3
$ gstat foo
  File: foo
  Size: 96        	Blocks: 0          IO Block: 4096   directory
Device: 1,14	Inode: 24970749    Links: 3
Access: (0755/drwxr-xr-x)  Uid: (952471260/  pgierz)   Gid: (    0/   wheel)
Access: 2024-08-09 15:22:46.780003463 +0200
Modify: 2024-08-09 15:21:55.836772107 +0200
Change: 2024-08-09 15:21:55.836772107 +0200
 Birth: 2024-08-09 15:21:26.892522871 +0200
```

...so I'll need to think of something else. But that is just a matter of changing the logic inside `is_younger_than`. (I like this name, it makes it easy to read 😉)

pgierz: The access time changes, maybe we can use that...

pgierz:
I added an Audit Log, which has all possible info you could possibly wish for to check modification times, access times, and checksums. See in src/esm_runscripts/checksum_helpers.py
 
This, together with the git stuff, should give all possible information someone would want if their model changed halfway during the experiment.

pgierz:
Only thing left to do is parallelising the git diff and checksum parts.

pgierz:
Can we not close this please? This would be super useful...

PR #1209, opened 2024-08-19
ignore provenance for baseline test file diffs

mandresm:
In the baseline tests, when comparing new generated files with the "Truth" ignores the provenance. Comparing the provenance is misleading for example, in the case you added a line at the beginning of a yaml file, instead of getting that variable as the only difference, all the parameters defined in this file pop up as differences because their line number in their provenance has change.

This PR makes sure you only compare values of the parameters, and not their provenance.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.37.2
+current_version = 6.37.3
 commit = True
 tag = True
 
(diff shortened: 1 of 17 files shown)


pgierz (approved):
Looks reasonable. Good catch. 👍

mandresm:
#bump

PR #1187, opened 2024-06-01
feat/jinja templates

mandresm:
After this PR, files that are handled by ESM-Tools can be `jinja2` templates and will be rendered at the time of the copying.

This was a request by @JanStreffing to be able to handle variable substitution in `xml` files. Jan, could you test it? You can do your xml and yaml modifications in a new branch departing from this one. We can merge in release that branch right after this one gets merged, and there should be no conflicts.

### How does it work

ESM-Tools automatically tries to resolve jinja templates during movements. It searches for sources finishing with `.j2` and then it knows that instead of copying/linking/moving it needs to render the template and produce the new target file. We pass `config` to `jinja`, that means you just need to specify the variable you want to subsitute in the right format with the full path separated by points:

**xios.yaml**
```diff
config_sources:
-        domain_def: ${xml_dir}/domain_def.xml
+        domain_def: ${xml_dir}/domain_def.j2
        field_def: ${xml_dir}/field_def.xml
        file_def: ${xml_dir}/file_def.xml
```
**domain_def.xml**
```xml
<domain_definition>
  <!-- Definition of the native domain of the model -->
  <domain id="reduced_gaussian" long_name="reduced Gaussian grid" type="gaussian" />

  <!-- Definition of regular Gaussian domains -->
  <domain_group id="regular_domains" type="rectilinear" >
    <domain id="regular" long_name="regular grid" ni_glo="{{xios.ni_glo}}" nj_glo="{{xios.nj_glo}}" >
      <generate_rectilinear_domain />
      <interpolate_domain order="1" write_weight="true" />
    </domain>
  </domain_group>
</domain_definition>
```

or with python dict syntax

```xml
<domain_definition>
  <!-- Definition of the native domain of the model -->
  <domain id="reduced_gaussian" long_name="reduced Gaussian grid" type="gaussian" />

  <!-- Definition of regular Gaussian domains -->
  <domain_group id="regular_domains" type="rectilinear" >
    <domain id="regular" long_name="regular grid" ni_glo="{{xios['ni_glo']}}" nj_glo="{{xios['nj_glo']}}" >
      <generate_rectilinear_domain />
      <interpolate_domain order="1" write_weight="true" />
    </domain>
  </domain_group>
</domain_definition>
```

**TODO**

- [x] Add documentation
- [x] Test
- [x] Error handling for not defined extension

PR #1223, opened 2024-09-24
Feat/file movements parallelization

mandresm:
This was reviewed already in #1186 but the target branch was not `release`. The idea was that in this branch we will also add the dask parallelization, but we need this changes for the jinja feature asap (#1187)


configs/defaults/general.yaml
```diff
@@ -4,3 +4,5 @@ intermediate_movements:
 - "config"
 - "bin"
 - "input"
+compute_file_checksums: True
+parallel_file_movements: threads
```
mandresm: For the time being let's keep this both false. Once dask parallelization is available we can turn them on by default.
```suggestion
compute_file_checksums: False
parallel_file_movements: False
```

mandresm:
#approve-changes

mandresm:
#bump

PR #1238, opened 2024-10-21
fix the esm_master logger

mandresm:
Fixes the logger for the esm_master.

As it currently was, none of the logger messages coming from `esm_runscripts` functions were printed because there was no definition on where to print this messages for `esm_master`.


pgierz (approved):

pgierz:
#bump

Issue #1239, opened 2024-10-23
AWIESM2 output data went to "unknown" folder


Mccino:
**Describe the bug**
ECHAM output data went to the "Unkown" folder for some simulations and some related questions.

**To Reproduce**
The correct simulations can be found: **_/home/a/a270198/work_ab0246/AWIEMS2/HI_EXP/Pre_HIST_**
The broken simulations can be found: **_/home/a/a270198/work_ab0246/AWIEMS2/HI_EXP/RES_HIST_**

**System (please complete the following information):**
 - Supercomputer: Levente
 - Version: esm_tools, version 6.29.0

**Additional context**
There is another problem with the broken simulations. ECHAM is a restart simulation, and the output generates two files with similar names. For example, for the year 2010, you can find "RES_HIST_201001.01_echam.nc" and "RES_HIST_201001.02_echam.nc" in the unknown folder. However, only the file with "02" can be read using CDO or ncdump. I’m wondering if anyone can provide some hints on this, or if I did something incorrectly when restarting ECHAM.

pgierz:
Hi @Mccino,

output being moved to the `unknown` folder happens if the file name is not matched by one of the output patterns. If you have a look at the `finished_config.yaml` (in the `config` folder), you'll see which files it knows about. For me, it looks something like this:

```console
$ yq .echam.outdata_targets basic-002_finished_config.yaml
aclcim_nc: /albedo/work/user/pgierz/SciComp/Tutorials/AWIESM_Basics/experiments/basic-002/outdata/echam/basic-002_185901.01_aclcim.nc # no provenance info
g3b1hi_nc: /albedo/work/user/pgierz/SciComp/Tutorials/AWIESM_Basics/experiments/basic-002/outdata/echam/basic-002_185901.01_g3b1hi.nc # no provenance info
g3bday_nc: /albedo/work/user/pgierz/SciComp/Tutorials/AWIESM_Basics/experiments/basic-002/outdata/echam/basic-002_185901.01_g3bday.nc # no provenance info
g3bid_nc: /albedo/work/user/pgierz/SciComp/Tutorials/AWIESM_Basics/experiments/basic-002/outdata/echam/basic-002_185901.01_g3bid.nc # no provenance info
g3bim_nc: /albedo/work/user/pgierz/SciComp/Tutorials/AWIESM_Basics/experiments/basic-002/outdata/echam/basic-002_185901.01_g3bim.nc # no provenance info
[...]

PR #1240, opened 2024-10-24
Feat/echam streams

pgierz:
Automatically create ECHAM streams from mvstreamctl filetags. 

Should be useful for #1239

### DIFF
--- a/src/esm_runscripts/echam.py
+++ b/src/esm_runscripts/echam.py
@@ -34,5 +34,12 @@ def append_namelist_dependent_sources(config):
     jsbach_streams = config["jsbach"]["streams"]
     ignore_these_tags = econfig.get("ignore_tags", jsbach_streams)
     mvstream_tags = [tag for tag in mvstream_tags if tag not in ignore_these_tags]
-    mvstream_dict = {tag: f"{expid}*_{tag}" for tag in mvstream_tags}
+    mvstream_dict = {tag: f"{expid}*{tag}" for tag in mvstream_tags}
+    if namelist["runctl"].get("out_filetype") == 2:
+        # Using NetCDF Outputs:
+        mvstream_dict = {k: v + ".nc" for k, v in mvstream_dict.items()}
+    logger.info("Updating outdata_sources...")
+    for k, v in mvstream_dict.items():
+        logger.info(f"{k}: {v}")
     econfig["outdata_sources"].update(mvstream_dict)
+    logger.info("...done!")
--- a/src/esm_runscripts/echam.py
+++ b/src/esm_runscripts/echam.py
@@ -18,15 +18,21 @@ def _get_mvstream_tags_from_namelist(namelist):
     for chapter, contents in namelist.items():
         if chapter == "mvstreamctl":
             tag = contents.get("filetag")
-            mvstream_tags.append(tag)
+            if tag is not None:
+                mvstream_tags.append(tag)
     return mvstream_tags
 
 
 def append_namelist_dependent_sources(config):
     expid = config["general"]["expid"]
-    work_dir = config["general"]["work_dir"]
     econfig = config["echam"]
-    namelist = econfig["namelist_objs"]
+    try:
+        namelist = econfig["namelist_objs"]
+    except KeyError:  # Namelists not yet loaded...
+        namelist = f90nml.read(econfig["namelist_dir"] + "/namelist.echam")
     mvstream_tags = _get_mvstream_tags_from_namelist(namelist)
-    mvstream_dict = {tag: f"{work_dir}/{expid}*_{tag}" for tag in mvstream_tags}
-    config["outdata_sources"].update(mvstream_dict)
+    jsbach_streams = config["jsbach"]["streams"]
+    ignore_these_tags = econfig.get("ignore_tags", jsbach_streams)
+    mvstream_tags = [tag for tag in mvstream_tags if tag not in ignore_these_tags]
+    mvstream_dict = {tag: f"{expid}*_{tag}" for tag in mvstream_tags}
+    econfig["outdata_sources"].update(mvstream_dict)
--- a/src/esm_runscripts/echam.py
+++ b/src/esm_runscripts/echam.py
@@ -1,0 +1,32 @@
+import f90nml
+from loguru import logger
+
+
+def _get_mvstream_tags_from_namelist(namelist):
+    mvstream_tags = []
+    if not isinstance(namelist, f90nml.namelist.Namelist):
+        try:
+            namelist = f90nml.read(namelist)
+        except FileNotFoundError:
+            logger.error(f"Namelist specified by {namelist} could not be found")
+        except TypeError as e:
+            logger.error(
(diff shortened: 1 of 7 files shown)


src/esm_runscripts/echam.py
```diff
@@ -0,0 +1,176 @@
...
+    - The output file type is checked, and if it is NetCDF (indicated by
+      ``out_filetype`` == 2), the file extension `.nc` is appended to the tags.
+    - The function logs the updates made to the ``outdata_sources``.
+    """
+    expid = config["general"]["expid"]
+    econfig = config["echam"]
+    try:
+        namelist = econfig["namelist_objs"]
+    except KeyError:  # Namelists not yet loaded...
+        namelist = f90nml.read(econfig["namelist_dir"] + "/namelist.echam")
```
mandresm:
```suggestion
        namelist = f90nml.read(f"{econfig['namelist_dir']}/namelist.echam")
```

src/esm_runscripts/echam.py
```diff
@@ -0,0 +1,176 @@
...
+    mvstream_tags = _get_mvstream_tags_from_namelist(namelist)
+    jsbach_streams = config["jsbach"]["streams"]
```
mandresm:
```suggestion
    jsbach_streams = config["jsbach"].get("streams", [])
```
Isn't it better/safer to use a get here?

src/esm_runscripts/echam.py
```diff
@@ -0,0 +1,176 @@
...
+    ignore_these_tags = econfig.get("ignore_tags", jsbach_streams)
```
mandresm: Why is the default ignore ignore jsbach_streams? Is it so that the jsbach streams are moved to the jsbach directory instead of echam?

I find this behavior a bit confusing because if I declare an `echam.ignore_tags` then the jsbach streams won't be ignored anymore, so there is a big change of behavior behind the scenes. What do you think @pgierz? Maybe I am not understanding this part...

src/esm_runscripts/echam.py
```diff
@@ -0,0 +1,176 @@
...
+
+    for chapter, contents in namelist.items():
+        if chapter == "mvstreamctl":
+            tag = contents.get("filetag")
+            if tag is not None:
+                mvstream_tags.append(tag)
+    return mvstream_tags
+
+
+def append_namelist_dependent_sources(config):
```
mandresm, line 136: Shouldn't this happen for `echam` but also `jsbach`? Doesn't `jsbach` also have its own streams?

mandresm (changes requested):
Nice work @pgierz :) Hopefully will make many echam headaches go away.

Just a few comments

src/esm_runscripts/echam.py
```diff
@@ -0,0 +1,176 @@
...
+    mvstream_tags = [tag for tag in mvstream_tags if tag not in ignore_these_tags]
+    mvstream_dict = {tag: f"{expid}*{tag}" for tag in mvstream_tags}
+    if namelist["runctl"].get("out_filetype") == 2:
+        # Using NetCDF Outputs:
+        mvstream_dict = {k: v + ".nc" for k, v in mvstream_dict.items()}
+    logger.info("Updating outdata_sources...")
+    for k, v in mvstream_dict.items():
+        logger.info(f"{k}: {v}")
+    econfig["outdata_sources"].update(mvstream_dict)
+    logger.info("...done!")
```
mandresm: Shouldn't all these `logger.info` be `logger.debug`? It gets very verbose.

src/esm_runscripts/echam.py
pgierz: yes, that is safer, the user *might* have the silly idea to remove the default JSBACH streams.

src/esm_runscripts/echam.py
pgierz: This is a little bit convoluted: jsbach and echam are controlled by the same namelist, so I would need some way to filter out which tags go to which model. In principle this would just be defined by which model config they come from.

src/esm_runscripts/echam.py
pgierz: Good point, see new commit (coming later today)

pgierz: Fixed in [`d6aab9`](https://github.com/esm-tools/esm_tools/pull/1240/commits/d6aab9175ff008236d79cc4e8f2e9a20aa79534b).

src/esm_runscripts/echam.py
pgierz: See https://github.com/esm-tools/esm_tools/pull/1240/commits/509ac21d102e41dcaaa1e9cbe95dca73834cceb7.

pgierz:
The latest few commits use the extremely useful feature of being able to write tests directly in your docstrings under the examples section. The included files are referenced directly from the root of the repo in the CI actions. Check it out 😄 

> src/esm_runscripts/echam.py::esm_runscripts.echam._get_mvstream_tags_from_namelist PASSED [ 71%]
src/esm_runscripts/jsbach.py::esm_runscripts.jsbach._get_comments_for_streams PASSED [ 85%]

mandresm:
Some of the dryruns in esm_tests are failing

src/esm_runscripts/echam.py
mandresm: >  jsbach and echam are controlled by the same namelist

Do you mean that they can be controlled from the same namelist? If that is the case, then I understand the rest of your reply and the new `jsbach.py`

mandresm:
Once the CI tests work, I would recommend running `esm_tests` in levante with this branch to check that also actual simulations run correctly. I can take care of that to unload you a bit @pgierz. Just let me know when I can go ahead and run the real tests

src/esm_runscripts/echam.py
pgierz: Yes, the output generated from `jsbach` is defined in `mvstreamctl` chapters of `namelist.echam` (since JSBACH technically speaking is just a submodule inside of ECHAM), and does not have any output control in it's own namelist.

src/esm_runscripts/echam.py
pgierz: Fixed in https://github.com/esm-tools/esm_tools/pull/1240/commits/82f649d38a7fd9d219674d3cbe1e63be5f43803e

pgierz:
Should be good to go. The failing levante tests mainly fail due to changes in how the streams get picked up, as well as new main keys in the echam config, e.g.:

```diff
  	Differences in run/awicm/awicm1-CMIP6-initial-monthly//run_18500101-18500131/config//awicm1-CMIP6-initial-monthly_finished_config.yaml:
  		--- 
  		+++ 
  		@@ -118,6 +118,7 @@
  		   forcing_dir: /pool/data/ECHAM6//input/r0007/T127
  		   greenhouse_dir: /pool/data/ECHAM6/
  		   namelist_dir: <HOME_DIR>/esm_tools/namelists//echam/6.3.04p1/PI-CTRL
  		+  has_namelist_streams: true
```

Issue #1236, opened 2024-10-21
esm_master double git clone 


JanStreffing:
While trying to build a new setup version awicm3-v3.3 with FESOM2.6 I get the following error:

```
ERROR: destination already exists
---------------------------------
Some coponents couldn't be downloaded because their destination paths already exists (see list below). If you want to download again the model consider deleting that folder. If instead, you want to keep that, you can use other esm_master commands (e.g. esm_master comp-<model>-<version>). Destinations already present:
- awicm3-v3.3/oasis
```

I found that the clone command for oasis happens twice:

```
esm_master install-awicm3-v3.3
    Executing commands in this order:
         mkdir -p awicm3-v3.3
         cd awicm3-v3.3
         git clone -b awicm-3.1 https://gitlab.dkrz.de/modular_esm/oasis3-mct.git oasis
         git clone -b awicm-3.1 https://gitlab.dkrz.de/modular_esm/oasis3-mct.git oasis
         git clone -b master https://gitlab.dkrz.de/ec-earth/xios-2.5.git xios
         git clone -b awicm3-frontiers-freeze-candidate-1 https://gitlab.dkrz.de/ec-earth/runoff-mapper.git rnfmap
         git clone -b awicm-3.2 https://gitlab.dkrz.de/ec-earth/oifs-43r3.git oifs-43r3
         chmod 700 -R oifs-43r3
         git clone -b 2.6 https://github.com/FESOM/fesom2.git fesom-2.6
```

Of course, the second one will fail. I remember I already had this problem once before, and it was solved, but I can not recall how, nor do I find an issue here from last time.

mandresm:
What branch are you using?

JanStreffing:
Right.. https://github.com/esm-tools/esm_tools/tree/feat/awicm_v33_fesom_26

mandresm:
I'll look at this tomorrow.

JanStreffing:
I just ran into this one again. Did you have a look at it already?

mandresm:
Hi @JanStreffing, I think I just fixed and pushed it to the branch `feat/awicm_v33_fesom_26`. I am also pull-requesting the bug fix in `release` (#1245)

JanStreffing:
Ty for the fix, that works.
After doing a git pull on the branch i now have:
git status:
```
On branch feat/awicm_v33_fesom_26
Your branch is up to date with 'origin/feat/awicm_v33_fesom_26'.

Changes not staged for commit:
[...]

PR #1245, opened 2024-12-02
fix/repeated git clones

mandresm:
fix a bug where repeated git commands were allowed in esm_master (#1236)


pgierz (approved):
👍🏻

mandresm:
I am fixing the problems with the test container

pgierz:
Hm, odd. 404 when installing `jq`? What do we use `jq` for? I thought the apt server was down, but forcing a rerun doesn't resolve the problem.

I haven't looked at the test container in a very long time. Try switching to use just a regular Ubuntu...? I am not sure why we are using a custom one anyway. If we want to mimic the HPC more closely, we could use a Rocky Linux container.

mandresm:
Merging from release with the CI fixed

mandresm:
#bump !!!!!!

PR #1246, opened 2024-12-03
Fix echam hist ssp forcing

chrisdane:
Hi

In the current release (`595f0b4b`), `echam.yaml:allow_hist2ssp_transition` does not seem to work (for year 2015, i.e. the first ssp-year, not all forcing files are found in an `awicm-1.0-recom` experiment on levante).

In addition, the rationale in `echam.yaml`: 
```
choose_scenario:
        HIST:
                forcing_files:
                        ozone: ozone_hist_scenario
                        greenhouse: greenhouse_hist_scenario
                        MAC-SP: MAC-SP_hist_scenario
        cmip6hist:
                scenario_type: cmip6
        ssp126:
                scenario_type: cmip6
```
is imho not the best choice, as it leads to e.g.
```
choose_scenario_type:
        cmip6:
                forcing_files:
                        ozone: ozone_hist_scenario
                        greenhouse: greenhouse_hist_scenario
                        MAC-SP: MAC-SP_hist_scenario
```
which I find rather confusing (are they hist or scenario or both?!), as well as to
```
ls namelists/echam/<echam_version>
...
HIST/
...
PI-CTRL/
cmip6/
```
, which does not make too much sense since `cmip6` is not an experiment/scenario.

So this PR is merely a restructuring of how paths of echam+jsbach forcing files+namelists are constructured. I tested it in many CMIP6 historical and ssp* experiments with the `awicm-1.0-recom` setup.

Over time some small additions in recom were added as well. Please excuse the delay in pushing this.

Kind regards,
Chris


mandresm:
I need to book time for this PR. It will have a big impact in multiple research groups so it cannot be just merged. I will only have this ammount of time by the end of January.

PR #1193, opened 2024-06-17
Draft: Feat/awicm3 v3.4

JanStreffing:
superseeds https://github.com/esm-tools/esm_tools/pull/1124 that became obsolete as the HEAD is testing out ideas that were rejected.

PR #1251, opened 2024-12-09
Feat/awicm3 v3.3 and v3.4

JanStreffing:
Superseeds: https://github.com/esm-tools/esm_tools/pull/1193


namelists/fesom2/2.5/awicm3/v3.2/namelist.io
```diff
@@ -35,8 +35,8 @@ io_list =  'sst       ',1, 'm', 4,
            'MLD3      ',1, 'm', 4,
            'tx_sur    ',1, 'm', 4,
            'ty_sur    ',1, 'm', 4,
-           'temp      ',1, 'y', 4,
-           'salt      ',1, 'y', 8,
+           'temp      ',1, 'm', 4,
```
JanStreffing, line 38: This change is intentional. Needed for CMIPtool. I had too many bad runs with only yearly output.

JanStreffing: Also only in fesom namelist template for AWI-CM3

JanStreffing:
Checks are giving errors like:

SUBMITTING (100.0%) awicm3/awicm3-v3.2-TCO95L91-CORE2_initial

```
git submodule init

git submodule update

  File "/opt/conda/envs/esm_tools/bin/esm_tests", line 33, in <module>
    sys.exit(load_entry_point('esm-tools', 'console_scripts', 'esm_tests')())
  File "/__w/esm_tools/esm_tools/src/esm_tests/cli.py", line 114, in main
    results = format_results(info)
  File "/__w/esm_tools/esm_tools/src/esm_tests/output.py", line 435, in format_results
    if not state["comp"]:
KeyError: 'comp'
Error: Process completed with exit code 1.
```

configs/machines/levante.yaml
```diff
@@ -189,36 +190,72 @@ choose_compiler_mpi:
...
             TBBROOT: /sw/spack-levante/intel-oneapi-compilers-2022.0.1-an2cbq/tbb/2021.5.0
             TBBMALLOC_DIR: $TBBROOT/lib/intel64/gcc4.8
             MPI_HOME: /sw/spack-levante/openmpi-4.1.2-yfwe6t
             MPI_ROOT: /sw/spack-levante/openmpi-4.1.2-yfwe6t
             mpi_ROOT: /sw/spack-levante/openmpi-4.1.2-yfwe6t
             MPI_DIR: /sw/spack-levante/openmpi-4.1.2-yfwe6t
             mpi_DIR: /sw/spack-levante/openmpi-4.1.2-yfwe6t
-            LD_LIBRARY_PATH: $MPI_HOME/lib:$TBBMALLOC_DIR:$LD_LIBRARY_PATH
+            LD_LIBRARY_PATH: $MPI_HOME/lib:$NetCDF_C_ROOT/lib:$NetCDF_Fortran_ROOT/lib:$HDF5ROOT/lib:$AEC_ROOT/lib:$TBBMALLOC_DIR:$LD_LIBRARY_PATH
+            LD_LIBRARY_PATH[(0)]: /sw/spack-levante/intel-oneapi-mkl-2022.0.1-ttdktf/mkl/2022.0.1/lib/intel64:$LD_LIBRARY_PATH
```
JanStreffing: Test if this is needed. It looks wrong to add LD_LIBRARY_PATH for intel compiler in system_gnu_libs block

configs/machines/levante.yaml
```diff
@@ -278,6 +315,50 @@ choose_iolibraries:
...
+    system_intel_libs: # Attempt to make intel work for OpenIFS 48r1, currently disfunctional
+        add_module_actions:
+            - "load ${mod_netcdf_c}"
+            - "load ${mod_netcdf_f}"
+            - "load ${mod_hdf5}"
+            - "load libaec/1.0.5-intel-2021.5.0"
+            - "load fftw/3.3.10-gcc-11.2.0"
+
+        add_export_vars:
+            IO_LIB_ROOT: "/work/ab0246/HPC_libraries/intel-oneapi-compilers/2022.0.1-gcc-11.2.0/openmpi/4.1.2-intel-2021.5.0"
```
JanStreffing: Check if this is needed. Looks unused below.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -325,35 +409,13 @@ oifs:
...
-                                compiler_mpi: intel2022_ompi2022
-                        blogin:
-                                iolibraries: geomar_libs
-                        glogin:
-                                iolibraries: geomar_libs
-                        levante:
-                                iolibraries: system_libs
                 add_export_vars:
                         # Turn on FESOM coupling
                         FESOM_USE_CPLNG: '"active"'
```
JanStreffing: Check if still used anywhere. Else, remove.

namelists/oifs/48r1/xios/context_ifs.xml
```diff
@@ -18,8 +18,8 @@
     </variable_group>
 
     <variable_group id="fullpos_freq">
-      <variable id="nfrhis" name="NFRHIS" type="int"> -6 </variable>
-      <variable id="nfrpos" name="NFRPOS" type="int"> -6 </variable>
+      <variable id="nfrhis" name="NFRHIS" type="int"> 6 </variable>
```
JanStreffing, line 21: Turn into jinja tempate and fill in from esm-tools

pgierz (changes requested):
Small changes found while discussing the merge with @JanStreffing, see his comments. He will work on some last fixes.

This still needs the `further_reading` behaviour fixes as noted in #1285 and #1253 before it can reliably be used across various systems.

JanStreffing:
Todo add correct module list for 48r1 here

JanStreffing:
For OpenIFS 48r1 we need a python available, so that the preprocessor can do [fypp](https://fypp.readthedocs.io/en/stable/fypp.html). However, on levante, if I load the python3 module last, as I would happen when it comes from `oifs48.env.yaml` this overwrites any loaded `netcdf` module. Therefore, the  python3 module has to be loaded first, and the netcdf from python3 is then overwritten by the module load netcdf.

Is there a feature for this already? Something like export vars like this:
```
LD_LIBRARY_PATH[(3)]
```

pgierz:
I think the logic for the `[(#)]` applies to all `export_vars` key/value pairs, so the `LD_LIBRARY_PATH` "should" fit that pattern as well. 

Do you have a reproducible counter example?

src/esm_runscripts/oasis.py
```diff
@@ -50,7 +50,10 @@ def __init__(
             else:
                 self.namcouple += [" $NLOGPRT", "           " + "1 -1", " $END"]
         else:
-            self.namcouple += [" $NLOGPRT", "           " + str(debug_level), " $END"]
+            if mct_version >= (5, 0):
+                self.namcouple += [" $NLOGPRT", "           " + str(debug_level) + " 0 0", " $END"]
+            else:
+                self.namcouple += [" $NLOGPRT", "           " + "1", " $END"]
```
mandresm, line 56: Just to confirm @JanStreffing, you want that here the debug level is always 1? It might be good to keep it as it was so that we can have control of it externally and that both versions behave the same. Does something speak against it?

configs/setups/oifsamip/oifsamip.yaml
```diff
@@ -32,37 +37,41 @@ general:
...
-                - oasis3mct/oasis3mct.env.yaml
+        choose_general.version:
+                "cy43":
+                        further_reading:
+                                - oifs/oifs43.env.yaml
+                                - oasis3mct/oasis3mct.env.yaml
+                "cy48":
+                        further_reading:
+                                - oifs/oifs48.env.yaml
+                                - xios/xios.env.yaml
```
mandresm, line 60: If `oifs`, `oasis` and `xios` yaml files load their own further reading (which I think they do), then you don't need this choose block here anymore.

configs/machines/levante.yaml
```diff
@@ -72,13 +72,18 @@ computer:
     pool_dir:  "/pool/data"
 
     # kh 15.02.22 preferred on levante (at the moment), on mistral intelmpi was the default
-    compiler_mpi: intel2022_openmpi
+    compiler_mpi: gnu11_ompi4
```
mandresm: I'm going to revert this change and leave `intel2022_openmpi` as default. With `v3.4` `gnu11` should be correctly selected now

configs/machines/levante.yaml
```diff
@@ -72,13 +72,18 @@ computer:
...
+    #compiler_mpi: intel2022_openmpi
+    #compiler_mpi: intel2022_impi2021
 
     # seb-wahl: choose libaries below, default is the system libs via modules
-    iolibraries: system_libs
+    iolibraries: system_gnu_libs
+    #iolibraries: system_intel_libs
+    #iolibraries: system_intel_impi_libs
```
mandresm: Same as above

configs/machines/levante.yaml
mandresm: Confirmed, changing this does not affect to awicm3 3.3 or 3.4 envs.

configs/components/oasis3mct/oasis3mct.yaml
```diff
@@ -125,6 +135,45 @@ oasis3mct:
...
+        #cp INSTALL_OASIS.ESMTOOLS/build-static/lib/psmile.MPI1/*.mod build/lib/psmile/.
+        #cp INSTALL_OASIS.ESMTOOLS/build-static/lib/scrip/*mod build/lib/psmile/scrip/.
+        #cp INSTALL_OASIS.ESMTOOLS/build-static/lib/mct/*mod build/lib/psmile/mct/.
+        #cp INSTALL_OASIS.ESMTOOLS/lib/libmct.a build/lib/psmile/mct/.
+        #cp INSTALL_OASIS.ESMTOOLS/lib/libmpeu.a build/lib/psmile/mct/.
+        #cp INSTALL_OASIS.ESMTOOLS/lib/libpsmile.a build/lib/psmile/.
+        #cp INSTALL_OASIS.ESMTOOLS/lib/libscrip.a build/lib/psmile/scrip/.
+
+
+
```
mandresm, line 176: If commented, can it then be removed?

```suggestion
```

configs/components/oifs/oifs43.env.yaml
```diff
@@ -47,19 +53,14 @@ computer:
...
-            SZIPROOT: "/sw/spack-levante/libaec-1.0.5-gij7yv"
-            HDF5ROOT: "/sw/spack-levante/hdf5-1.12.1-tvymb5"
-            HDF5_ROOT: $HDF5ROOT
-            NETCDFROOT: "/sw/spack-levante/netcdf-c-4.8.1-2k3cmu"
-            NETCDFFROOT: "/sw/spack-levante/netcdf-fortran-4.5.3-k6xq5g"
-            ECCODESROOT: "/work/ab0246/HPC_libraries/intel-oneapi-compilers/2022.0.1-gcc-11.2.0/openmpi/4.1.2-intel-2021.5.0"
 
             PATH: "$ECCODESROOT/bin:$PATH"
-            LD_LIBRARY_PATH[(3)]: "$ECCODESROOT/lib:$LD_LIBRARY_PATH"
+            LD_LIBRARY_PATH[(3)]: "$LD_LIBRARY_PATH:$ECCODESROOT/lib"
```
mandresm: Just to confirm that this is intentional. In this way `ECCODESROOT/lib` has less preference than the libs defined in `LD_LIBRARY_PATH`, right? Is this the intended behavior? If not you can reverted with the following suggestion.

```suggestion
            LD_LIBRARY_PATH[(3)]: "$ECCODESROOT/lib:$LD_LIBRARY_PATH"
```

configs/components/oifs/oifs43.cmip.yaml
```diff
@@ -0,0 +1,24 @@
+oifs:
```
mandresm, line 1: For every new file I will be passing now a yaml formater that indents to 4. I will leave non-new files untouch in this PR so that we can see the diffs clearly.

In the future the yaml formater will run in the CI and make sure that everything is formatted correctly.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -91,22 +95,45 @@ general:
...
+            major_version: v3.3
+            couplings:
+            - fesom-2.6+oifs-43r3-awicm-3.2+xios-2.5
+            add_include_models:
+            - xios
+            add_further_reading:
+            - oifs/oifs43.env.yaml
+            - xios/xios.env.yaml
+            - rnfmap/rnfmap.env.yaml
+            - fesom/fesom.env.yaml
```
mandresm: This is not needed anymore. One can remove it safely because they are now loaded form the component and they are kept. I will take care of removing them in a commit soon.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -552,8 +681,16 @@ xios:
 rnfmap:
         pool_dir: ${general.pool_dir}
         runoff_maps_name: _${fesom.resolution}
-
-        include_env_from_component_files: False
+        choose_general.major_version:
+                "v3.3":
+                        # TODO: change this to a control on the patch version general.version
```
mandresm: This needs to be dealt with.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -672,7 +808,7 @@ oasis3mct:
                                         - 'A_CurX:A_CurY <--gauswgt_i-- u_feom:v_feom'
                 'v3.2':
                         version: 6.0-YAC
-                        mct_version: 4.0
+                        mct_version: 6.0
```
mandresm: Is this correct? Bumping up the mct_version of the `v3.2`?

configs/setups/awicm3/awicm3.yaml
```diff
@@ -724,6 +896,12 @@ oasis3mct:
                                 rstos.nc:
                                         - 'A_SST:A_Ice_frac:A_Snow_thickness:A_Ice_temp:A_Ice_albedo <--gauswgt_i-- sst_feom:sie_feom:snt_feom:ist_feom:sia_feom'
 
+        # TODO: change this to a control on the patch version general.version
```
mandresm: Same as above, we have to discuss how we want to control this.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -952,23 +1130,30 @@ computer:
...
-            blogin:
-                iolibraries: geomar_libs
-                compiler_mpi: intel2019_ompi
-            glogin:
-                iolibraries: geomar_libs
-                compiler_mpi: intel2019_ompi
-            levante:
-                iolibraries: system_libs
-                compiler_mpi: intel2022_openmpi
+        choose_general.major_version:
```
mandresm, line 983: I think there is some redundancy here with the oifs envs yamls. Now conflicts will be catch, but it might be worth it have it them in either one place or another.

mandresm (changes requested):
There are some minor changes, and I have some questions, but in general it's all pretty straight forward. I now need to look at the integration tests and see if there are any unexpected changes, but we are getting very close.

src/esm_runscripts/oasis.py
JanStreffing: I think this is fine.

configs/machines/levante.yaml
JanStreffing: correct, this was just the workaround from before you fixed things; back when the switches were not working.

configs/components/oasis3mct/oasis3mct.yaml
JanStreffing: Kinda? I guess we could dig it out again from the git history if we ever want to use oasis with yac interpolation again.

configs/components/oifs/oifs43.env.yaml
JanStreffing: I forgot exactly why this was done. Especially for 43 it does not seem to make sense to me. I reverted back.

configs/components/oifs/oifs43.cmip.yaml
JanStreffing: great stuff!

configs/setups/awicm3/awicm3.yaml
JanStreffing: Meaning no more major_version? Everything is choose_version?

configs/setups/awicm3/awicm3.yaml
JanStreffing: Yes, v3.2 was with oasis 6.0 and yac. in 3.2.2 i went back to 4.0.

JanStreffing: But then again this never really worked. So it should not be used. too many people use v3.2 over 3.2.2. I will revert it.

JanStreffing: i also deleted the line about using version 6.0-YAC above..

JanStreffing: Perhaps should be tested with v3.2

configs/setups/awicm3/awicm3.yaml
JanStreffing: lets have a chat next week.

configs/setups/awicm3/awicm3.yaml
JanStreffing: needs to be looked at? or do we wait until there are conflicts, if there are any?

configs/machines/levante.yaml
```diff
@@ -79,6 +79,7 @@ computer:
 
     module_actions:
             - "purge"
+            - "load python3"
```
mandresm: Let's be specific about the version, because otherwise the users will install on a python version, and then wonder why esm_runscripts is not running anymore their tidying job after python load in the .run.
```suggestion
            - "load python3/2023.01-gcc-11.2.0"
```

This should be fixed once we sort out the python environments into a smarter way, but for now we need to be careful with the python versions

configs/setups/awicm3/awicm3.yaml
mandresm: I'll test v3.2

configs/components/oasis3mct/oasis3mct.yaml
mandresm: If this is something you might want to use in the future, then let's keep it.

mandresm:
None AWICM3 compilation seems to work in Juwels. It's probably a pretty simple fix, however, that would keep delaying this merge. @JanStreffing, should we treat this into a different issue/PR?

mandresm:
Including Python in the module loads seems to be breaking the linking of some older versions of FESOM. What is the reason behind including the module loading of Python @JanStreffing? Is it to be able to install oifs-48r1 with the bundle? If that's the case, would you be okay with me moving this into the `oifs-48r1.env.yaml`?

JanStreffing:
> None AWICM3 compilation seems to work in Juwels. It's probably a pretty simple fix, however, that would keep delaying this merge. @JanStreffing, should we treat this into a different issue/PR?

I'll spend half an hour looking into it.

JanStreffing:
> Including Python in the module loads seems to be breaking the linking of some older versions of FESOM. What is the reason behind including the module loading of Python @JanStreffing? Is it to be able to install oifs-48r1 with the bundle? If that's the case, would you be okay with me moving this into the `oifs-48r1.env.yaml`?

Yes, oifs48 need python at build time for some preprocessor bits. I don't think this will make a difference? It would still appear in fesom2 comp script in awicm3, no?

mandresm:
> > Including Python in the module loads seems to be breaking the linking of some older versions of FESOM. What is the reason behind including the module loading of Python @JanStreffing? Is it to be able to install oifs-48r1 with the bundle? If that's the case, would you be okay with me moving this into the `oifs-48r1.env.yaml`?
> 
> Yes, oifs48 need python at build time for some preprocessor bits. I don't think this will make a difference? It would still appear in fesom2 comp script in awicm3, no?

fesom-2.6 doesn't have a problem with that :)

configs/components/fesom/fesom-2.6.yaml
```diff
@@ -43,6 +43,9 @@ fesom:
       "2.6.8":
         branch: "2.6.8"
 
+      "2.6.9":
+        branch: "2.6.8"
```
mandresm: This is incorrect right?

```suggestion
      "2.6.9":
        branch: "2.6.9"
```

JanStreffing: oh yes!

configs/setups/awicm3/awicm3.yaml
mandresm: I mean:
```suggestion
                        choose_general.version:
                                "v3.3.0":
                                        version: awicm-3.1
                                "*":
                                        version: v1.1
```
This is to reduce the number of choose_resolution dependencies. At the moment version depends on -> general.major_version -> fesom.version which also depends on -> general.version.

In the way I am saying there is one choose less to be resolved in the backend: general.major_version -> general.version

mandresm: Do not commit this suggestion, I made one below equivalent to this one

configs/setups/awicm3/awicm3.yaml
```diff
@@ -552,8 +681,16 @@ xios:
...
         pool_dir: ${general.pool_dir}
         runoff_maps_name: _${fesom.resolution}
-
-        include_env_from_component_files: False
+        choose_general.major_version:
+                "v3.3":
+                        # TODO: change this to a control on the patch version general.version
+                        choose_fesom.version:
+                                "2.6.8":
+                                        version: awicm-3.1
```
mandresm:
```suggestion
                        choose_general.version:
                                "v3.3.0":
                                        version: awicm-3.1
```

mandresm (changes requested):

configs/setups/awicm3/awicm3.yaml
mandresm: v3.2 works after this change. I am resolving this.

configs/setups/awicm3/awicm3.yaml
```diff
@@ -724,6 +896,12 @@ oasis3mct:
                                 rstos.nc:
                                         - 'A_SST:A_Ice_frac:A_Snow_thickness:A_Ice_temp:A_Ice_albedo <--gauswgt_i-- sst_feom:sie_feom:snt_feom:ist_feom:sia_feom'
 
+        # TODO: change this to a control on the patch version general.version
+        choose_fesom.version:
+                "2.6.8":
```
mandresm:
```suggestion
        # This variable is only used for major versions of v3.3
        choose_general.version:
                "v3.3.0":
```

mandresm:
#bump

mandresm (approved):

JanStreffing:
:tada::tada::tada:

PR #1271, opened 2025-01-20
oasis restart file generation with TCO2559 files

JanStreffing:


src/esm_runscripts/oasis.py
```diff
@@ -140,7 +140,7 @@ def add_coupling(
             if time.lower() in [
                 "stocha",
                 "instant",
-                "accumul",
```
JanStreffing: What happened here.. fixing..

mandresm:
Hi @JanStreffing!

Do you think this would work in all HPCs? Is nc4c supported in all versions of cdo used?

JanStreffing:
Should be. But just to be careful, we can do try except.

JanStreffing:
Ready for merge

src/esm_runscripts/oasis.py
```diff
@@ -665,8 +665,12 @@ def prepare_restarts(self, restart_file, all_fields, models, config):
         # before (i.e. when using LOCTRANS)
         if os.path.isfile(restart_file):
             logger.debug(f"{restart_file} already exits, overwriting")
-        logger.info("cdo -O merge " + filelist + " " + restart_file)
-        os.system("cdo -O merge " + filelist + " " + restart_file)  # + enddate)
+        try:
+            logger.info("cdo -O -f nc4c merge " + filelist + " " + restart_file)
+            os.system("cdo -O -f nc4c merge " + filelist + " " + restart_file)  # + enddate)
```
mandresm: this will not work because `os.system` does not raise an error if the command inside does not work. I am making a fix.

mandresm:
Real tests in Levante are failing. In particular awicm3-v3.2 was not compiling. After merging main I am running the tests again. I could compile it outside of the tests.

JanStreffing:
I'll have a look at it later today.

mandresm:
All tests passing now, except for one AWICM-REcoM, but that's because restart files are missing. So not related to this issue. #1290

It can be merge now then

mandresm:
#bump

mandresm (approved):

PR #1277, opened 2025-01-21
Provenance hints to user_note and user_error and error handling for missing coupling direction

### LINKED ISSUE #1268: Possible naming problem in esm_tools.
The working branch for this error is: feat/awiesm3-3.4_bugfix

In the awiesm.yaml we (with Jan) made the following modifications:

```
lpj_guess:
        ...
        coupling_fields:
                "[[veg_atm_fields-->FIELD]]":
                        grid: ${lpj_guess.mygrid}-land
        grids:
        ...       
               TL159-land:
                        name: TL159-land
                        nx: "${nx}"
                        ny: "${ny}" 
                        oasis_grid_type: "D"
                        oyac_grid_type: "LL"
                        oyac_io_per_node: 4 

```
but we got the following error message after the  esm_runscript. Until we changed the naming it worked.
```

AttributeError: 'NoneType' object has no attribute 'get'
  File "/home/a/a270270/.local/bin/esm_runscripts", line 33, in <module>
    sys.exit(load_entry_point('esm-tools', 'console_scripts', 'esm_runscripts')())
  File "/home/a/a270270/esm_tools/src/esm_runscripts/cli.py", line 289, in main
    setup()
  File "/home/a/a270270/esm_tools/src/esm_runscripts/sim_objects.py", line 119, in __call__
    self.config = prepexp.run_job(self.config)
  File "/home/a/a270270/esm_tools/src/esm_runscripts/prepexp.py", line 27, in run_job
    evaluate(config, "prepexp", "prepexp_recipe")
  File "/home/a/a270270/esm_tools/src/esm_runscripts/helpers.py", line 71, in evaluate
    config = esm_plugin_manager.work_through_recipe(
  File "/home/a/a270270/esm_tools/src/esm_plugin_manager/esm_plugin_manager.py", line 159, in work_through_recipe
    config = getattr(submodule, workitem)(config)
  File "/home/a/a270270/esm_tools/src/esm_runscripts/prepexp.py", line 260, in call_esm_runscripts_from_prepexp
    _call_esm_runscripts_internally(config, new_command, scriptsdir)
  File "/home/a/a270270/esm_tools/src/esm_runscripts/prepexp.py", line 193, in _call_esm_runscripts_internally
    subprocess.check_call(command.split(), cwd=exedir)
  File "/sw/spack-levante/mambaforge-22.9.0-2-Linux-x86_64-kptncg/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['esm_runscripts', 'awiesm3-v3.4-levante-TCO95L91-CORE2_1y.yaml', '-e', 'AWIESM3_Test22', '--open-run', '--no-motd', '--last-jobtype', 'prepcompute', '-t', 'prepcompute']' returned non-zero exit status 1.
```

mandresm:
- Error handling for #1268
- adds `get_first_provenance` method to more easily use the first occurrence of provenance in a dictionary or list to report suggestions in the error messages
- adds hint functionality to `user_error` and `user_note`: if you add a list of hints as input to one of this methods, you can add `<HINT_#>` placeholders to your message for `user_note` and `user_error`. These place holders will be substituted using the hint's properties you define for each hint.

For #1268 it generates this user error now:
Screenshot 2025-01-21 at 19 23 12

Closes #1268

### DIFF
--- a/src/esm_parser/provenance.py
+++ b/src/esm_parser/provenance.py
@@ -523,6 +523,23 @@ class DictWithProvenance(dict):
 
         return provenance_dict
 
+    def get_first_provenance(self):
+        """
+        Recursively loops through the dictionary keys and returns the first provenance
+        found in the nested values.
+
+        Returns
+        -------
+        first_provenance : esm_parser.provenance.Provenance
+            The first provenance found in the nested values
+        """
+        first_provenance = None
+        for key, val in self.items():
+            if isinstance(val, PROVENANCE_MAPPINGS):
+                return val.get_first_provenance()
+            elif hasattr(val, "provenance"):
+                return val.provenance[-1]
+
     def __setitem__(self, key, val):
         """
         Any time an item in a DictWithProvenance is set, extend the old provenance of
@@ -772,6 +789,23 @@ class ListWithProvenance(list):
                 provenance_list.append(None)
 
         return provenance_list
+
+    def get_first_provenance(self):
+        """
+        Recursively loops through the list elements and returns the first provenance
+        found in the nested values.
+
+        Returns
+        -------
+        first_provenance : esm_parser.provenance.Provenance
+            The first provenance found in the nested values
+        """
+        first_provenance = None
+        for elem in self:
+            if isinstance(elem, PROVENANCE_MAPPINGS):
+                return elem.get_first_provenance()
+            elif hasattr(elem, "provenance"):
+                return elem.provenance[-1]
 
     def __setitem__(self, indx, val):
         """
--- a/src/esm_parser/provenance.py
+++ b/src/esm_parser/provenance.py
@@ -523,7 +523,7 @@ class DictWithProvenance(dict):
 
         return provenance_dict
 
-    def get_first_provenance(self):
+    def extract_first_nested_values_provenance(self):
         """
         Recursively loops through the dictionary keys and returns the first provenance
         found in the nested values.
@@ -536,7 +536,7 @@ class DictWithProvenance(dict):
         first_provenance = None
         for key, val in self.items():
             if isinstance(val, PROVENANCE_MAPPINGS):
-                return val.get_first_provenance()
+                return val.extract_first_nested_values_provenance()
             elif hasattr(val, "provenance"):
                 return val.provenance[-1]
 
@@ -790,7 +790,7 @@ class ListWithProvenance(list):
 
         return provenance_list
 
-    def get_first_provenance(self):
+    def extract_first_nested_values_provenance(self):
         """
         Recursively loops through the list elements and returns the first provenance
         found in the nested values.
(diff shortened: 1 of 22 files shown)


mandresm:
This can still be improved, for example, by making an `Hint` object and an `UserNote` object. However, I think improving it would only make sense if we refactor the whole `user_note` thing, move it to it's own module, and including the loguru logger into it. WDYT @pgierz?

pgierz:
Yes, plus 100000000 points. Having those be in `esm_parser` is not where they belong. Suggestion: put it in the `esm_tools` module directly, which is where we should keep things that are needed in all of the sub packages.

```python
from loguru import logger
# Alternatively: from esm_tools.logging import logger (see implementation in pymorize)

class Hint
   def __init__(self, message):
       self.message = message

   def error(self, message=None):
        if message is None:
            message = self.message
        logger.error(message)
   # Same with info, debug, critical, etc
```

...something like that. Maybe with static methods so you don't need to initialise the object.

mandresm:
> Suggestion: put it in the esm_tools module directly, which is where we should keep things that are needed in all of the sub packages.

done!

src/esm_parser/provenance.py
```diff
@@ -523,6 +523,23 @@ def get_provenance(self, index=-1):
...
+        -------
+        first_provenance : esm_parser.provenance.Provenance
+            The first provenance found in the nested values
+        """
+        first_provenance = None
+        for key, val in self.items():
+            if isinstance(val, PROVENANCE_MAPPINGS):
+                return val.get_first_provenance()
+            elif hasattr(val, "provenance"):
+                return val.provenance[-1]
```
pgierz: I would rename this to "oldest" since you getting the last element of the list (the one that was set at the very beginning). First is a bit confusing here, since I would intuitively expect you to get back `val.provenance[0]`

src/esm_parser/provenance.py
```diff
@@ -773,6 +790,23 @@ def get_provenance(self, index=-1):
...
+        first_provenance : esm_parser.provenance.Provenance
+            The first provenance found in the nested values
+        """
+        first_provenance = None
+        for elem in self:
+            if isinstance(elem, PROVENANCE_MAPPINGS):
+                return elem.get_first_provenance()
+            elif hasattr(elem, "provenance"):
+                return elem.provenance[-1]
+
```
pgierz: As above

pgierz (changes requested):
I have one naming suggestion, otherwise this is good to go.

src/esm_parser/provenance.py
mandresm: it gives you the last provenance of the first value it finds in a list or a dictionary, that's why I wrote first. But I'll find a better name

mandresm: a fair point is ChatGPTs answer:

> what does it matter? The provenance you seek is just another step in an endless cycle of function calls and data structures, a fleeting glimpse of order in an otherwise chaotic and indifferent universe.

pgierz (approved):

mandresm:
#bump

mandresm:
#bump

Issue #1279, opened 2025-01-22
LPJ-GUESS Directory Setup in ESM-Tools


hajlaci:
Hi Miguel,

We’ve discovered that LPJ-GUESS requires a unique directory for each processor it runs on, named in the format run<processor_number>. Additionally, these directories must contain copies of all the forcing files. This setup is currently handled during the precompilation stage using a script engine.

We need to reproduce this process within esm_tools and would greatly appreciate your help in getting this done.

Cheers,
Laszlo

mandresm:
Hi @hajlaci,

That is an interesting way of doing things from the LPJ-GUESS side... I have some questions:

- Do these directories need any other files other than copies of the forcings?
- Is the number of processors something that is given during compilation, or the same binary can run with different numbers of processors specified via `lpj_guess.nproc`?

In principle, doing this should not be difficult to do with ESM-Tools. You'll need to use the file dictionaries for the forcing files and expand them with a list loop (https://esm-tools.readthedocs.io/en/latest/yaml.html#list-loops) to generate all the `forcing_in_work` targets for `run<number_of_processors>`, in a similar way you did with the `[[coupling_fields->FIELDS]]`.

There are some examples of how one can do such a thing in the `echam.yaml`:

https://github.com/esm-tools/esm_tools/blob/66e23a5f9670a6781837f134c49d980e5d61c512/configs/components/echam/echam.yaml#L813-L838

Something on these lines (metacode, not really tested):

```yaml
lpj_guess:
    # Make a list of processor numbers
    proc_list: $(( list(range($nproc)) )) # if the first proc number needs to be 0 or $(( list(range(1, $nproc+1)) )) if the first proc number needs to be 1

     forcing_sources: 
         "my_forcing_label_[[proc_list-->PROC_NUM]]": /path/to/forcing/file

    forcing_in_work:
         "my_forcing_label_[[proc_list-->PROC_NUM]]": runPROC_NUM/file
```

Please, let me know if I am understanding the problem correctly and answer the questions I asked at the beginning. If I am understandind the problem correctly, I would recommend that you try this suggestion. If you get stuck, commit and write down here the commit_id and the branch and I'll have a look at it. I am not reachable tomorrow and on Friday and on Monday and Tuesday I'm on a retreat, so responding to this thread is actually the fastest way of getting a response from me (cause I'll be paying attention to GitHub :) )

hajlaci:
Hi @mandresm,

[...]

PR #1281, opened 2025-01-23
add @mandresm contributions to the fix

mandresm:
- Remove keep_list syntax (if we can avoid it, I'd rather do not add more syntax here)
- Makes sure it does the old list handling for date operations, so that date operations are still handled correctly
- Sets the provenance accordingly

This is tested and produces a list such as discussed in #1279

### DIFF
--- a/src/esm_parser/esm_parser.py
+++ b/src/esm_parser/esm_parser.py
@@ -2330,12 +2330,14 @@ def do_math_in_entry(tree, rhs, config):
     else:
         keep_list = False
     entry = " " + str(entry) + " "
+    date_operation = False
     while "$((" in entry:
         math, after_math = entry.split("))", 1)
         math, before_math = math[::-1].split("(($", 1)
         math = math[::-1]
         before_math = before_math[::-1]
         if DATE_MARKER in math:
+            date_operation = True
             all_dates = []
             steps = math.split(" ")
             steps = [step for step in steps if step]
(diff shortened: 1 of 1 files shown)


mandresm:
Hey @pgierz ! This is a PR to your original branch with hopefully a few improvements. Feel free to merge if you like it :)

src/esm_parser/esm_parser.py
```diff
@@ -2413,14 +2410,18 @@ def do_math_in_entry(tree, rhs, config):
                     math = math + "all_dates[" + str(index) + "]"
                     index += 1
         result = eval(math)
-        if isinstance(result, list):
-            if keep_list:
-                return ListWithProvenance(result, [])
+        if isinstance(result, list) and date_operation:
             result = result[
                 -1
             ]  # should be extended in the future - here: if list (= if diff between dates) than result in seconds
```
pgierz, line 2416: Was this only the case for dates? I didn't know that...

src/esm_parser/esm_parser.py
```diff
@@ -2413,14 +2410,18 @@ def do_math_in_entry(tree, rhs, config):
...
+        elif isinstance(result, list):
+            entry = ListWithProvenance(result, None)
+            entry.set_provenance(rhs.provenance)
```
pgierz, line 2419: I might not understand something about provenance 100% yet, but why not this version:
```suggestion
            entry = ListWithProvenance(result, rhs.provenance)
```

pgierz (approved):
Looks good, but I have a question. See below:

src/esm_parser/esm_parser.py
mandresm: For providing provenance values during instanciation of `ListWithProvenance` (or `DictWithProvenance`) the provenance needs to have the same structure as the original list or dict. That is to say, if you have a `dict` with some key-values, what you need to give in the second argument of the instancing is also a `dict` with the same keys, but the values are provenances. `set_provenance` is used to assign the same unique provenance to a whole list/or dict with provenance, which is what we want in this case (we are expanding a string into a list so we want to assign a single provenance information to the new elements of the list.

src/esm_parser/esm_parser.py
mandresm: It was the case for every list, but I think the only way we had of generating lists from a string on the past was via Date operations (because date objects don't contain this functionality so that functionality ended up here, somehow)

PR #1286, opened 2025-01-30
auto Tripyview 1draft

SimonPorr:
Implementation of the necessary files for Auto_Tripy in FESOM 2.6 folder and YAML.

### DIFF
--- a/configs/components/fesom/Subjob_tripy.sh
+++ b/configs/components/fesom/Subjob_tripy.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+#Home directory and reset of variable
+cd ~
+source ~/.bash_profile
+
+cd ~
+
+# Function to check for py39 envirionmet. If necessary installing and activating.
+#this is just to not trigger errors while activating py39
+#module unload netcdf-c
+#module unload netcdf-fortran
+#Function for activating py39
+activate_env() {
+    cd ~
+    source ~/.bash_profile
+    conda activate py39
+}
+
+cd /work/ab0995/a270275/runconfig
+# Check if conda ist installed
+if ! command -v conda &> /dev/null; then
+    echo "Conda is not installed. Plesa install conda or miniconda"
+    exit 1
+fi
+
+# Überprüfen, ob das Environment "py39" existiert
+if conda info --envs | grep -q "^py39 "; then
+    echo "Py39 is exisiting already. Activating now..."
+    activate_env
+else
+    echo "The Py39 env is not exisiting, installing..."
+    cd~
+    source ~/.bash_profile
+    conda create -y --name py39 python=3.9
+    activate_env
+fi
+
+#check if yq library is installed if not installing
+cd ~
+cd /work/ab0995/a270275/runconfig
+if ! command -v yq &> /dev/null; then
+    echo "Installing yq..."
+    pip install yq
+fi
+
+#import necessary variables from main config
+outdata_path=$1
+start_date=$2
+next_date=$3
+initial_date=$4
+final_date=$5
+mesh_dir=$6
+
+#Change dates to years
+start_year=$(date -d "$start_date" +%Y)
+next_year=$(date -d "$next_date" +%Y)
+initial_year=$(date -d "$initial_date" +%Y)
+final_year=$(date -d "$final_date" +%Y)
+#export necessary variables
+export mesh_dir
+export outdata_path
+export start_year
+export next_year
+export initial_year
+export final_year
+
+
+output_file="tripyrun_${start_year}.yaml"
+
+export output_file
+
+#Create folder for Tripy configs and results in the experiment folder
+
+target_dir="$outdata_path/Tripyview/config"
+if [ ! -d "$target_dir" ]; then
+    echo "Directory $target_dir not existing. Creating..."
+    mkdir -p "$target_dir"
+fi
+
+# go to Directory
+# Verwende envsubst, um den Platzhalter in der YAML-Datei zu ersetzen
+envsubst < ESM_auto_tripy.yaml > "$output_file"
+mv "$output_file" "$target_dir"
+#Run the new createt Tripy config
+cd "$target_dir" 
+tripyrun "$output_file"
+
+
+
--- a/configs/components/fesom/Subjob_tripy.sh
+++ b/configs/components/fesom/Subjob_tripy.sh
@@ -1,46 +1,27 @@
 #!/bin/bash
-#Home directory and reset of variable
-cd ~
-source ~/.bash_profile
-
-cd ~
-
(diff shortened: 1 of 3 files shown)


configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+#Home directory and reset of variable
+cd ~
+source ~/.bash_profile
```
pgierz: I would think it is cleaner to have the script not rely on anything set up in the user's `.bash_profile`. What is in your `.bash_profile` that you need to have here?

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+#Home directory and reset of variable
+cd ~
```
pgierz: The script should be runnable from anywhere, so I would recommend that you don't switch directories too much, especially not outside of the current experiment structure unless you have a good reason to.

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+source ~/.bash_profile
+
+cd ~
```
pgierz: As above...

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+#module unload netcdf-c
+#module unload netcdf-fortran
+#Function for activating py39
+activate_env() {
+    cd ~
+    source ~/.bash_profile
+    conda activate py39
+}
+
+cd /work/ab0995/a270275/runconfig
```
pgierz: Again, you should already be inside of your experiment folder when this script runs

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+
+# Function to check for py39 envirionmet. If necessary installing and activating.
+#this is just to not trigger errors while activating py39
+#module unload netcdf-c
+#module unload netcdf-fortran
+#Function for activating py39
+activate_env() {
+    cd ~
+    source ~/.bash_profile
+    conda activate py39
```
pgierz: I think a better name than `py39` can be used, something that is more specific to this job, maybe `esm-tools_post_tripyview` (in general: `esm-tools_[job_part]_[specifier]`)

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+}
+
+cd /work/ab0995/a270275/runconfig
+# Check if conda ist installed
+if ! command -v conda &> /dev/null; then
+    echo "Conda is not installed. Plesa install conda or miniconda"
```
pgierz: Minor typo:

```suggestion
    echo "Conda is not installed. Please install conda or miniconda"
```

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+    exit 1
+fi
+
+# Überprüfen, ob das Environment "py39" existiert
+if conda info --envs | grep -q "^py39 "; then
+    echo "Py39 is exisiting already. Activating now..."
+    activate_env
+else
+    echo "The Py39 env is not exisiting, installing..."
+    cd~
```
pgierz: This will fail, you need:

```suggestion
    cd ~
```

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
pgierz, line 1: I would suggest you turn on error-exit:

```suggestion
#!/bin/bash -e
```

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+if conda info --envs | grep -q "^py39 "; then
+    echo "Py39 is exisiting already. Activating now..."
+    activate_env
+else
+    echo "The Py39 env is not exisiting, installing..."
+    cd~
+    source ~/.bash_profile
+    conda create -y --name py39 python=3.9
+    activate_env
+fi
```
pgierz: Here again, I would suggest a different environment name.

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+
+target_dir="$outdata_path/Tripyview/config"
+if [ ! -d "$target_dir" ]; then
+    echo "Directory $target_dir not existing. Creating..."
+    mkdir -p "$target_dir"
+fi
+
+# go to Directory
+# Verwende envsubst, um den Platzhalter in der YAML-Datei zu ersetzen
+envsubst < ESM_auto_tripy.yaml > "$output_file"
```
pgierz, line 70: What is `envsubst`? Where is it defined or loaded?

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+    activate_env
+fi
+
+#check if yq library is installed if not installing
+cd ~
+cd /work/ab0995/a270275/runconfig
+if ! command -v yq &> /dev/null; then
+    echo "Installing yq..."
+    pip install yq
+fi
```
pgierz: Do you need `yq`? It isn't anywhere in your script. Also, the `pip install` might not work correctly, you likely will need the `--user` flag unless you are sure that you are in a conda environment or a virtual environment. If you are in a conda environment, I would just add this inside the environment specification yaml.

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+#export necessary variables
+export mesh_dir
+export outdata_path
+export start_year
+export next_year
+export initial_year
+export final_year
+
+
+output_file="tripyrun_${start_year}.yaml"
```
pgierz: What happens if the user isn't running in yearly mode?

configs/components/fesom/Subjob_tripy.sh
```diff
@@ -0,0 +1,89 @@
...
+    pip install yq
+fi
+
+#import necessary variables from main config
+outdata_path=$1
+start_date=$2
+next_date=$3
+initial_date=$4
+final_date=$5
+mesh_dir=$6
```
pgierz, line 33: I would add some kind of print statement for these, as a debugging measure.

pgierz (changes requested):
This looks like a solid start :-)

I have some recommendations that might be useful.

PR #1288, opened 2025-02-04
fix/force overwrite in file movements

mandresm:
add the possibility to force overwritting date-stampted files while doing file movements to solve a problem from @christian-stepanek, where tidy has failed due to quota and when ran again esm_runscripts does not want to overwrite the file.

For enforcing this behavior, one will be able to set `general.force_overwrite_in_file_movements": True` in the runscript.

**Please, use this only in cases where you understand very well the problem you are dealing with.**

TODO
- [x] Add option to the documentation
- [x] Check if the file is identical, if files are identical then there is no warning needed or overwrite-enforcing needed.
- [x] Real tests in levante


pgierz (approved):

mandresm:
Thanks for approving @pgierz. Cannot be merged yet, I'd like to finish the TODOs and run some real tests.

christian-stepanek:
> Thanks for approving @pgierz. Cannot be merged yet, I'd like to finish the TODOs and run some real tests.

Thanks a lot for your work with this @pgierz and @mandresm. When one is stuck with this problem, it causes major headaches to get a simulation to run again. I am working now with the Xth try to restart one of my simulations. There is always a different issue / file missing or already available.

mandresm:
@christian-stepanek, I'm just running all the tests in levante just to be sure nothing is broken by this change. They should be done in 2 hours or so, and if nothing is broken I will merge it into `release`. Then you'll only need to update to release and set in your runscript `general.force_overwrite_in_file_movements: True` with a big TODO comment on top, something like: `TODO: remove this line right after a run is successful and forget about this parameter` ;)

mandresm:
#approve-changes

mandresm:
#bump

mandresm:
Merged in `release` @christian-stepanek!

christian-stepanek:
@mandresm @pgierz Thanks a lot for you help with this. I used the switch in the problematic simulation and did not anymore get errors because of pre-existing files.

PR #1291, opened 2025-02-05
Fix/subgrid oro param

ackerlar:
some updates to the orography update in the awiesm-pism coupling


mandresm (commented):
Thanks @ackerlar for this PR. The only problem I see is that the tests fail because you made dEBM a submodule of ESM-Tools and it cannot clone it. I can fix that, by simply telling git not to update that submodule on the tests, or adding a secret to github actions to get access to that repository.

However, I am wondering if it makes sense to have it added as a submodule when all the other models are managed by esm_master. What do you both @pgierz and @ackerlar think about this topic?

ackerlar:
Ah, this was not intended! I fix it after our group meeting

ackerlar:
Hopefully fixed here [https://github.com/esm-tools/esm_tools/pull/1293](https://github.com/esm-tools/esm_tools/pull/1293)

PR #1302, opened 2025-02-17
update to: https://github.com/FESOM/fesom2/releases/tag/2.6.7

JanStreffing:


mandresm (approved):

mandresm:
#bump

ackerlar (commented):
a short test of fesom-2.6.6 and fesom-2.6.7 gives different results! I would therefore rather opt for another awiesm sub-version

JanStreffing:
That would be because the optimization has been turned on. For levante. Did you see a noticable speedup?

ackerlar:
> That would be because the optimization has been turned on. For levante. Did you see a noticable speedup?

No speedup, but I have run one year only

PR #1303, opened 2025-02-17
add awiesm-2.6-1 with fesom-2.6-7

ackerlar:
- keep awiesm-2.6 with fesom-2.6.6
 - add awiesm-2.6.1 with fesom-2.6.7

unsure about awicm3 @JanStreffing


configs/setups/awiesm/awiesm.yaml
```diff
@@ -330,7 +340,17 @@ fesom:
...
-                        version: "2.6-paleodyn"
+                        version: "2.6"
+                        ALE_scheme: 'zstar'
+                        namelist_dir: ${fesom.model_dir}/config/
+                        add_namelist_changes:
+                                namelist.tra:
+                                        tracer_phys:
+                                                surf_relax_s: 0.0
+                        outdata_targets:
+                                fesom: "*.fesom.*.nc"
```
JanStreffing: Is this something version specific? I would like to use that for all versions of fesom2. Users that have sub-yearly runs are rare. IMO those users can then add the alternative output name scheme into the runscript.

JanStreffing (changes requested):

configs/setups/awiesm/awiesm.yaml
JanStreffing: Otherwise this looks good.

JanStreffing (approved):

configs/components/fesom/fesom-2.6.yaml
```diff
@@ -3,7 +3,7 @@
 
 model: fesom
 branch: "2.6"
-version: "2.6"
+version: "2.6-7"
```
mandresm: Why `2.6-7` and not simply `2.6.7`?

configs/components/fesom/fesom-2.6.yaml
```diff
@@ -24,13 +24,11 @@ required_plugins:
...
   "2.6":
-    branch: "2.6.7"
+    branch: "2.6.6"
 
-  "2.6-paleodyn":
+  "2.6-7":
     branch: "2.6.7"
-    wiso_code: true
-    icb_code: true
-    namelist_dir: "${model_dir}/config/"
```
mandresm, line 33: Why does these lines need to be removed?

configs/setups/awiesm/awiesm.yaml
```diff
@@ -93,13 +93,19 @@ general:
         - '2.1-recom'
         - '2.1-paleodyn-2.5'
         - '2.6'
+        - '2.6-1'
```
mandresm: Same as above, why `-1` and not `.1`?

configs/setups/awiesm/awiesm.yaml
```diff
@@ -325,20 +335,29 @@ fesom:
         # --- icebergs ---
         with_icb: ${general.with_icb}
         # --- icebergs ---
+                        
+        outdata_targets:
+                fesom: "*.fesom.*.nc"
```
mandresm, line 340: This will most likely break older versions of this setup. I'll come back to you later on with a suggestion for a solution, if there is really a problem.

mandresm (changes requested):
Just a few comments that need to be addressed before merging

configs/components/fesom/fesom-2.6.yaml
JanStreffing: This should not be part of a fesom version. These are just switches, can be set for any newer fesom version.

configs/setups/awiesm/awiesm.yaml
JanStreffing: afaik the extra . would not work, as it breaks some esm_master logic for what is a version. Did not test this recently though.

configs/setups/awiesm/awiesm.yaml
JanStreffing: How so?

The monthly files break pyfesom2 and tripyview scripts, which are basically the only way to analyse FESOM2 output for most users.

configs/setups/awiesm/awiesm.yaml
ackerlar: For me, `.1` does not work at the moment. The the fesom config is taken from `fesom.yaml` and not `fesom-2.6.yaml`. I will have a closer look later today

configs/setups/awiesm/awiesm.yaml
pgierz: I would argue this should be an impulse to Patrick to make the tripyview tools more flexible. It should not rely on hard-code file patterns...

I see no difficulties in this (rather aggressive) glob, though.

mandresm:
#approve-changes

configs/setups/awiesm/awiesm.yaml
mandresm: > How so?
> 
> The monthly files break pyfesom2 and tripyview scripts, which are basically the only way to analyse FESOM2 output for most users.

Many of our old simulations have a different name convention. That means restarting from them with this new change will not be out of the box, but you'll need to change somethings in the runscript or rename the files. But on a second thought, if we want this change for the future we have to eventually make it happen, so let's go for it I'd say. We will assist people who cannot branch off from old simulations.

configs/setups/awiesm/awiesm.yaml
mandresm: That is indeed strange, let's fix that before we start to patch things in weird ways. @pgierz, can you take care of exploring why 2.6.1 is reading `fesom.yaml` and not `fesom-2.6.yaml`?

mandresm:
This will need another approve-changes after the problem with the path number (2.6.1) is solved

pgierz:
Can't say what right now, but the `2.6.1` problem seems to be very deep. esm-master defaults to the very very old fesom:

```
    Executing commands in this order:
         mkdir -p awiesm-2.6.1
         cd awiesm-2.6.1
         git clone -b 2.8mct-awiesm-2.1 https://gitlab.awi.de/paleodyn/Models/oasis3-mct.git oasis
         git clone -b 6.3.05p2-awiesm-2.1 https://gitlab.awi.de/paleodyn/Models/echam6.git echam-6.3.05p2
         git clone -b master https://gitlab.dkrz.de/modular_esm/fesom-1.4.git fesom-1.4
         sed -i '/set(FESOM_COUPLED/s/OFF/ON/g' fesom-2.6/CMakeLists.txt
         sed -i '/ECHAM6_COUPLED/s/OFF/ON/g' echam-6.3.05p2/config/mh-linux
         sed -ir '/..FC_DEFINE}__cpl_mpiom/s/..FC_DEFINE}__cpl_mpiom//g' echam-6.3.05p2/configure.ac
         cd ..
```

pgierz:
Problem is here:
https://github.com/esm-tools/esm_tools/blob/393c3ea9c079a3076487071b339f51244914eae5/src/esm_environment/esm_environment.py#L368-L376

and here:
https://github.com/esm-tools/esm_tools/blob/393c3ea9c079a3076487071b339f51244914eae5/src/esm_parser/esm_parser.py#L2801-L2806

Basically, the parser expects each version to be in it's own file, it looks there first, or, it looks in one without a version suffix. But defining two versions in one file (which has a version suffix in its name) is not permitted.

Not sure what to do here. Other examples:
* OpenIFS: Just has one model yaml
* FESOM: only has `MAJOR`.`MINOR` (but not `.PATCH`)
* AWICM3 (setup level): Only has one yaml.

ackerlar:
Can we make a decision on this? Maybe merge this PR and agree on some common convention afterwards? Or undo this PR [https://github.com/esm-tools/esm_tools/pull/1302](https://github.com/esm-tools/esm_tools/pull/1302) because it leads to different results for awiesm-2.6 with the default setup...

mandresm:
@ackerlar, can you point me at working runscripts for running fesom-2.6.7 and awiesm-2.6.1 in `levante`? Then I can include them in the tests (at a later date, to not block this urgent PR)

mandresm:
I tested installation of `fesom-2.6.7` and `awiesm-2.6.1` with  the patch version numbers. It works :)

mandresm:
#approve-changes

mandresm (approved):

mandresm:
#bump

PR #1304, opened 2025-02-17
Fix/remove obselet recom cpl

ackerlar:
as I understand, these recom/wiso coupling fields for oasis are not used/not working with the latest awiesm2 version anymore

### DIFF
--- a/configs/setups/awiesm/awiesm.yaml
+++ b/configs/setups/awiesm/awiesm.yaml
@@ -266,15 +266,6 @@ echam:
                                         submodelctl:
                                             lupdate_orog: True
 
-        # with_recom switch
-        # -----------------
-        # Settings for running with REcoM
-        choose_general.with_recom:
-            True:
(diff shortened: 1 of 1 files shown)


a270105 (changes requested):
I think that we still need to keep this part:

`        choose_general.with_recom:
            True:
                # Set fesom.with_recom: true when general.with_recom: true
                with_recom: true`

Since some namelist changes will be done in the fesom yaml file if fesom.with-recom is TRUE.  I guess that the merge by Martin does not affect this.

If the merge was already done, L269-L277 shoul be removed and the wiso_fields in L450, as well. But Miguel should confirm if the merge is completed.

ackerlar:
> I think that we still need to keep this part:
> 
> ` choose_general.with_recom: True: # Set fesom.with_recom: true when general.with_recom: true with_recom: true`
> 
> Since some namelist changes will be done in the fesom yaml file if fesom.with-recom is TRUE. I guess that the merge by Martin does not affect this.
> 
> If the merge was already done, L269-L277 shoul be removed and the wiso_fields in L450, as well. But Miguel should confirm if the merge is completed.

Thanks @a270105 ! I updated the branch. Could you have another look?

a270105 (approved):

ackerlar:
can this be merged?

PR #1300, opened 2025-02-16
AI Rewrite: ESM Calendar

pgierz:
Using the Artificial Intelligence to rewrite some of our more horrible insides. Part 1: The Calendar.

This pull request includes several changes aimed at improving the codebase and adding new functionality. The most important changes involve updating dependencies, refactoring code for better readability, and adding new tests.

Rewrite:
* esm_calendar.py: Completely rewritten by CoPilot.

Dependency updates:
* [`setup.py`](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R42): Added the `deprecated` library to the dependencies.

Code refactoring:
* [`src/esm_runscripts/batch_system.py`](diffhunk://#diff-ea280cdce62a2ebbbc1dda0ac96b5ba1ab1ad052e2b533a5bd0b423afa7dc7e9R7-L9): Moved the import of `loguru` to the top of the file for consistency.
* [`src/esm_runscripts/oasis.py`](diffhunk://#diff-f1543c1113264714a5adf1e6959151c790f77c7685ad5d6b6a361bb2d4eee348L7-R9): Reordered imports to maintain consistency and readability.
* [`src/esm_runscripts/prepare.py`](diffhunk://#diff-7ac61855d05ae3bcc0a7ef95e97a6349a01c070259ed1877f37c238543e89860R8-L13): Moved the import of `loguru` to the top of the file for consistency.
* [`src/esm_runscripts/prev_run.py`](diffhunk://#diff-453f6b70e21425fd87e1a0456488392653e0f61133a6d039cec96db3b6b957abR6-L9): Moved the import of `loguru` to the top of the file for consistency.

Functionality improvements:
[...]

PR #1305, opened 2025-02-18
Ai/esm environment

pgierz:
Similar to #1300, but for esm-environment. This time we set up jinja2 templates for the environment batch files, which makes a lot of the code collapse.

Still very much WIP, so far works for
* FESOM 2.6 on Albedo

The big missing part here is setting up the configuration dictionary to already have compiletime or runtime settings. I feel that this should be done outside from the actual environment generator to cleanly separate the responsibilities, but functions to prepare that could go into esm-environment.


mandresm:
Can this be closed @pgierz?

PR #1306, opened 2025-02-19
Fix/small pism cpl fix

ackerlar:


ackerlar:
incldued in [https://github.com/esm-tools/esm_tools/pull/1327](https://github.com/esm-tools/esm_tools/pull/1327)

PR #1307, opened 2025-02-19
Fix/environment changes and further reading conflicts

mandresm:
Tries to solve #1253 and #1285.

Commits cherry-picked from `awicm3-v3.4/environment_changes_and_further_reading_conflicts`, whose parent is commit `9034da4b85cfe17e32b53d7ef500111f93707d7f` from the branch `feat/awicm3-v3.3_and_v3.4`.

Still a draft.

PR #1311, opened 2025-03-03
Feat/remove environment changes

mandresm:
# Description

Refactor `esm_environmet` and get rid of `environment_changes`. All goes via `computer` now.

# TODO

- [x] Reduce the component/setup `if` in the `ConfigSetup` to one unique case.
- [x] Merge as much as the `deep_update` functions and `dict_merge` as possible
  - [x] Double check that we need `deep_update_further_reading`. Maybe it can be included in the `dict_merge` function or `deep_update`. Comment by @pgierz in #1307
- [x] choose_ conflict detection
- [x] component-level (and further reading) conflict control
- [x] Add new structure to all yaml files
- [x] Doctstrings for changed functions
- [x] Remove all `environment_changes` from the yaml files and transfer them to the `computer` key.
- [x] Will probably destroy the generation of the model list in the documentation: https://esm-tools.readthedocs.io/en/feat-remove_environment_changes/Supported_Models.html fix it in that case
- [x] Rewrite documentation for esm_environment, remove environment_changes from the documentation.
- [x] add deprecation error for `environment_changes`
- [x] add extra functionality for environment vars
- [x] unit testing for extra functionality
    - [x] Environment variables filters
    - [x] Environment variables sorting
    - [x] Further_reading conflict control
    - [x] component-level conflict control
    - [x] chooses
- [x] Cleanup
- [x] remove deprecated methods in esm_environment
- [x] MOTD


src/esm_environment/esm_environment.py
```diff
@@ -525,11 +541,66 @@ def get_shell_commands(self):
...
         return environment
 
+    def select_env_vars_based_on_provenance(self, env_var_key):
+        pass
+
+    def select_env_vars_based_on_var_attributes(self, env_var_key):
+        pass
+
+    def sort_env_vars(self, env_var_key, category_order):
+        import ipdb
```
pgierz: Still needs to be removed:
```suggestion
```

src/esm_environment/esm_environment.py
```diff
@@ -525,11 +541,66 @@ def get_shell_commands(self):
...
 
         return environment
 
+    def select_env_vars_based_on_provenance(self, env_var_key):
+        pass
+
+    def select_env_vars_based_on_var_attributes(self, env_var_key):
+        pass
+
+    def sort_env_vars(self, env_var_key, category_order):
```
pgierz, line 516: I'll try to write docs based on my understanding, please improve or amend as necessary:

```suggestion
    def sort_env_vars(self, env_var_key, category_order):
        """"
        Sorts environment variables based upon their order in the original configuration files, respecting their provenance.
        
        Parameters
        ----------
        env_var_key : str
            The environment variable to be sorted (e.g. ``LD_LIBRARY_PATH``)
        category_order : list
            The category priorities to respect, i.e. ``["computer", "component", "setup"].
            
        Mutates
        -------
        self.config : dict
           The EnvironmentInfo ``config`` dictionary is modified specifically for
           ``env_var_key``, with the resorted, provenance-aware order.
        """
```

src/esm_environment/esm_environment.py
```diff
@@ -525,11 +541,66 @@ def get_shell_commands(self):
...
+        for category in category_order:
+            if category not in ordering_dict:
+                continue
+            ii, jj, keys, values = zip(*sorted(ordering_dict[category], key=lambda x: (x[0], x[1])))
+            if env_vars_type == "dict":
+                for key, value in zip(keys, values):
+                    new_env_vars[key] = value
+            elif env_vars_type == "list":
+                for value in values:
+                    new_env_vars.append(value)
```
pgierz, line 579: I'm not sure I 100% follow this, and I think there might be different behaviour for dicts/lists. In the case of dict, you will potentially override a key (if it already exists), but in the case of dictionary, you would append the value and always make the list longer. So, if a key-thing (like `LD_LIBRARY_PATH`) appears twice, you only get 1 entry in the dictionary, but you would get two things in the list way...?

src/esm_master/compile_info.py
pgierz, line 1: These changes are to support the new "main" headers (e.g. `fesom`) at the top level, right?

src/esm_master/compile_info.py
```diff
@@ -589,6 +594,9 @@ def has_model(self, model):
         return False
 
     def split_raw_target(self, rawtarget, setup_info):
+        #if "oifs-43r3" in rawtarget:
+        #    import ipdb
+        #    ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_master/compile_info.py
```diff
@@ -656,6 +664,8 @@ def setup_or_model(self, model):
         return kind_of_model
 
     def output_available_targets(self, search_keyword):
+        #import ipdb
+        #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -478,15 +478,28 @@ def attach_to_config_and_reduce_keyword(
...
+                # old ways...
+                if "model" in tmp_config:
+                    config_to_write_to[tmp_config["model"]] = tmp_config
+                # new way!
+                else:
+                    # TODO: add an if if computer changes are to be ignored for the standalone files
+                    dict_merge(config_to_write_to, tmp_config)
 
+                #import ipdb
+                #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -478,15 +478,28 @@ def attach_to_config_and_reduce_keyword(
...
 
-                config_to_write_to[tmp_config["model"]] = tmp_config
+                # TODO: remove this if and always do merging after component keys in all yamls
+                # old ways...
+                if "model" in tmp_config:
+                    config_to_write_to[tmp_config["model"]] = tmp_config
+                # new way!
+                else:
+                    # TODO: add an if if computer changes are to be ignored for the standalone files
+                    dict_merge(config_to_write_to, tmp_config)
```
pgierz: Where would the switch for this `if` be? Hopefully not in `dict_merge`, but in the function signature?

src/esm_parser/esm_parser.py
```diff
@@ -478,15 +478,28 @@ def attach_to_config_and_reduce_keyword(
...
+                #ipdb.set_trace()
                 for attachment in CONFIGS_TO_ALWAYS_ATTACH_AND_REMOVE:
                     logger.debug("Attaching: %s", attachment)
-                    attach_to_config_and_remove(
-                        config_to_write_to[tmp_config["model"]],
-                        attachment,
-                        all_config=None,
-                    )
+                    #import ipdb
+                    #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -503,6 +516,12 @@ def attach_single_config(config, path, attach_value, all_config=None, **kwargs):
     all_config : dict
         main configuration
     """
+    import ipdb
+    #if attach_value == "fesom.env.yaml":
+    #    ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -538,12 +581,22 @@ def attach_to_config_and_remove(config, attach_key, all_config=None, **kwargs):
     -------
     The ``config`` is modified **in place**!
     """
+    if attach_key != "further_reading":
+        #import ipdb
+        #ipdb.set_trace()
+        pass
+
+    #import ipdb
+    #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -718,12 +771,21 @@ def deep_update(chapter, entries, config, blackdict={}):
             # Trying to update a dictionary from the same file (in blackdict)
             # an ``add_`` returns an error
             if isinstance(blackdict[chapter], dict):
+                # Keys don't have provenance. In this case we know we are dealing with a
+                # choose block so the line of the conflicting choose is the line of the
+                # first value -3.
```
pgierz, line 747: What about whitespace?

src/esm_parser/esm_parser.py
```diff
@@ -2851,6 +2913,9 @@ def __init__(self, model, version, user_config):
             user_config=user_config,
         )
 
+        #import ipdb
+        #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -2897,28 +2962,36 @@ def _config_init(self, user_config):
...
+        setup_config = copy.deepcopy(general_config)
 
-        setup_config = {
-            "computer": computer_config,
-            "general": general_config,
-        }
+        # Computer config takes precedence over general config (order in dict_merge)
+        dict_merge(setup_config, computer_config)
+        #import ipdb
+        #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -2954,6 +3027,50 @@ def _config_init(self, user_config):
                     current_model=component,
                     user_config=user_config,
                     setup_config=setup_config)
+        # New organization of the standalone files
+        elif setup_name in self.config:
+            #import ipdb
+            #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -2954,6 +3027,50 @@ def _config_init(self, user_config):
...
+            dict_merge(setup_config, self.config)
+
+            # TODO: this block below is absolutely nuts, remove it when cleaning
+            setup_config["general"]["valid_setup_names"] = valid_setup_names = list(
+                setup_config
+            )
+            setup_config["general"]["valid_setup_names"].remove(setup_name)
+            setup_config["general"]["valid_model_names"] = valid_model_names = [
+                setup_name
+            ]
```
pgierz: .....what?

First, take every key in the dictionary setup_config, as a list. Then, throw away the own name (since "awicm3" is actually "general"?) and then, hardwire model names to be just the setup name?

I though `"valid_model_names"` was supposed to be something like `["oifs", "fesom"]`

src/esm_parser/esm_parser.py
```diff
@@ -3018,7 +3135,10 @@ def _config_init(self, user_config):
                     new_model_list.append(model)
             setup_config["general"]["include_models"] = new_model_list
 
+        #import ipdb
+        #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_parser/esm_parser.py
```diff
@@ -3047,17 +3167,19 @@ def _config_init(self, user_config):
...
-                    )
+                        for attachment in CONFIGS_TO_ALWAYS_ATTACH_AND_REMOVE:
+                            attach_to_config_and_remove(
+                                this_config[model], attachment, all_config=this_config
+                            )
 
         # Allows the ``general`` section to be able to handle attachable files (e.g.
         # ``further_reading``)
+        #import ipdb
+        #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_runscripts/prepare.py
```diff
@@ -172,6 +172,8 @@ def model_env_into_computer(config):
     # from esm_parser import basic_choose_blocks, dict_merge, user_note, user_error, pprint_config
 
     # Get which type of changes are to be applied to the environment
+    #import ipdb
+    #ipdb.set_trace()
```
pgierz:
```suggestion
```

src/esm_runscripts/sim_objects.py
```diff
@@ -103,6 +108,8 @@ def __init__(self, command_line_config=None, user_config=None):
         self.config["prev_run"] = prev_run.PrevRunInfo(self.config)
         self.store_prev_objects()
 
+        #import ipdb
+        #ipdb.set_trace()
```
pgierz:
```suggestion
```

pgierz (changes requested):
I've only focused on the Python parts for now, and I still need to again read the changes in the parser to fully get my head around it.

src/esm_tests/tests.py
```diff
@@ -789,9 +789,9 @@ def exist_files(files, path, version):
         exception_list = find_exceptions(f)
 
         # Command's logic
-        if " except " in f and version in exception_list:
+        if " except " in f and str(version) in exception_list:
```
pgierz, line 792: This makes me cry on the inside. How about:

```suggestion
        version_in_exception_list = str(version) in exception_list
        version_not_in_exception_list = not version_in_exception_list
        f_contains_except = " except " in f
        f_contains_in = " in " in f
        if f_contains_except and version_in_exception_list:
            continue
        elif f_contains_in and version_not_in_exception_list:
            continue
            
```

src/esm_tests/tests.py
```diff
@@ -520,7 +520,7 @@ def check(info, mode, model, version, out, script, v):
     """
     # Set variables
     success = True
-    mode_name = {"comp": "compilation", "submission": "submission", "run": "runtime"}
+    mode_name = {"comp": "compilation", "submission": "submission", "run": "run"}
```
pgierz, line 523: Would it be an extraordinary amount of work to just keep all of these things identical? Just `"comp"` --> `"compilation"` (on the key side) or `"compilation"` --> `"comp"` (on the value side)

src/esm_parser/provenance.py
```diff
@@ -41,17 +43,19 @@
 from ruamel.yaml import YAML
 
 import esm_parser
-import esm_tools
 from esm_calendar import Date
+from esm_tools import user_error
 
 CATEGORY_HIERARCHY = [
+    None,
```
pgierz, line 50: What does `None` mean in this circumstance?

src/esm_master/esm_master.py
```diff
@@ -34,7 +34,8 @@ def main_flow(parsed_args, target):
 
     user_config = write_minimal_user_config(setups2models.config)
     # Miguel: Move this somewhere else after talking to Paul and Dirk
-    user_config["general"]["run_or_compile"] = "compiletime"
+    user_config["computer"] = user_config.get("computer", {})
+    user_config["general"]["execution_mode"] = user_config["computer"]["execution_mode"] = "compile"
```
pgierz: why is this kept in both dictionaries? I am afraid this might be fragile, and lead to unintended mismatches:

```python
>>> my_dict_a = {"account": "computing.computing", "thing_a": "setting_a"}
>>> my_dict_b = {"name": "albedo", "thing_b": "setting_b"}
>>> my_dict_a["execution_mode"] = my_dict_b["execution_mode"] = "compile"
>>> my_dict_a
{'account': 'computing.computing', 'thing_a': 'setting_a', 'execution_mode': 'compile'}
>>> my_dict_b["execution_mode"] = "run"
>>> my_dict_b
{'name': 'albedo', 'thing_b': 'setting_b', 'execution_mode': 'run'}
>>> my_dict_a
{'account': 'computing.computing', 'thing_a': 'setting_a', 'execution_mode': 'compile'}
```

src/esm_environment/esm_environment.py
```diff
@@ -37,53 +38,55 @@ class EnvironmentInfos:
...
     ----------
-    run_or_compile : str
+    execution_mode : str
         A string indicating whether ``EnvironmentInfos`` was instanced from a
-        compilation operation (``compiletime``) or a run (``runtime``).
-    complete_config : dict
+        compilation operation (``compile``) or a run (``run``).
+    config : dict
         Dictionary containing all the compiled information from the `yaml` files
         needed for the current `ESM-Tools` operation.
```
pgierz: I think the parameter order changed, so this needs to be reflected in the documentation as well.

pgierz (approved):
OK, I have read and understood most of this. The changes to esm-environment will need to be read a second time to make sure I understand everything, but, so this is no longer blocking, it can be merged (subject to some small cleanup which you already have on your list)

src/esm_environment/esm_environment.py
mandresm: You can have `LD_LIBRARY_PATH` multiple times if you use these pseudo-indexes:

```yaml
export_vars:
    A_VAR_TO_BE_EXPORTED: the_value
    A_VAR_TO_BE_EXPORTED[(1)]: $A_VAR_TO_BE_EXPORTED:another_value
```

and as explained in the documentation (https://esm-tools.readthedocs.io/en/latest/esm_environment.html#environment-variables)

"""
As a dictionary, export_vars is not allowed to have repeated keys. This could be a problem when environments are required to redefine a variable at different points of the script. To overcome this limitation, repetitions of the same variable are allowed if the key is followed by an integer contained inside [(int)]:
"""

These feature has been available to `export_vars` for a few years now and it seems to have worked so far pretty well. The idea behind "not appending" into a dictionary is that some times you really will want to change the value of a given key, while for a list this is much less likely (with keys you typically either want to overwrite the entire list of append to it with `add_`)

src/esm_parser/esm_parser.py
mandresm: This is already handled somewhere else. The switch is handled in the esm_environment.py with a var called `include_env_from_component_files`, which is by default set to true

src/esm_parser/esm_parser.py
mandresm: That is a very good point, however, I could not figure out how to do this without implementing provenance on keys. Feel free to open an issue for that, we can solve it at another time.

src/esm_parser/esm_parser.py
mandresm: Yeah I don't get those lines either, but changing the logic seems to create a lot of problems, so I leave the logic intact. We can address this when we refactor the parser (one day...)

mandresm:
Do not merge this branch yet. I'm almost done with resolving all the conflicts. Then I will run some real esm_tests again. I'll merge it today or tomorrow.

src/esm_parser/provenance.py
mandresm: If `category` does not exist but provenance exist, a get command in the provenance will return a `None`.

src/esm_master/esm_master.py
mandresm: sorted

mandresm:
#approve-changes

mandresm:
#approve-changes

mandresm:
#bump

PR #1321, opened 2025-03-10
fix: jsbach pool_dir_couple

pgierz:
Fix for missing pool_dir_couple


pgierz:
OK, now we are back to only Albedo, Blogin, and Ollie broken, sorry about that.

PR #1322, opened 2025-03-10
CI: Removes Broken Tests

pgierz:
For now, esm-tests is broken for:
* blogin
* albedo
* ollie

To prevent errors when merging, these tests are disabled for now.

### DIFF
--- a/.github/workflows/esm_tests_computer_workflow.yml
+++ b/.github/workflows/esm_tests_computer_workflow.yml
@@ -39,10 +39,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
       with:
-        submodules: true
+        submodules: recursive
         fetch-depth: 0
-        token: ${{ secrets.GITHUB_TOKEN }}
-        ref: ${{ github.event.pull_request.head.ref }}
+        # token: ${{ secrets.GITHUB_TOKEN }}
+        # ref: ${{ github.event.pull_request.head.ref }}
     - name: install esm-tools
       run: |
         source /opt/conda/etc/profile.d/conda.sh;
--- a/.github/workflows/esm_tests.yml
+++ b/.github/workflows/esm_tests.yml
@@ -7,10 +7,10 @@ name: esm_tests
 on: [pull_request]
 
 jobs:
-  ollie:
-    uses: ./.github/workflows/esm_tests_computer_workflow.yml
-    with:
-        computer: ollie1
+  # ollie:
+  #   uses: ./.github/workflows/esm_tests_computer_workflow.yml
+  #   with:
+  #       computer: ollie1
   levante:
     uses: ./.github/workflows/esm_tests_computer_workflow.yml
     with:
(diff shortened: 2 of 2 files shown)


mandresm:
I added Ollie back. It has some scenarios that are not in other HPCs that need to be compared against in the dry runs, at least until we get the equivalent ones for Albedo. I also removed vilma from the tests in ollie because it was failing. Vilma was never really generallized enough for production (hardcoded paths in Ollie) so it makes sense to not tested as I don't see anyone using vilma through ESM-Tools any time soon and we will need to rework it a bit.

pgierz:
The Vilma/Pism runscript needs an account, I've updated the submodule for resources accordingly, let's see how far the tests get now...

pgierz:
No version bump, as this changes the test infrastructure itself...

.github/workflows/esm_tests_computer_workflow.yml
```diff
@@ -39,10 +39,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
       with:
-        submodules: true
+        submodules: recursive
         fetch-depth: 0
-        token: ${{ secrets.GITHUB_TOKEN }}
-        ref: ${{ github.event.pull_request.head.ref }}
+        # token: ${{ secrets.GITHUB_TOKEN }}
+        # ref: ${{ github.event.pull_request.head.ref }}
```
mandresm, line 45: Hi @pgierz, what is the reason of removing this? I think removing the `ref` will break the CI. That line as far as I remember was taking care of checking out the branch of the PR

pgierz: Token was already set to the default value, and the ref setting by default:

> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow.

...which I think is what we wanted and I couldn't get my head around what the setting was really pointing to.

mandresm: The `ref` is what I am worried about in fact. That basically says "checkout the the head of the branch of the PR". I added that line because for some reason otherwise it wouldn't use the correct branch and commit. It was a pain to configure that for esm_tests. I don't understand why without it, it wouldn't work as the obvious default behaviour would be to checkout the current PR branch. Why you still got problems with that line in, it's a mistery to me as well. I've experience similar problems on the past when I touch the `.github` files, but they vanish when merge on `release`. I am a bit nervous of that been gone, it might mean that the `base` branch is always checkout which will mean that esm_tests will always give you a false positive.

Can you check that this line been gone has the right behaviour? For example, modifying the workflow to print the commit sha and modifying some of the yamls (echam for example) to change a namelist parameter and then creating a test PR? This should flag all echam namelists as changed and set esm_tests CI as failed for levante and ollie. You could also check that the commit id is the expected one.

Sorry that this requires more extra work. I'm really not comfortable risking just one bit in `esm_tests`, they've done too much good to us.

pgierz: Draft in #1326 

I'll change something in echam and we can see

mandresm: This ended breaking approve-changes. I'll be reverting it back on release

PR #1323, opened 2025-03-10
Fix/include debm install script

ackerlar:
two changes:
 - include a script for debm installation (not debm itself) used by the coupling scripts during the atmosphere2pism coupling step
 - use oromea instead of geosp for downscaling to keep smaller scale features

### DIFF
--- a/couplings/dEBM/install_debm.sh
+++ b/couplings/dEBM/install_debm.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+MACHINE=$1
+
+if [[ ${MACHINE} == "albedo" ]]; then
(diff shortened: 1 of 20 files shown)


pgierz:
I'm going to re-merge release back into this one, we have working tests again which we can trust 🙂. If that works out, I'll put this in.

pgierz (approved):

pgierz:
#bump

ackerlar:
anything to do from my side?

mandresm:
This is a parallel way of installing models way from `esm_master`. It would be ideal to have debm installed by `esm_master` in the future. We cannot support users using this setup until then.

ackerlar:
But this feature has been there anyway

PR #1324, opened 2025-03-10
Fix/pism cpl prec downscale

ackerlar:
turn off precipitation downscaling per default in awiesm pism coupling


ackerlar:
included in [https://github.com/esm-tools/esm_tools/pull/1327](https://github.com/esm-tools/esm_tools/pull/1327)

PR #1330, opened 2025-03-17
Fix/dual hemisphere setup

ackerlar:


configs/components/fesom/fesom-2.6.yaml
```diff
@@ -198,12 +198,6 @@ choose_icb_code:
         True:
             use_icebergs: "${with_icb}"
             use_icesheet_coupling: "${with_icb}"
-            add_namelist_changes:
-                namelist.config:
-                    icebergs:
-                        use_icebergs: "${use_icebergs}"
-                        use_icesheet_coupling: "${use_icesheet_coupling}"
-                        steps_per_ib_step: 8
```
mandresm, line 206: What's the reason for movin to this to the `choose_with_icb` and out of the `choose_icb_code`? I guess this has to do with the fact that FESOM is probably always compiled now with icb, correct? In that case, the `choose_icb_code` does not make any sense anymore and we can remove it, and move this whole block directly to the `fesom._namelist_changes` section, so something like this:

```yaml
use_icebergs: "${with_icb}"
use_icesheet_coupling: "${with_icb}"
namelist_changes:
    manelist.config:
            ...
            icebergs:
                use_icebergs: "${use_icebergs}"
                use_icesheet_coupling: "${use_icesheet_coupling}"
                steps_per_ib_step: 8
```

You can also move `use_icebergs` and `use_icesheet_coupling` out of the `choose` block.

If you follow this suggestion you'll always be modifying this part of the namelist which will always allow you to have control of that part of the namelist independently of the value of `with_icb`. The other option is that you want to only modify the namelist if `with_icb: True`. Then, moving this block to the choose below, as you did, makes sense, but you should also move `use_icebergs` and `use_icesheet_coupling` and remove the `choose_icb_code`

configs/components/fesom/fesom-2.6.yaml
```diff
@@ -302,6 +296,13 @@ choose_with_icb:
                     scaling: "scaling"
                     felem: "felem"
                     #num_non_melted_icb_file: "num_non_melted_icb_file"
+            
+            add_namelist_changes:
+                namelist.config:
+                    icebergs:
+                        use_icebergs: "${use_icebergs}"
+                        use_icesheet_coupling: "${use_icesheet_coupling}"
+                        steps_per_ib_step: 8
```
mandresm: See comment above.

mandresm (changes requested):
Thanks Lars, only one minor comment.

mandresm (approved):
All good from my side. Is this ready to merge from your side @ackerlar ?

ackerlar:
@mandresm thank you! Yes, I am ready to merge

mandresm:
#approve-changes

mandresm:
#bump

PR #1331, opened 2025-03-20
Random Expids

pgierz:
Tired of inventing names? Do you find `test` a meaningless waste of your time? What were you testing? Was it something sensible?

Welcome to the random era! Now your experiments will be honored with randomness! Life is unpredictable. Now, your experiment names can be, too.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ requirements = [
     "pytest==7.1.2",
     "pyyaml==6.0.1",
     "questionary==1.10.0",
-    "randomnames==0.2.1",
+    "randomname==0.2.1",
     "ruamel.yaml==0.17.32",
     "semver==2.13.0",
     "sqlalchemy>=1.4.39",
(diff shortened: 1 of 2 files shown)


mandresm (changes requested):
Is there a mention in the readthedocs about the `test` default? If so, can you write something about it?

pgierz:
I can, yes. 

There is also a bit of a problem with this that I still need to fix: if you don't provide anything at all, it used to take the default of `"test"`. Now, it generates a random name (which I find enjoyably silly), but, during the internal resubmit process, it then picks a random DIFFERENT name, and starts going in circles, never actually submitting. For this to work sensibly, it needs to ensure the same name is used during the resubmit...

mandresm:
Also if I as a user, I launch a simulation with a command and then I expect that with the same command I can continue the simulation again, things will get confusing, because for continuing a simulation I would need to specify with `-e` the name of the experiment.

mandresm:
I am closing this PR. It will break too many things and it is not needed.

mandresm:
Also, if we add a feature that is not needed, but adds confusion or bugs, the users won't be too happy. Sorry @pgierz

Issue #1332, opened 2025-03-25
error: File movement conflict


chrisdane:
Hi

Within an experiment we have to rerun a year that was already modeled. We did this by simply setting the `scripts/<expid>_awicm.date` to the desired start date (2015-01-01 in this example, i.e. year 2015 was already modeled).

In the tidy phase of this 2nd run of year 2015 we get this error:
```
====================================================================================
::: Executing the step:  copy_stuff_back_from_work    (step [6/8] of the job:  tidy)
====================================================================================

::: Copying files
2025-03-25 12:50:40.359470
::: copying file category: atmout
- source: /work/ab1095/a270211/out/awicm-1.0-recom/sofia/esm-hist_9211_y70/run_20150101-20151231/work/atmout
- target: /work/ab1095/a270211/out/awicm-1.0-recom/sofia/esm-hist_9211_y70/log/echam/atmout
2025-03-25 12:50:40.362382

ESC[31mERROR: File movement conflict
-----------------------------ESC[0m
The file ESC[31m/work/ab1095/a270211/out/awicm-1.0-recom/sofia/esm-hist_9211_y70/log/echam/atmout_20150101-20151231ESC[0m already exists. Skipping movement:
/work/ab1095/a270211/out/awicm-1.0-recom/sofia/esm-hist_9211_y70/run_20150101-20151231/work/atmout -> /work/ab1095/a270211/out/awicm-1.0-recom/sofia/esm-hist_9211_y70/log/echam/atmout_20150101-20151231
```

Whats the correct thing to do if we want to repeat a run (here a year) within an experiment? I couldn't find this case in the cookbook on the docu page.

(i had the impression that this worked in the past but probably I remember wrongly ...)

runscript in experiment dir:
```
/work/ab1095/a270211/out/awicm-1.0-recom/sofia/esm-hist_9211_y70/scripts/awicm_recom_esm-fw_levante_hist.run.yaml
```

Thanks a lot!

**System (please complete the following information):**
 - Supercomputer: levante
 - Version 6.43.1

mandresm:
Hi @chrisdane, see here https://esm-tools.readthedocs.io/en/latest/esm_variables.html#runtime-variables the parameter `force_overwrite_in_file_movements`. That variable is set to `False` by default to prevent that ESM-Tools accidentally overwrites one file. But for what you are trying to do you can give a go to:

```yaml
general:
    force_overwrite_in_file_movements: True
```

chrisdane:
[...]

PR #1334, opened 2025-03-28
Update filelists.py

pgierz:
Fix for #1332


src/esm_runscripts/filelists.py
```diff
@@ -1276,19 +1276,19 @@ def avoid_overwriting(config, source, target):
...
                 )
-            elif config["general"]["force_overwrite_in_file_movements"]:
+                return date_stamped_target
+            if config["general"]["force_overwrite_in_file_movements"]:
                 os.remove(date_stamped_target)
+                # Change to user_note, the default was user_error (above)
                 warning_function = user_note
-            else:
-                # This will exit(1) (default in configs/defaults/general.yaml)
-                warning_function = user_error
```
mandresm: I think the behavior should be:
1. If force_overwrite -> the user specified to overwrite so we only throw a user_error
2. If no force_overwrite -> unless the user is rerunning some simulation, this is a really strange scenario that can lead to data loss and consequently we want to through an error.

mandresm (changes requested):

src/esm_runscripts/filelists.py
pgierz: > If force_overwrite -> the user specified to overwrite so we only throw a user_error

You mean `user_note`, right?

> If no force_overwrite -> unless the user is rerunning some simulation, this is a really strange scenario that can lead to data loss and consequently we want to through an error.

The current implementation does this, I think...? Further up, the default `warning_function` is `user_error`. So, it is "unless the user says to allow overwrite, throw an error"

mandresm: no, I mean `user_error`

mandresm: I mean that these lines should be changed to:

```python
            if config["general"]["force_overwrite_in_file_movements"]:
                os.remove(date_stamped_target)
                # Change to user_note, the default was user_error (above)
                warning_function = user_note
            else:
                # This will exit(1) (default in configs/defaults/general.yaml)
                warning_function = user_error
```

mandresm: I don't see any flaw with the original logic except for the missing user_note in the import. What is exactly the problem/error you are trying to resolve here?

mandresm (approved):

src/esm_runscripts/filelists.py
mandresm: I revert it most of it as it was, but now we don't get the user_note missing error.

mandresm: Applied the changes suggested by me

mandresm:
#bump

PR #1345, opened 2025-05-02
fix/math in list multikeys

### LINKED ISSUE #1343: Copying LPJ-GUESS restart files
I'm running the coupled AWIESM3 model with LPJ-GUESS, and I'm having trouble getting the LPJ-GUESS restart mechanism to work across years.

We've completed a two-year simulation (with N deposition turned off for now), but I can't properly copy the LPJ-GUESS state files to the correct run directories so that the model can continue from the restart.

I tried the following approach in the `lpj_guess` section of `awiesm3.yaml:`

```
nproc: 128
proc_list_entry: '[str(x) for x in range(1,${nproc}+1)]' 
proc_list: $(( ${proc_list_entry} ))
```

And for the restart handling:

```
       restart_files:
                "state_[[proc_list-->PROC_NUM]]": "[[proc_list-->PROC_NUM - 1]].state"

        restart_in_work:
                "state_[[proc_list-->PROC_NUM]]": "runPROC_NUM/guess.restart"

        restart_sources:
                "state_[[proc_list-->PROC_NUM]]": "work/lpjg_state_{{ end_date.syear + 1 }}/[[proc_list-->PROC_NUM - 1]].state"

        file_movements:
                lpjg_restart:
                        all_directions: copy
```

Although esm-tools doesn’t crash, it also doesn’t copy the restart files at all, so LPJ-GUESS starts without the proper initial state, defeating the restart mechanism.

I suspect the problem is related to how esm-tools processes these template expressions, especially the use of `PROC_NUM - 1` in file names or paths.

Any suggestions for a working configuration would be greatly appreciated.

mandresm:
Allows to include a math operation in multikeys using the recursion variable, for example:

```yaml
restart_in_in_work:
    state_[[list_proc-->PROC]]: run$(( PROC+1 ))/PROC.state
```

Closes #1343

**TODO**

- [x] Unit tests


mandresm:
Not only this works for the unit tests but also adding this to any component (I tried with FESOM):

```yaml
fesom:
    proc_list_entry: '[str(x) for x in range(0,${nproc})]'
    proc_list: $(( ${proc_list_entry} ))
    add_restart_in_in_work:
        state_[[proc_list-->PROC]]: run$(( PROC+1 ))/PROC.state
```

@pgierz is on holidays, so let's go YOLO and merge without review

mandresm:
#bump

PR #1347, opened 2025-05-07
remove empty additional_flags in slurm.yaml to avoid them overriding …

mandresm:
…the ones in the machine files


mandresm:
#approve-changes

PR #1348, opened 2025-05-07
fix/resolve nested chooses in resolve_chooses_with_var

mandresm:
change the resolve_choose_with_var function to be able to handle nested chooses


mandresm:
#bump

PR #1365, opened 2025-06-13
fix/map_constructor for duplicated keys checker

mandresm:
adds file,line,col to the error that reports duplicated keys within a single file


mandresm:
It adds then last two lines to the error message indicating the location of the conflicting keys:

Screenshot 2025-06-13 at 12 45 05

mandresm:
#bump

PR #1370, opened 2025-07-02
extracts coupling as git submodule for easier control

pgierz:


mandresm:
Hi @pgierz,

I discussed this with @ackerlar once and I said it would be better to treat this as a standard component of ESM-Tools (like a model). Having a submodule of this does not look consistent with the way we are doing things currently. IMO if we do a submodule with this why aren't we doing submodules for FESOM, Oasis etc?

Let's discuss which option is better before we make a decision. Here my thoughts about it:

**Advantages of submodule**

- One can keep the exact same file structure

**Disadvantages**

- ESM-Tools version is linked to the coupling scripts version. That means that for a given AWIESM version one cannot use a different version of the coupling scripts, unless you manually check it out in the module

What do you guys think? @ackerlar, @pgierz

mandresm:
Is it possible to export the git history of this when moving it to a different repository? In this way @ackerlar can also get credit for his changes.

If it is not possible, the first commit on that repo should have him as co-author of that commit.

PR #1374, opened 2025-07-09
iterative coupling can be used with PALEO and PI-CTRL scenario

ackerlar:
Fixes this hierachy conflict in `jsbach.yaml` when using `iterative_coupling` with scenarios


mandresm:
Hi @ackerlar, for better clarity and simplicity of the echam.yaml, can you move this lines to awiesm.yaml, into the jsbach section? That's where they belong, unless you are doing echam-standalone with pism.

PR #1376, opened 2025-07-14
Fixes itercoup paleo

YugengChen:
Fixed an issue where:

Running the AWI-ESM2.6 coupled with PISM model previously failed due to an outdated xarray version.

After updating xarray, a subsequent error occurred where the esm_tools path could not be found.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ requirements = [
     "jinja2==3.1.4",
     "loguru==0.6.0",
     "numpy>=1.19.5",  # Maximum version for Python 3.6 support
-    "packaging==21.3",
+    "packaging>=21.3",
     "pandas>=1.1.5",  # Correct compatiability with xarray for Python 3.6
     "psutil==5.9.1",
     "pytest==7.1.2",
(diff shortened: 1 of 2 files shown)


mandresm:
Hi @YugengChen, which version of ESM-Tools are you using?

mandresm:
Are you running esm_tools from a conda environment? I am asking because I think I am also seeing the same problem as you with the path of esm_tools not been found, but only when I use esm_tools from a conda environment.

YugengChen:
> Are you running esm_tools from a conda environment? I am asking because I think I am also seeing the same problem as you with the path of esm_tools not been found, but only when I use esm_tools from a conda environment.

Hello @mandresm, I am using version 6.51.0 and I am not using conda. The above code has resolved the error that occurred when submitting the job. However, after running AWI-ESM, an error arises again when generating the forcing files for PISM, and I have not yet resolved it. This issue only occurs when coupling AWI-ESM and PISM on Albedo. If you have resolved this issue, please let me know.

mandresm:
Hi @YugengChen,

Sorry for the delay on the reply. Somehow this has slipped away from our radar at the ESM-Tools team.

I guess this is still a problem correct? Do you have an experiment that reproduces this behavior? If so, can you share a path to the experiment?

YugengChen:
Hello @mandresm, sorry for the late reply, I am on vacation in China right now. Yes, this is still a problem, I have a directory which is: /albedo/work/projects/p_Tidal_Pliocene/plio_out/t1928_wind

Issue #1392, opened 2025-10-23
compilation flag -fPIC for oasis and fesom


a270105:
**Describe the problem you are facing**
I can't compile awiesm-2.6 with recom using my esm-tools branch 'awiesm-2.6-recom'. The error occurs when linking shared libraries during fesom compilation, e.g.:
```
ld: /work/ab0995/a270105/esm_code/awiesm-2.6-recom/oasis/build/lib/psmile/libpsmile.a(mod_oasis_getput_interface.F90.o): relocation R_X86_64_32 against `__STRLITPACK_135' can not be used when making a shared object; recompile with -fPIC
```

**To Reproduce**
```
cd esm_tools
git checkout awiesm-2.6-recom
esm_master install-awiesm-2.6-recom
```

**Dirty solution**
'-fPIC' is included in oasis3mct.env.yaml. So I don't really understand why this needs to be added.
I tried to add the flag in CMake files. But it didn't work. At the end I added cmake option directly in comp-fesom... and comp-oasis...shell scripts and recompiled all model components by running shell scripts.

**New problems**
After successfully compiling the model, I got errors by running experiment:
```
Traceback (most recent call last):
  File "/home/a/a270105/.local/bin/esm_runscripts", line 33, in <module>
    sys.exit(load_entry_point('esm-tools', 'console_scripts', 'esm_runscripts')())
  File "/home/a/a270105/esm_tools/src/esm_runscripts/cli.py", line 289, in main
    setup()
  File "/home/a/a270105/esm_tools/src/esm_runscripts/sim_objects.py", line 139, in __call__
    self.prepcompute()
  File "/home/a/a270105/esm_tools/src/esm_runscripts/sim_objects.py", line 230, in prepcompute
    self.config = prepcompute.run_job(self.config)
  File "/home/a/a270105/esm_tools/src/esm_runscripts/prepcompute.py", line 26, in run_job
    config = evaluate(config, "prepcompute", "prepcompute_recipe")
  File "/home/a/a270105/esm_tools/src/esm_runscripts/helpers.py", line 71, in evaluate
    config = esm_plugin_manager.work_through_recipe(
  File "/home/a/a270105/esm_tools/src/esm_plugin_manager/esm_plugin_manager.py", line 159, in work_through_recipe
    config = getattr(submodule, workitem)(config)
  File "/home/a/a270105/esm_tools/src/esm_runscripts/filelists.py", line 1395, in report_missing_files
[...]

PR #1393, opened 2025-10-27
fix/type checking in _check_fesom_missing_files

mandresm:
add error if the path in _check_fesom_missing_files cannot be interpreted as a file

Solves issue 2 in #1392


pgierz (approved):

mandresm:
#bump

PR #1399, opened 2025-11-17
Fesom 2.7.0 standalone

JanStreffing:


mandresm:
Looks good to me but for some reason, CI tests are failing. I'm looking into that. Also including fesom2.7 into the `esm_tests`

mandresm:
FESOM is failing to compile in the first test I ran. See `/work/ab0246/a270152/testing/comp/fesom/fesom-2.7/comp.out`

JanStreffing:
I thought it was a good idea to turn OMP on by default. Seems thats not the case.

Two solutions:
a) turn omp off in the fesom 2.7 in the comp_command.
b) make a change in the cmakelist.txt in fesom to prevent this error:

```
if(${ENABLE_OPENMP})
   target_link_libraries(${PROJECT_NAME}.x PRIVATE OpenMP::OpenMP_Fortran)
endif()
```

I suggest we do both. a) for a quick fix now, b) for 2.7.1.

mandresm:
The errors during compilation still persist after having set `-DOPENMP=OFF`. See agian `/work/ab0246/a270152/testing/comp/fesom/fesom-2.7/comp.out`

JanStreffing:
Sorry for the long wait. Can you do the test again? It seems to have been cleaned up.

mandresm:
Sorry for the delay @JanStreffing, I have reproduced the error again when I turned `-DOPENMP=OFF`. The log file is in the same path as before: `/work/ab0246/a270152/testing/comp/fesom/fesom-2.7/comp.out`

JanStreffing:
does the same error happen when you copy the comp script from
/work/ab0246/a270092/model_codes/comp-fesom-2.7_script.sh
?

pgierz:
I made a small correction to the comp command (stray comma) and found two mesh directories that were pointing to ollie (@patrickscholz's folder) and one in the home of @christian-stepanek. I've made a small note about both, those should migrate to a shared pool eventually.

I can have a look at running esm-tests for this later this week.

JanStreffing:
We need to get back to this. The release should get 2.7. Any suggestions when to meet?

mandresm:
This is related to #1464.

I can meet on Thursday 28th of May (tomorrow)

pgierz:
Status check: do we still need this? Is is superseded?

JanStreffing:
do we have fesom 2.7 standalone on release branch of esm_tools?

pgierz:
superseded

PR #1404, opened 2026-01-05
fixing perl on levante

ufukozkan:
fixing perl library


mandresm:
Hi @ufukozkan,

Thanks for your contribution. However, can you explain why this change was needed? Which problem were you trying to solve? I have just run tests on Levante and I have not seen any problem with perl, that's why I am curious, so that I can better understand the problem.

Issue #1380, opened 2025-08-18
Evaluation of parallel_file_movements: 'threads' with AWI-ESM3 on levante@dkrz.de


JanStreffing:
I ran two simulations AWI-ESM3 at the same time (same HPC and filesystem load) on levante. One with:

```
general:
    parallel_file_movements: 'threads'
```

one without this option.

My findings over the first 4 years / simulation legs are:

### Sequential
| Run | Compute Time | Total Time | Cleanup Time |
|-----|-------------|------------|--------------|
| 1   | 19m40s       | 26m09s     | 6m29s        |
| 2   | 19m31s       | 27m41s     | 8m10s        |
| 3   | 21m02s       | 26m53s     | 5m51s        |
| 4   | 19m34s       | 25m00s.    |  5m26s      |

### Threaded
| Run | Compute Time | Total Time | Cleanup Time |
|-----|-------------|------------|--------------|
| 1   | 20m28s       | 24m45s     | 4m17s        |
| 2   | 20m56s       | 27m38s     | 6m42s        |
| 3   | 19m37s       | 22m58s     | 3m21s        |
| 4 | 19m31.s | 23m34s | 4m03s |

Three observations: 

1. The spread in cleanup time is quite large, likely due to file system load issues. 
2. At equal load I seem to see a speedup. 
3. That speedup may be in the range 1.3x - 1.5x. 

I'll add more years to see if the differences hold. @mandresm 

Experiment folders:
/work/bb1469/a270092/runtime/awiesm3-develop/run_dynamic_allocation_01/
/work/bb1469/a270092/runtime/awiesm3-develop/test_par_file_threads_01/

mandresm:
Thanks for reporting @JanStreffing! If you provide me with the path to work of one of this simulations I can test how long does it take with my test dask script to copy those files somewhere else (a good estimation of what it'll take for ESM-Tools in the future once dask does the job). Then we can prioratize it's implementation if the speed up is satisfactory.

mandresm:
oh! I've just seen you share the path already. Nevermind, I'll check how it goes.

JanStreffing:
I think a decent chunk of them time is also the fesom bin restart tar and untar.

JanStreffing:
Here are some more numbers from a month of daily runs:

```
Serial: {'min': '03:34', 'max': '07:57', 'mean': '05:03'}
Threads: {'min': '01:51', 'max': '03:40', 'mean': '03:18'}
[...]

PR #1435, opened 2026-02-09
Feat/dask file movements

mandresm:
Dask parallelization work for parallel file movements. This PR includes:
- Dask cluster initialization in esm_tools tasks with subprocess. Commands to submit the cluster are configurable through `slurm.yaml` and `pbs.yaml` of machines to add your favourite launcher. Subprocess submission is favoured over dask-mpi pythonic approach or SLURMCluster because we need to use the allocation resources and I rather relay on the HPC launcher than in dask-mpi which adds dependencies to MPI.
- Support for dask cluster with SLURM. PBS support will be on a different PR.
- Dask cluster status function
- Configurable parallel file movements using the dask cluster, with fall-back into threads parallelization (in one single node) and then into sequential if problems araise.
- pre_recipe steps
- improvements to provenance and fixes to `clean_provenance`
- a CONTEXT.md to be used with coding agents
- unit tests for the dask cluster

[Documentation of Parallel File Movements](https://esm-tools--1435.org.readthedocs.build/en/1435/esm_runscripts.html#parallel-file-movements)
[Documentation of new dask parameters](https://esm-tools--1435.org.readthedocs.build/en/1435/esm_variables.html#dask-variables)

## TODO
- [x] Documentation of the new feature
- [x] esm_tests on Levante
- [x] esm_tests changes review
- [x] test with monthly AWICM3

## Performance improvements
* Copying files in `tidy`  for monthly AWICM3 simulation with TCO95
  * No parallelization: 1 min 15 s
  * Threads: 3 s
  * Dask: 9 s + 4 s initialization (expected to scale better than threads for larger simulations)


mandresm:
@pgierz, can you have a look at this and provide a review? It's almost finished

mandresm:
Closes #1380

src/esm_runscripts/dask_cluster.py
```diff
@@ -0,0 +1,339 @@
...
+    Parameters
+    ----------
+    dask_scheduler_json : str
+        Path to the dask scheduler JSON file containing the TCP address.
+    client_timeout : float, optional
+        Timeout in seconds for connecting to the scheduler (default 0.01).
+    test : bool, optional
+        If ``True``, submit a test task to verify the cluster can execute
+        work (default ``False``).
+
```
pgierz: I would name `test` something else. `submit_test_task`?

src/esm_runscripts/dask_cluster.py
```diff
@@ -0,0 +1,339 @@
...
+        client.close()
+        return (DaskStatus.NO_WORKERS, n_workers)
+    elif test:
+        status = DaskStatus.RUNNING
+        try:
+            client.submit(test_dask).result()
+            logger.trace(
+                f"Dask test task succeeded on scheduler at {tcp_address} with {n_workers} workers"
+            )
+            status = DaskStatus.TESTED
```
pgierz, line 196: As above: a clearer name would be good here, in my opinion. 

```suggestion
    elif submit_test_task:
        status = DaskStatus.RUNNING
        try:
            client.submit(test_dask).result()
            logger.trace(
                f"Dask test task succeeded on scheduler at {tcp_address} with {n_workers} workers"
            )
            status = DaskStatus.TESTED
```

src/esm_runscripts/filelists.py
```diff
@@ -1204,18 +1253,20 @@ def copy_files(config, filetypes, source, target):
...
-                        ):
+                        if config["general"].get("parallel_file_movements", False) in [
+                            "threads",
+                            "dask",
+                        ]:
+                            # Parallel
                             future = client.submit(
                                 movement_method,
-                                config,
+                                {},  # do not pass a config to avoid serialization issues. Serialization is possible but even after that the config is too big to be sent to many workers efficiently
```
pgierz, line 1263: too big? its a few kbytes...? I would be curious what the workers say about this, are there obvious logs I could look at?

src/esm_runscripts/batch_system.py
pgierz, line 1: why do you constrain post_run_commands to only be defined in computer? What if others wanted to define a post_run_command? Are there reasons for/against that?

src/esm_runscripts/dask_cluster.py
```diff
@@ -0,0 +1,339 @@
...
+    The integer ordering allows comparison: a cluster with status >= RUNNING
+    is ready to accept work.
+    """
+
+    MISSING_JSON = 0
+    SCHEDULER_ERROR = 1
+    WORKERS_ERROR = 2
+    NO_WORKERS = 3
+    RUNNING = 4
+    TESTED = 5
```
pgierz, line 77: I am wondering: does Dask not have something internal that we can use for this sort of status?

src/esm_runscripts/dask_cluster.py
```diff
@@ -0,0 +1,339 @@
...
+        The unchanged simulation configuration.
+    """
+    # Load parameters
+    dask_config = config.get("dask", {})
+    dask_scheduler_json = dask_config["scheduler_json"]
+    log_scheduler = f'{config["general"]["thisrun_log_dir"]}/dask_scheduler.log'
+
+    nnodes = int(os.getenv(config["computer"]["nnodes_envvar"], 1))
+
+    init_scheduler_cmd = config["dask"].get("init_scheduler_cmd")
```
pgierz, line 282: Can you add in the comment what this looks like?

pgierz (commented):
Looks good. I have questions more than anything else, and a lot of this boils down to needing to "manually" inject the dask client commands into our generated run script, which is a side effect of switching from the Python to the shell world back and forth.

I am also very happy that there are unit tests around some of the more tricky things :)

src/esm_runscripts/batch_system.py
mandresm: Having post_run_commands in multiple sections can lead to conflicts and weird behavior as you cannot predict the final order of the commands from the different sections. In my view it also adds confusion for the users on where they should put their post_run_commands, when in reality all of them will end up in a single place: the `.run` file. It is now possible to call sections from every yaml, so it makes sense to control this parameter from one single section and let the user decide whether they want to `add_post_run_commands` or completely take control of all the `post_run_commands` and overwrite the whole list from a given yaml without having to track down every single component that has post_run_commands.

src/esm_runscripts/dask_cluster.py
mandresm: Not to my knowledge, or at least not as complete as this for the problems I have been facing when implemeting it. This status are very representative of the whole pains I went through :)

src/esm_runscripts/filelists.py
mandresm: The config is not serializable. It has a method to spit its equivalent serializable config. I tried with that and it was working, but then I thought "why not to keep it even simpler? all we want to do is to copy/move/link from one known source to a known target and config is only used for jinja templating, that can be done sequentially later on"

src/esm_runscripts/dask_cluster.py
mandresm: Agreed and added

src/esm_runscripts/dask_cluster.py
mandresm: Done!

mandresm:
Hi @pgierz, I have addressed all your comments and answer your questions. Please, let me know if you'd like me to still change something.

I still have to finish checking the esm_tests results in Levante, but they seem good so far.

src/esm_runscripts/dask_cluster.py
pgierz: I think you added a comment to the wrong one 😂 I meant for the initial command. That is probably something like `dask scheduler --host $SCHEDULER_NODE &
sleep 5`? Or alternatively `dask worker ...`

src/esm_runscripts/filelists.py
pgierz: This has the added advantage that you have a smaller communication load. Not that it should matter with the size of the things we are sending around, but if you only need the config to be passed for the sake of the call signature, you can just send an empty dict. 

...I will bet that we eventually stumble on this at some point in the future and say "emmm. Yes. Empty dict. Of course that makes sense". 

What happens to the return value of this?

src/esm_runscripts/dask_cluster.py
mandresm: 🤣 good point! I just fixed that :)

src/esm_runscripts/filelists.py
mandresm: It's false if it catches an error internally. If the error occurs when you ask for the output then it is catched as well. In both cases the thing is rerun sequentially:

https://github.com/esm-tools/esm_tools/blob/8747a4008fa0a15af59a27018fd16c8f1b30fe6a/src/esm_runscripts/filelists.py#L1276-L1288

src/esm_runscripts/dask_cluster.py
pgierz: excellent! I just thought this would be good to have as a comment once. Not that this should *ever* be done, but it might be that you want to so something like:

```
interface_result = init_schemduler_cmd.split("--interface")[1] 
interface_result = ib42
... re-glue command
```

Seeing an example in comment is good for that sort of thing.

src/esm_runscripts/filelists.py
pgierz: OK.

pgierz (approved):

mandresm:
#bump

src/esm_runscripts/batch_system.py
seb-wahl: I totally understand your arguments but we should then also move the pre_run_commmands into the computer section. Or are they reasons why they need to be in the components section?

mandresm: That is correct, they should be in the computer section, I miss migrating them there. Would you do that in your geomar branch?

Issue #1419, opened 2026-01-19
D6: File Handling Redesign and Syntax


mandresm:
# ESM-Tools File Operations Redesign — Filedicts

This deliverable has been using as its base the old GitHub project for the refactoring of file handling https://github.com/orgs/esm-tools/projects/12, that already included many of the current requirements and design considerations reflected in this document. It is also based in existing unfinished developments in `sprint/filedicts/main` branch that correspond to that old GitHub project. The deliverable has been written using @claude -code in planning mode, using the existing resources mentioned above and new prompted design considerations.

## Overview

ESM-Tools currently handles file operations during `esm_runscripts` through a fragmented system of parallel flat dictionaries (`input_files`, `input_sources`, `input_in_work`, `forcing_files`, etc.) distributed across multiple sections of
component configs. This approach is hard to read, difficult to extend, and couples file metadata to intermediate directory structures that add unnecessary complexity and file duplication.

This document describes the new design — **filedicts** — which unifies file definitions into structured objects, removes intermediate staging directories inside `run_DATE/`, and simplifies the operation model, where each phase declares in the backend the file operation directions.

----

#### `files:` block

Partly taken from https://github.com/orgs/esm-tools/projects/12/views/1?pane=issue&itemId=8130162

Discussion in: [esm-tools/esm-design#2](https://github.com/esm-tools/esm-design/issues/2)

Files are defined in a `files:` block, grouped by type. The type is inferred from the group key: no `type:` attribute needed per file. Each group may contain a `defaults:` sub-block for shared attributes. Individual file entries can be scalars (shorthand) or full objects.

```yaml
files:
    input:
        defaults:
            path_in_pool: ${echam.input_dir}
            prepare: copy
        cldoptprops:                           # null → name = label in all locations
        janspec: janspec.nc                    # scalar → name_in_pool
        jansurf:                               # object → explicit overrides only
            name_in_pool: jansurf.nc
            name_in_run: unit.24
        rrtmglw: /other/pool/path/rrtmg.nc    # has "/" → full path, ignores defaults
    forcing:
        defaults:
            path_in_pool: ${echam.forcing_dir}
            prepare: link
        sst: pisst.nc
        sic:
            name_in_pool: pisic${current_date.year}.nc
            name_in_run: unit.96
            include_years_before: 1
    outdata:
        defaults:
            tidy: copy
        histogram:                             # null → name_in_run = label
        atm_data: atmosphere_output.nc         # scalar → name_in_run
[...]

PR #1448, opened 2026-02-25
add a first draft of D6 - filedicts design

mandresm:
This PR includes a draft for D6 - filedicts design. I have copy/pasted the markdown in the description of issue #1419 

@pgierz, please go ahead and review it. Even if I used claude-code to produce it I have put a lot of though into it and revised the old Filedicts project to don't re-think the design from scratch (reuse as much as possible from that project). So hopefully you don't find that the document is LLM-garbage :)

### DIFF
--- a/filedicts_design.md
+++ b/filedicts_design.md
@@ -1,13 +1,13 @@
 # ESM-Tools File Operations Redesign — Filedicts
 
-This deliverable has been using as its base the old GitHub project for the refactoring of file handling https://github.com/orgs/esm-tools/projects/12, that already included many of the current requirements and design considerations reflected in this document. It is also based in existing unfinished developments in `sprint/filedicts/main` branch that correspond to that old GitHub project. The deliverable has been written using @claude -code in planning mode, using the existing resources mentioned above and new prompted design considerations.
+This deliverable uses as its base the old GitHub project for the refactoring of file handling https://github.com/orgs/esm-tools/projects/12, that already included many of the current requirements and design considerations reflected in this document. It is also based in existing unfinished developments in `sprint/filedicts/main` branch that belongs to that old GitHub project. The deliverable has been written using @claude -code in planning mode, using the existing resources mentioned above and new prompted design considerations.
 
 ## Overview
 
 ESM-Tools currently handles file operations during `esm_runscripts` through a fragmented system of parallel flat dictionaries (`input_files`, `input_sources`, `input_in_work`, `forcing_files`, etc.) distributed across multiple sections of
 component configs. This approach is hard to read, difficult to extend, and couples file metadata to intermediate directory structures that add unnecessary complexity and file duplication.
 
-This document describes the new design — **filedicts** — which unifies file definitions into structured objects, removes intermediate staging directories inside `run_DATE/`, and simplifies the operation model, where each phase declares in the backend the file operation directions.
+This document describes the new design — **filedicts** — which unifies file definitions into structured objects, removes intermediate staging directories inside `run_DATE/`, and `run_DATE` itself in favor of `work_DATE` (one work per run at the root of the experiment dir), and simplifies the operation model, where each phase declares in the backend the file operation directions.
 
 ----
 
@@ -57,22 +57,25 @@ files:
             name_in_run: restart.nc
 ```
 
+File types are limitted to the current existing ones, and there will be a mechanism to control that the "types" are correct.
+
+Note about the `null` values: we might want to avoid implementing this feature. Have None as the value (or in fact, anything that might be considered `false`-y) will make anything one does with comparison logic fragile. Also, tracking the provenance of the value might be challenging, but maybe not, as `None`s also have provenance in the current implementation. Maybe, it's a matter of internally swapping the `None` of the files with the key name, while preserving the `None`'s provenance. Placeholders with `${}` could be a problem, because they might not be resolved inside the keys. All this is something to consider during the implementation.
+
 #### File object attributes
 
 | Attribute | Description |
 |---|---|
 | `name_in_pool` | Filename in the pool/source location |
-| `name_in_run` | Filename in `run_DATE/` (the working directory) |
+| `name_in_work` | Filename in `work_DATE/` (the working directory) |
 | `name_in_exp` | Filename in the experiment tree (defaults to `name_in_run`) |
 | `path_in_pool` | Path to the file in the pool (excluding filename) |
-| `prepare` | Operation for `pool/exp → run`: `copy`, `link`, `move` |
-| `tidy` | Operation for `run → exp/<type>`: `copy`, `link`, `move` |
-| `include_years_before` | Years before current to include (requires `StringWithDate`) |
-| `include_years_after` | Years after current to include (requires `StringWithDate`) |
-| `description` | Human-readable description |
+| `prepare_operation` | Operation for `pool/exp → run`: `copy`, `link`, `move` |
+| `tidy_operation` | Operation for `run → exp/<type>`: `copy`, `link`, `move` |
+| `include_timedelta_before` | For example, years before current to include (requires `StringWithDate`) |
+| `include_timdelta_after` | For example, years after current to include (requires `StringWithDate`) |
+| `description` | Human-readable description [optional] |
 | `allowed_to_be_missing` | If `true`, missing file does not raise an error |
 | `is_reusable` | if `true` copy from exp instead of from pool, like bins and inputs (default: false) |
-| description | a file description [optional] |
 | filetype | filetype, like NetCDF [optional], not sure if we should implement this attribute |
 
 #### Files with varying paths depending on dates
--- a/filedicts_design.md
+++ b/filedicts_design.md
@@ -1,0 +1,352 @@
+# ESM-Tools File Operations Redesign — Filedicts
+
+This deliverable has been using as its base the old GitHub project for the refactoring of file handling https://github.com/orgs/esm-tools/projects/12, that already included many of the current requirements and design considerations reflected in this document. It is also based in existing unfinished developments in `sprint/filedicts/main` branch that correspond to that old GitHub project. The deliverable has been written using @claude -code in planning mode, using the existing resources mentioned above and new prompted design considerations.
+
+## Overview
+
+ESM-Tools currently handles file operations during `esm_runscripts` through a fragmented system of parallel flat dictionaries (`input_files`, `input_sources`, `input_in_work`, `forcing_files`, etc.) distributed across multiple sections of
+component configs. This approach is hard to read, difficult to extend, and couples file metadata to intermediate directory structures that add unnecessary complexity and file duplication.
+
+This document describes the new design — **filedicts** — which unifies file definitions into structured objects, removes intermediate staging directories inside `run_DATE/`, and simplifies the operation model, where each phase declares in the backend the file operation directions.
+
+----
+
+## Design
+
+### Syntax
+
+#### `files:` block
+
+Partly taken from https://github.com/orgs/esm-tools/projects/12/views/1?pane=issue&itemId=8130162
+
+Discussion in: [esm-tools/esm-design#2](https://github.com/esm-tools/esm-design/issues/2)
+
+Files are defined in a `files:` block, grouped by type. The type is inferred from the group key: no `type:` attribute needed per file. Each group may contain a `defaults:` sub-block for shared attributes. Individual file entries can be scalars (shorthand) or full objects.
+
+```yaml
+files:
+    input:
+        defaults:
+            path_in_pool: ${echam.input_dir}
+            prepare: copy
+        cldoptprops:                           # null → name = label in all locations
+        janspec: janspec.nc                    # scalar → name_in_pool
+        jansurf:                               # object → explicit overrides only
+            name_in_pool: jansurf.nc
+            name_in_run: unit.24
+        rrtmglw: /other/pool/path/rrtmg.nc    # has "/" → full path, ignores defaults
+    forcing:
+        defaults:
+            path_in_pool: ${echam.forcing_dir}
+            prepare: link
+        sst: pisst.nc
+        sic:
+            name_in_pool: pisic${current_date.year}.nc
+            name_in_run: unit.96
+            include_years_before: 1
+    outdata:
+        defaults:
+            tidy: copy
+        histogram:                             # null → name_in_run = label
+        atm_data: atmosphere_output.nc         # scalar → name_in_run
+    restart:
+        defaults:
+            prepare: copy
+            tidy: copy
+        jan_restart:
+            name_in_run: restart.nc
+```
+
+#### File object attributes
+
+| Attribute | Description |
+|---|---|
+| `name_in_pool` | Filename in the pool/source location |
+| `name_in_run` | Filename in `run_DATE/` (the working directory) |
+| `name_in_exp` | Filename in the experiment tree (defaults to `name_in_run`) |
+| `path_in_pool` | Path to the file in the pool (excluding filename) |
+| `prepare` | Operation for `pool/exp → run`: `copy`, `link`, `move` |
+| `tidy` | Operation for `run → exp/<type>`: `copy`, `link`, `move` |
+| `include_years_before` | Years before current to include (requires `StringWithDate`) |
+| `include_years_after` | Years after current to include (requires `StringWithDate`) |
+| `description` | Human-readable description |
+| `allowed_to_be_missing` | If `true`, missing file does not raise an error |
+| `is_reusable` | if `true` copy from exp instead of from pool, like bins and inputs (default: false) |
+| description | a file description [optional] |
+| filetype | filetype, like NetCDF [optional], not sure if we should implement this attribute |
+
+#### Files with varying paths depending on dates
+
+For files whose paths change depending on dates the syntax will be:
+```yaml
+files:
+    <file1_label>:
+        description: "some string"
+        path_in_pool:
+            "/path/to/first/file${year}":
+                   from: <year>
+                   to: <year>
+            "/path/to/second/file${year}":
+                   from: <year>
+```
+
+#### File selection via `choose_`
+
+All files defined in `files:` are active by default. Scenarios override specific attributes via `choose_`. Only the differing attributes need to be specified:
+
+```yaml
+files:
+    forcing:
+        defaults:
+            path_in_pool: ${echam.forcing_dir}
+        sst:
+            name_in_pool: pisst.nc      # default: PI-CTRL
+            name_in_run: unit.20
+
+choose_scenario:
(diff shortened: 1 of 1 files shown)


filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+        jan_restart:
+            name_in_run: restart.nc
+```
+
+#### File object attributes
+
+| Attribute | Description |
+|---|---|
+| `name_in_pool` | Filename in the pool/source location |
+| `name_in_run` | Filename in `run_DATE/` (the working directory) |
```
pgierz: Maybe this should be `name_in_work`, and we have `work_DATE` instead of `run_DATE`. I need to think about how I feel about that though...

In any case, getting rid of `run_*` is a *very very good* thing. 👍🏻

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+| `name_in_exp` | Filename in the experiment tree (defaults to `name_in_run`) |
+| `path_in_pool` | Path to the file in the pool (excluding filename) |
+| `prepare` | Operation for `pool/exp → run`: `copy`, `link`, `move` |
+| `tidy` | Operation for `run → exp/<type>`: `copy`, `link`, `move` |
+| `include_years_before` | Years before current to include (requires `StringWithDate`) |
+| `include_years_after` | Years after current to include (requires `StringWithDate`) |
+| `description` | Human-readable description |
+| `allowed_to_be_missing` | If `true`, missing file does not raise an error |
+| `is_reusable` | if `true` copy from exp instead of from pool, like bins and inputs (default: false) |
+| description | a file description [optional] |
```
pgierz: You have `description` twice

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+### Syntax
+
+#### `files:` block
+
+Partly taken from https://github.com/orgs/esm-tools/projects/12/views/1?pane=issue&itemId=8130162
+
+Discussion in: [esm-tools/esm-design#2](https://github.com/esm-tools/esm-design/issues/2)
+
+Files are defined in a `files:` block, grouped by type. The type is inferred from the group key: no `type:` attribute needed per file. Each group may contain a `defaults:` sub-block for shared attributes. Individual file entries can be scalars (shorthand) or full objects.
```
pgierz, line 24: Is there some mechanism to control that the "types" are correct? Is it then trivial to add a new type by having e.g. `files.namelists`? 

That example is chosen on purpose: some users may not realize we have `files.config` for namelists, implement a new category of files, and then we have a whole new folder that pops up in the experiment tree, auto-magically. May be cool if it works cleverly out of the box, but may also be a source of accidental confusion.

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+```yaml
+files:
+    input:
+        defaults:
+            path_in_pool: ${echam.input_dir}
+            prepare: copy
+        cldoptprops:                           # null → name = label in all locations
```
pgierz, line 32: I like this, but is is going be very easy to shoot yourself in the foot. Have None as the right-hand-side (or in fact, anything that might be considered "false"-y) will make anything one does comparison logic with fragile.

When we implement this, I'd consider having some kind of placeholder style object here that we inject while parsing.

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+#### File object attributes
+
+| Attribute | Description |
+|---|---|
+| `name_in_pool` | Filename in the pool/source location |
+| `name_in_run` | Filename in `run_DATE/` (the working directory) |
+| `name_in_exp` | Filename in the experiment tree (defaults to `name_in_run`) |
+| `path_in_pool` | Path to the file in the pool (excluding filename) |
+| `prepare` | Operation for `pool/exp → run`: `copy`, `link`, `move` |
```
pgierz: Not sure I like the name `prepare` for this. Maybe `operation`?

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+| `tidy` | Operation for `run → exp/<type>`: `copy`, `link`, `move` |
```
pgierz: Ah...ok...I see. Then perhaps: `prepare_op` and `tidy_op`? We can discuss...

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+| `include_years_before` | Years before current to include (requires `StringWithDate`) |
+| `include_years_after` | Years after current to include (requires `StringWithDate`) |
```
pgierz: We could consider doing this in a generic way so that it is not necessarily years. Ideally you would have something like `include_timedelta_before` (or after) and then have something like this:

```yaml
files:
    input:
        my_favorite_file:
            include_timedelta_before: datetime.TimeDelta(years=1)  # I forgot the actual object signature, you you get the idea
```

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+files:
+    <file1_label>:
+        description: "some string"
+        path_in_pool:
+            "/path/to/first/file${year}":
+                   from: <year>
+                   to: <year>
+            "/path/to/second/file${year}":
+                   from: <year>
+```
```
pgierz, line 94: Here again, it might be good to have a generic thing to handle any resolution of time, not just years.

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+                sst:
+                    name_in_pool: histsst.nc    # only override what changes
+
+    ssp585:
+        add_files:                              # add a file not in base config
+            forcing:
+                ozone:
+                    name_in_pool: ozone_ssp585.nc
+                    name_in_run: ozone.nc
+```
```
pgierz, line 122: This is effectively a conditional dict merge, if I am reading it correctly? Sounds like the right call to me.

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+A reserved `include:` list mechanism per type could be implemented in the future if explicit whitelisting is needed, or perhaps it can be implemente also as an attribute of file objects.
```
pgierz, line 124: I am afraid I don't understand what you mean here. You could have an attribute on your file object that defines a conditional, and the parser could then spit out a true or false for that. I'll see if I can invent an example...

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+In order to simplify the amount of writing needed for specifying file operations we allow for shorthand file definitions, consisting of a label (key) followed by a path (value). The path can be an absolute path (starting with `/`) or a relative path. The filedicts attribute that takes the value of the shorthand-path is determined by the source of file operations for that specific phase. For example:
+
+- For `prepare` input-like file types `input`, `forcing`, `config` and `restart` the scalar is `name_in_pool`, or `name_in_exp`.
+- For `tidy` output-like file types `outdata`, `restart`, `log`, and `mon` the scalar is `name_in_run`.
+
+Other shorthand rules are:
+- For `null` values the label is used as name in all relevant locations. The label cannot have placeholders (no `${}`).
+- Plain string (no `/`) → filename only; path comes from `defaults.path_in_pool`
+- String with `/` → full path, overrides `defaults.path_in_pool`
```
pgierz, line 136: As above, I feel like null is dangerous, we need to think about this

Missing here is nested relative paths. If I remember correctly, some models have a nested subfolder in `work` (or, I guess now `run`):

```
files:
    restart:
        my_fancy_nested_restart: run_${DATE}/component_${year}/my_restart.nc
```

If I understand it as designed, that will then end in the restart folder during the tidy phase as `my_restart.nc`, correct?

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+```
+
+If `include_years_before/after` is set but the name resolves to a plain `str`
+(i.e. the string contained no date variable), filedicts raises a `FileDictError`.
+
+#### `simulation_files` (removed)
+
+The `simulation_files` / `<type>_files` selection mechanism is replaced by the
+grouped `files:` block with Option C selection (all defined files active,
+scenario-specific changes via `choose_`).
```
pgierz, line 165: yes! 🥳

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+            prepare: link
+        ozone:
+            name_in_pool: ozon${current_year}.nc
+            name_in_run: ozon${current_year}.nc
+            include_years_before: 1
+            include_years_after: 0
+```
+
+If `include_years_before/after` is set but the name resolves to a plain `str`
+(i.e. the string contained no date variable), filedicts raises a `FileDictError`.
```
pgierz, line 159: I like this a lot. As above, my only thought is to do this in a general way for other temporal frequencies as well. Specifying just one year becomes trivial, you just need some wordier syntax. I am not opposed to it, and it makes it more flexible.

What do the others think?

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+Three locations replace the previous multi-level structure:
+
+| Location | Description |
+|---|---|
+| `pool` | Source on the HPC system — input data, forcing data |
+| `run` | `run_DATE/` — the flat working directory during simulation |
+| `exp` | Experiment tree — persistent storage, structure preserved |
+
+**`run_DATE/` is now flat.** Intermediate staging directories (`run_DATE/input/`, `run_DATE/work/`, `run_DATE/outdata/`, etc.) will be removed. All files land directly in `run_DATE/`. The `exp` tree structure won't be changed (`exp/outdata/<component>/`, `exp/restart/<component>/`, etc.).
```
pgierz, line 179: yes! 🥳 🎊 🚀

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+```yaml
+files:
+    forcing:
+        ozone:
+            name_in_pool: ozon${current_year}.nc
+            prepare: copy
+```
+
+The direction is defined in the phase itself as a system invariant, not user-configurable. This is subject to change in the future if needed. The `FileTypes` enum declares which phases apply to each type and the `exp` subdirectory name. Default operations (`copy`/`link`/`move`) per type are defined in `per_model_defaults` in the system defaults YAML.
```
pgierz, line 195: I don't foresee needing to change the direction. What did you have in mind here?

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+- [ ] more binary info in finished_config #990 
+
+---
+
+## Testing
+
+### Unit tests
+
+One unit test per feature, developed alongside each feature following the
+red/green/refactor cycle (see Development Workflow below).
```
pgierz, line 264: I would say "one unit test collection", not "one unit test". It may be we want to nail down edge case logic in some features

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+### Checksum tests (esm_tests)
+
+To verify the refactor is behaviour-preserving — identical files end up in
+identical locations:
+
+1. Compute checksums for all files moved/copied/linked on the `release` branch
+2. Compute checksums for the same run on `sprint/filedicts/main`
+3. Compare checksums for both `prepare` and `tidy` phases
+
+Any mismatch is a regression. Checksum tests run via `esm_tests`.
```
pgierz, line 283: 👍🏻

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+---
+
+## Retrocompatibility
+
+The new syntax will substitute to the old syntax. The old syntax won't be supported anymore from the new release version. That means that old runscripts referencing the old file lists will have to be changed to work with the new version containing this refactoring. Same goes for the configs, however, the configs' syntax will be changed to the new one as part of this release, and include test runs with esm_tests. Resuming or branching off old simulations will still be possible because the file structure in `exp` won't change.
+
+The users will be informed of such changes.
```
pgierz, line 291: 👍🏻 (actually +100) 

If we spent any effort on making this backward compatible, I would vote that it will be in automatically translating and gently telling the user "Hey, I have updated all of your configs" via `<translation tool>`. 

maintaining two syntaxes when the old one is supposed to be replaced sounds like a bad time.

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+---
+
+## Refactoring guidelines
+
+### General
+
+Taken from https://github.com/orgs/esm-tools/projects/12/views/1?pane=issue&itemId=8002017
+
+- One task, one function
```
pgierz, line 301: yes

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+- Do not change the YAML syntax excessively
```
pgierz, line 302: yes again

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+  - Runscripts and config files need minimal changes
```
pgierz, line 303: not sure I agree with this one. By nature, some of these changes will "look bloated". they aren't, they are making it more precise.

filedicts_design.md
```diff
@@ -0,0 +1,352 @@
...
+  - Changes should make configs easier to write and understand
```
pgierz, line 304: bigger yes

pgierz (commented):
Looks great. I have some specific comments in some of the sections that may be worth a more detailed discussion.

Overall, fantastic design document. I feel much more prepared in building this well after we have settled on everything :)

filedicts_design.md
pgierz: Something my LLM came up with which I think is worth considering:

>   Line 133: Nested relative paths syntax
> 
>   The question: how to handle run_DATE/subdir/file.nc?
> 
>   Proposed syntax:
> 
> ```yaml
>   files:
>       outdata:
>           my_nested_output:
>               name_in_run: component_output/year_${year}/output.nc
>               name_in_exp: output_${year}.nc  # flattened to basename
> 
> ```
>   Rules:
>   - `name_in_run` can contain path separators (relative to `run_DATE/`)
>   - `name_in_exp` can contain path separators (relative to `exp/<type>/<component>/`)
>   - If `name_in_exp` omitted, default to basename of `name_in_run`
> 
>   This means nested structure is explicitly controlled:
> 
>   * Flatten (default behavior)
>   name_in_run: subdir/output.nc
>   --> exp/outdata/component/output.nc
> 
>   * Preserve structure (explicit)
>   name_in_run: subdir/output.nc
>   name_in_exp: subdir/output.nc
>   --> exp/outdata/component/subdir/output.nc
> 
>   Recommendation: Add this to the document with explicit rules about path handling.

filedicts_design.md
pgierz: Again from the LLM:  

>   
>   Line 156: Generic timedelta
> 
>   Three options:
> 
>   1. Keep simple: `include_years_before: 1` (current design)
>   2. Go generic: `include_timedelta_before: {years: 1, months: 6}`
>   3. Support both: shorthand for years, verbose for complex cases
> 
>   Option 2 syntax:
> ```yaml
>   files:
>       forcing:
>           ozone:
>               include_timedelta_before:
>                   years: 1
>               include_timedelta_after:
>                   months: 6
> ```
>   Recommendation: Go with option 2 (generic only). It's slightly more verbose for the common case but avoids adding a second syntax later. Years become {years: 1}.
> 
> ❯ Can we write `files.forcing.ozone.include_timedelta_before: {years: 1}` in YAML? Is that allowed, like square brackets for lists?
> 
> ⏺ Yes, that's valid YAML. Curly braces are "flow style" for mappings, just like square brackets are flow style for sequences.
>```yaml 
>   # Flow style (inline)
>   include_timedelta_before: {years: 1}
>   include_timedelta_before: {years: 1, months: 6}
> 
>   # Block style (expanded)
>   include_timedelta_before:
>       years: 1
>       months: 6
> 
>   # Both parse identically to Python:
>   # {'include_timedelta_before': {'years': 1, 'months': 6}}
> ```
>   So the compact version works fine:
> ```yaml
>   files:
>       forcing:
>           ozone:
>               name_in_pool: ozon${current_year}.nc
>               include_timedelta_before: {years: 1}
>               include_timedelta_after: {months: 6}
> ```

pgierz: My own note here would be to keep it simple: if you put an integer on RHS instead of a dict, you automatically mean years. Then, `include_years_before: 1` internally becomes `include_timedelta_before: {years: 1}` and something (I guess the parser?) beeps at you to please be nice and write it correctly. It would not break though.

filedicts_design.md
pgierz: Correction/Clarification: getting rid of `run_*` *as a full subfolder nest* is a good thing ;-)

filedicts_design.md
pgierz: See my comment below for an idea of how this could be done cleanly

filedicts_design.md
pgierz: I am not happy that I am about to suggest this, but we already have a mechanism to do `eval`... we really *really* should not use it further, because `eval` is dangerous for all kinds of reasons. Buuuuuttttttt this would be one way:

```yaml
  files:
      input:
          ozone_ssp:
              name_in_pool: ozone_ssp585.nc
              include: "$(( '${scenario}' == 'ssp585' ))"

          debug_dump:
              name_in_run: debug.nc
              # imaginary switch that might turn on some namelist setting
              # and write a new output file
              include: "$(( ${debug_mode} and ${verbosity} > 2 ))"
```

filedicts_design.md
mandresm: I think you are right. And there is one extra reason to favor `work_DATE` over `run_DATE`: with my original suggestion, an ls in a new experiment would look identical to an old experiment in the sense that the user will see `run_DATE` folders, so the user won't be directly aware whether this is run with the newer ESM-Tools or the older one. If they see `work_DATE` instead, then they know right away "this is new" and it is also self understandable as it is telling them that now `work` is in their experiment root directly. I'll add this change to the design document.

filedicts_design.md
mandresm: I agree, we should have a mechanism to prevent from magically creating new file types. I have updated the document with that.

filedicts_design.md
mandresm: I agree. It would also be tricky to carry the provenance of it. I'll make a note in the document about this considerations.

filedicts_design.md
mandresm: I agree. What about `prepare_operation`, `tidy_operation`, etc.? I'm also fine with `op`.

filedicts_design.md
mandresm: We should keep this in mind for the implementation. I think it's a good idea indeed :)

filedicts_design.md
mandresm: Correct, and this is already native to the esm_parser so no coding require for this.

mandresm:
@JanStreffing asked for an additional tidy phase that can handle certain files indicated by the user (not restarts) and that can run in a separate sbatch, to save time for high resolution sims. I should include this into the plan.

PR #1449, opened 2026-03-06
add hostfiles to the tracked files in last_tested for esm_tests

mandresm:
add hostfiles to the tracked files in last_tested for esm_tests. I need this for the `geomar` merge, and in general it would be nice to track also hostfiles.

### DIFF
--- a/src/esm_tests/output.py
+++ b/src/esm_tests/output.py
@@ -19,7 +19,7 @@ be = "\033[0m"
 # Define default files for comparisson
 compare_files = {
     "comp": ["comp-"],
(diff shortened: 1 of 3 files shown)


pgierz (approved):
Looks good, approve. Some things worth noting for the future:

* Duplicate `compare_files` in output.py and tests.py makes sense for how it is structured, but we should at some point go for a single point of truth

* Ideally I'd like to use pathlib, but that would be a full refactor inside of esm-tests. Out of scope for this PR, we can catch up on that later

PR #1482, opened 2026-06-10
Feat/conda support

mandresm:
Adds support to run esm_tools from a conda environment, even if the environment changes at the top of the `.run` script destroy it (e.g. module purge).

It does this by extracting conda information (`CONDA_PREFIX` and `CONDA_ROOT`) during the `prepare` job and storing that info in a `<expid>/run-DATE/config/conda_info.yaml`. Then it uses this info to restore the conda environment after the `srun` command.

### Other features
- sets a config var `computer.launched_with_conda` so that it can be used in `choose_` blocks
- it also allows the user to define in the runscript their own env and conda root:
  ```yaml
  conda:
    env: <conda_env>
    root: <conda_root>
  ```

## TODOs

- [x] docstrings
- [x] documentation


src/esm_runscripts/batch_system.py
```diff
@@ -399,8 +400,38 @@ def get_post_run_commands(config):
...
 
+    @staticmethod
+    def get_conda_activate_commands(config):
+        commands = []
+
+        # Do not mix conda and venvs
+        if config["general"].get("use_venv", False):
+            return commands
+
+        conda_env = config["computer"].get("conda", {}).get("env", False)
```
pgierz: Do we use dpath yet?

src/esm_runscripts/batch_system.py
```diff
@@ -669,18 +699,21 @@ def write_simple_runscript(config, cluster, batch_or_shell="batch"):
                     cluster
                 ]["next_submit"]
 
-                # extra entries for each subjob
-                post_run_commands = batch_system.get_post_run_commands(config)
```
pgierz, line 672: Why is this block moved down?

src/esm_runscripts/conda_env.py
```diff
@@ -0,0 +1,39 @@
...
+    conda_exe = os.getenv("CONDA_EXE")
+    conda_root = str(Path(conda_exe).parent.parent) if conda_exe else None
+
+    return conda_env, conda_root
+
+def set_launched_with_conda(config):
+    conda_env, _ = get_conda_info()
+    config["computer"]["launched_with_conda"] = True if conda_env else False
+
+    return config
```
pgierz, line 53: Is there a check missing for existence of conda_dir? Also, why force strings instead of keeping Path?

src/esm_runscripts/conda_env.py
```diff
@@ -0,0 +1,39 @@
...
+    conda_info_file = f'{config["general"]["thisrun_config_dir"]}/conda_info.yaml'
+    if not os.path.isfile(conda_info_file):
+        return None, None
+
+    with open(conda_info_file, "r") as f:
+        conda_info = yaml.safe_load(f)
+
+    return conda_info.get("conda_env"), conda_info.get("conda_root")
+
+def write_conda_info_file(config):
```
pgierz, line 90: I think this one needs a small change:

Missing a return config in the "default" logic branch. Alternative return None inside the if block, but not in one case the config and in their implicit None...

pgierz (changes requested):
Looks good, but I have some small improvements

src/esm_runscripts/batch_system.py
mandresm: Mainly because of the line 680/705:

```python
                runfile.write("process=$!\n\n")
```

We need that `$!` caches the srun process and not the last post_run_command_process.

With this change, things look like this, which I think is correct:

```bash
time srun -l --kill-on-bad-exit=1 --cpu_bind=none \
--nodes=1 --ntasks=128 --ntasks-per-node=128 --cpus-per-task=1 --export=ALL,OMP_NUM_THREADS=1 ./fesom \
: --nodes=2 --ntasks=128 --ntasks-per-node=64 --cpus-per-task=2 --export=ALL,OMP_NUM_THREADS=2 ./oifs -v ecmwf -e awi3 \
: --nodes=1 --ntasks=1 --ntasks-per-node=1 --cpus-per-task=128 --export=ALL,OMP_NUM_THREADS=128 ./rnfma \
: --nodes=1 --ntasks=4 --ntasks-per-node=4 --cpus-per-task=32 --export=ALL,OMP_NUM_THREADS=32 ./xios.x  2>&1 &

# Call to esm_runscript to start subjobs:
# ['tidy']
process=$!

# Recover batch system environment variables
source /albedo/scratch/user/mandresm//conda_test/run_20000101-20000101/work//batch_system.env; rm /albedo/scratch/user/mandresm//conda_test/run_20000101-20000101/work//batch_system.env
# Activate conda environment
source /albedo/soft/sw/spack-sw/mambaforge/22.9.0-2-Linux-x86_64-cchydqq/bin/activate
conda activate /albedo/home/mandresm/.conda/envs/esm_tools

# Comment the following line if you don't want esm_runscripts to restart:
cd /albedo/scratch/user/mandresm//conda_test/scripts/
esm_runscripts awicm3-v3.1-albedo-TCO95L91-CORE2.yaml -e conda_test -t observe_compute -p ${process} -s 20000101 -r 1 -v --last-jobtype prepcompute --open-run
```

src/esm_runscripts/conda_env.py
mandresm: > Is there a check missing for existence of conda_dir?

Can you give an example of what do you mean? Like an `if not os.path.isfile(conda_exe)` then error?

> Also, why force strings instead of keeping Path?

Because `yaml.dump` in line 39 prints an pathlib object and not the absolute path. I wanted to keep things simple, and since conda_env is also a plain string, I thought to keep things uniform.

src/esm_runscripts/conda_env.py
mandresm: oh boy, good catch!

src/esm_runscripts/conda_env.py
```diff
@@ -0,0 +1,39 @@
...
+    conda_env, conda_root = get_conda_info()
+
+    if not conda_env:
+        return config
+
+    conda_info = {"conda_env": conda_env, "conda_root": conda_root}
+
+    conda_info_file = f'{config["general"]["thisrun_config_dir"]}/conda_info.yaml'
+    with open(conda_info_file, "w") as f:
+        yaml.dump(conda_info, f)
```
mandresm, line 126:
```suggestion
        yaml.dump(conda_info, f)
        
    return config
```

mandresm (commented):
Thanks for your review @pgierz!

See my replies :)

seb-wahl:
Please make sure this really nice feature is also documented.  I (and my "customers" at GEOMAR) alls use conda environments to manage their ESM-Tools installation.

src/esm_runscripts/conda_env.py
mandresm: Sorted!

mandresm:
@seb-wahl, can you review and approve if no further comments are needed?

Here is the documentation of the feature:
https://esm-tools--1482.org.readthedocs.build/en/1482/installation.html#installing-in-a-conda-environment
https://esm-tools--1482.org.readthedocs.build/en/1482/esm_runscripts.html#running-an-experiment-with-conda

mandresm:
The changes in the esm_tests can be approved. They are expected.

mandresm:
I'm merging this as it is blocking other PRs and it has been at least reviewed once by @pgierz

mandresm:
#bump

PR #1473, opened 2026-05-18
feat(esm-tools-plus/simcat): add esm_catalog — STAC-based experiment catalog

siligam:
## Context

Part of the **[ESM-Tools-plus/simcat](https://github.com/esm-tools/esm_tools/tree/esm-tools-plus/simcat/collapsed-collections)** initiative, which adds a STAC-based experiment catalog to ESM-Tools for indexing, querying and browsing climate model output.

This PR introduces `esm_catalog` — the catalog API, scanner, and CLI.
A companion PR adds [`esm_viz`](../../pulls) — the visualization service.

> **Reviewer tip:** this is a new top-level package. All files are additions. Start with [`src/esm_catalog/ARCHITECTURE.md`](src/esm_catalog/ARCHITECTURE.md) for an overview, then `src/esm_catalog/cli.py` for the user-facing entry points.

---

### `src/esm_catalog/` — the package (108 files, ~32k lines)

| Sub-package | Purpose |
|---|---|
| `api/` | FastAPI STAC server (stac-fastapi-api backed by DuckDB) |
| `scan/` | NetCDF / GRIB / namelist scanner; writes STAC items to DuckDB |
| `stac/` | STAC Collection/Item builders + ESM-specific extensions |
| `storage/` | DuckDB storage layer + personal collections |
| `hpc/` | HPC system detection (Albedo, Levante, …) |
| `integration/` | ESM-Tools runscript hooks for auto-scanning |
| `mcp/` | Model Context Protocol server for LLM tool access |
| `cli.py` | `esm-catalog` CLI (scan, serve, mcp, reindex, …) |

**STAC extensions defined:**

- `hpc` — HPC system name on assets (`hpc:system`)
- `namelist` — F90 namelist parameters (`nml:*`)
- `paleo` — deep-time metadata (`paleo:years_bp`)
- `datacube` — variable/dimension axes
- `contacts` — experiment owner info

**API highlights:**

- `GET /collections` — all experiments, filterable via CQL2
- `GET /experiments` — experiment-level search with CQL2 (`component=`, `variable=`, `nml:radctl.co2vmr > 284`)
- `GET /search` — STAC item search
- `GET /queryables` — OGC queryables for browser filter UI
- `GET /paleo-presets` — named paleo time periods (LGM, mid-Holocene, …)
- `GET /collections/{id}/nml-parameters` — namelist params for an experiment
- Personal collections CRUD (`/personal/…`)
- Catalog registry API (`/catalogs`) + web admin UI at `/ui`

[...]

PR #1483, opened 2026-06-13
feat(esm_catalog): PR-0 — package scaffold

siligam:
## PR-0 — esm_catalog package scaffold

First slice of breaking up #1473 (the 32k-line dump) into small, reviewable PRs, per @mandresm's review request. This PR is **pure scaffold — no functionality yet** — so the following feature PRs land on a clean, CI-tested foundation.

### What's here (+102 lines, 9 files)
- `src/esm_catalog/__init__.py` — package marker + `__version__`
- `src/esm_catalog/cli.py` — minimal Click group (`esm-catalog`); subcommands arrive in later PRs
- `setup.py` — `esm-catalog` console script + a `catalog` optional extra (pystac/shapely/universal-pathlib). **No `mcp` extra** — the MCP/LLM layer is being dropped from the product (archived on tag `archive/mcp-llm`)
- `src/esm_catalog/tests/test_smoke.py` — import + `--version` smoke test
- `.github/workflows/esm-catalog-tests.yml` — dedicated CI job (installs `.[catalog]`, runs the catalog tests on Py 3.9 + 3.12)
- `setup.cfg` — adds `src/esm_catalog` to `[tool:pytest] norecursedirs` so the **main** esm_tools suite ignores catalog tests (the dedicated job above runs them)
- `.gitignore` (`*.duckdb`), `docs/esm_catalog_index.rst` stub

### Notes for review
- The `catalog` deps are kept behind an optional extra so they don't touch the base install.
- A pre-existing malformed `[tool.pytest]` (dot) section in `setup.cfg` is left untouched (worth a separate cleanup; pytest already ignores it).

### Context
Part of the WP1 catalogue decomposition. Full plan: [decomposition design](https://github.com/esm-tools/esm_tools/blob/esm-tools-plus/simcat/collapsed-collections/docs/superpowers/specs/2026-06-13-esm-catalog-pr-decomposition-design.md). Next: **PR-A1a** (STAC data-model foundation), stacked on this branch.


.github/workflows/esm-catalog-tests.yml
```diff
@@ -0,0 +1,37 @@
...
+          python -m pip install --upgrade pip
+          # Install esm_catalog in isolation. A full ".[catalog]" pulls the entire
+          # esm-tools base, which pins ruamel.yaml.clib==0.2.7 — it fails to build
+          # on Python 3.12. The catalog is a standalone component, so we install it
+          # with --no-deps and add only the deps it actually imports (plus a modern
+          # pytest; the base pins pytest==7.1.2, which predates 3.12 support).
+          pip install -e . --no-deps
+          pip install -U "pytest>=7.4" click loguru pystac shapely universal-pathlib
+      - name: Run esm_catalog tests
+        run: pytest src/esm_catalog/tests -v
```
mandresm: I think this all belongs to `tests/esm_catalog` like the other pytests. Consequently, this whole file can be removed.

src/esm_catalog/tests/__init__.py
mandresm, line 1: can be removed, see my comment above

src/esm_catalog/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""esm_catalog — STAC-based catalog for ESM-Tools experiment output."""
+
+from __future__ import annotations
+
+__version__ = "0.1.0"
```
mandresm, line 5: Our convention is to keep all packages in ESM-Tools under the same version number, and that's what happens when we bump a version number with the GitHub Actions. Nothing to be done here on your side @siligam, but just so you know, GitHub actions might change this number before merging takes place.

setup.cfg
```diff
@@ -75,3 +75,6 @@ max-line-length = 88
 [tool.pytest]
 ignore = "src/esm_runscripts/coupling/"
 norecursedirs = tests/helpers
+
```
mandresm:
```suggestion
```

This is indeed broken so you can remove it

setup.cfg
```diff
@@ -75,3 +75,6 @@ max-line-length = 88
...
+[tool:pytest]
+norecursedirs = tests/helpers src/esm_catalog
```
mandresm:
```suggestion
[tool:pytest]
norecursedirs = tests/helpers
```

If you follow my suggestion of bringing the esm_catalog/tests into the root `tests` dir then I don't think you need this

setup.py
```diff
@@ -85,6 +85,14 @@
...
             "esm_tools=esm_tools.cli:main",
             "esm_utilities=esm_utilities.cli:main",
+            "esm-catalog=esm_catalog.cli:main",
+        ],
+    },
+    extras_require={
+        "catalog": [
+            "pystac>=1.8",
+            "shapely>=2.0",
+            "universal-pathlib>=0.2",
```
mandresm, line 95: Are this a requirement for the catalogs? I would prefer moving the catalog requirements into the general requirements of the tool, because the catalogs are going to be a substantial part of the tool that everyone should be using.

mandresm (changes requested):
Hi @siligam,

Thanks for this comprehensive and short PR :)

Just a few suggestions

.github/workflows/esm-catalog-tests.yml
siligam: Moved tests to `tests/test_esm_catalog/` (matching the `tests/test_<package>/` convention used by esm_environment, esm_parser, esm_viz, etc.) and removed this dedicated workflow — the scaffold smoke tests only need `click`, already a base dependency, so the main pytest suite picks them up directly. Fixed in 7a910f588.

src/esm_catalog/tests/__init__.py
siligam: Done — moved to `tests/test_esm_catalog/__init__.py`. Fixed in 7a910f588.

src/esm_catalog/__init__.py
siligam: Noted, thanks for the context.

setup.cfg
siligam: Good catch — that's pre-existing dead config (`[tool.pytest]` with a dot isn't a section pytest recognizes in setup.cfg; only `[tool:pytest]` is). Removed it and collapsed everything into one working `[tool:pytest]` section. Fixed in 7a910f588.

setup.cfg
siligam: Done — with tests moved into `tests/test_esm_catalog/` the `src/esm_catalog` exclusion is no longer needed. `setup.cfg` now has a single `[tool:pytest]` section with just `norecursedirs = tests/helpers`. Fixed in 7a910f588.

setup.py
siligam: Yes, these are needed for the STAC layer landing in the next PRs in this stack. For now I'd like to keep them under the `catalog` extra rather than folding them into base `install_requires`: the catalog package is still mid-rollout across a stack of PRs (this is PR-0 of ~13), so I'd rather not add pystac/shapely/universal-pathlib to every existing esm-tools install until the feature is complete and stable. Happy to revisit moving them into the base requirements once the stack lands and the catalog is something every user is actually expected to use.

.github/workflows/esm-catalog-tests.yml
siligam: Correction to my last reply: on closer look, this workflow is genuinely load-bearing for the rest of this PR stack, not redundant scaffolding. Six of the follow-on PRs extend this exact dependency-install line as the catalog grows (xarray/netCDF4 for scanning, cfgrib, duckdb, fastapi, pytz, ...) — it's the only thing running esm_catalog tests in CI for the whole stack, since base `install_requires` pins `ruamel.yaml.clib==0.2.7` (fails to build on 3.12) and the main esm-tools CI matrix only covers 3.8-3.11. So I've restored it in 83df3be93, repointed at `tests/test_esm_catalog` (the relocated path) rather than removing it outright.

setup.cfg
siligam: Correction to my last reply here too: `tests/test_esm_catalog` still needs to stay in `norecursedirs` (just relocated from `src/esm_catalog`) — downstream PRs in this stack add tests that import pystac/xarray/etc., which the main suite's environment doesn't install, so the main `pytest -v .` run needs to keep skipping that directory. Restored in 83df3be93.

siligam: One more correction here, found while watching CI roll through the rest of the PR stack: `src/esm_catalog` itself also needs to stay in `norecursedirs`, not just the relocated test dir. The main esm-tools workflow runs `pytest --doctest-modules src`, which executes doctests embedded in source-file docstrings — `src/esm_catalog/uri.py` has one that imports `universal-pathlib` (an extras-only dep), and dropping the `src/esm_catalog` exclusion broke that job starting at the next PR in the stack. Final state: `norecursedirs = tests/helpers src/esm_catalog tests/test_esm_catalog`.

siligam:
Summary of how the review feedback above landed, since the back-and-forth in the threads (a couple of which I had to self-correct) is easier to follow as one note:

**Tests relocated** — moved from `src/esm_catalog/tests/` to `tests/test_esm_catalog/`, matching this repo's `tests/test_<package>/` convention (`test_esm_environment`, `test_esm_parser`, `test_esm_viz`, ...).

**Dedicated `esm-catalog-tests.yml` workflow — kept, repointed at the new path.** Initially removed it per the suggestion above, since PR-0's own smoke tests don't need it. But while rebasing the rest of this PR stack onto the fix, I found 6 of the 12 follow-on PRs extend this exact workflow's dependency-install line as the catalog grows (xarray/netCDF4, cfgrib, duckdb, the FastAPI stack, pytz...) — it's the only thing running catalog tests in CI for the whole stack, since base `install_requires` pins `ruamel.yaml.clib==0.2.7` (fails to build on 3.12) and the main CI matrix only covers 3.8–3.11. Restored it.

**`setup.cfg` — collapsed the dead `[tool.pytest]` (dot, not recognized by pytest) into one working `[tool:pytest]` section**, with `norecursedirs = tests/helpers src/esm_catalog tests/test_esm_catalog`. Both `src/esm_catalog` *and* the relocated test dir need to stay excluded from the main suite: the main esm-tools workflow runs `pytest --doctest-modules src`, which executes doctests embedded in source-file docstrings (`src/esm_catalog/uri.py` has one that needs `universal-pathlib`, an extras-only dep) — dropping that exclusion broke `basic_tests` for a bit.

**Catalog deps stay under the `catalog` extra** (not folded into base `install_requires`) — the catalog is still mid-rollout across this PR stack (~13 PRs), so I'd rather not add pystac/shapely/etc. to every existing esm-tools install until it's complete. Open to revisiting once the stack lands.

**Also fixed while rebasing:** a pre-existing bug surfaced on Python 3.9 — `src/esm_catalog/api/responses.py` (added in PR-B1, #1490) uses `X | None` union syntax that pydantic v2 can't resolve on Python <3.10 without `eval_type_backport`. Latent since B1, it only started failing CI once PR-B3 wired it into the import chain. Added `eval_type_backport` as a dependency (pydantic's own recommended fix) rather than rewriting the ~34 files that use this syntax.

All 13 PRs in the stack (#1483–#1496) have been rebased onto the corrected base and are green: 162 catalog tests passing on both Python 3.9 and 3.12, plus the full main esm-tools suite.

mandresm:
Thanks for the summary. I'm going to have a look at it now.

.github/workflows/esm-catalog-tests.yml
mandresm: I see. Okay, then let's keep them separated until esm-tools build with python 3.12.

setup.cfg
```diff
@@ -72,6 +72,5 @@ max-line-length = 88
 
 [aliases]
 
-[tool.pytest]
-ignore = "src/esm_runscripts/coupling/"
-norecursedirs = tests/helpers
+[tool:pytest]
+norecursedirs = tests/helpers src/esm_catalog tests/test_esm_catalog
```
mandresm, line 77:
```suggestion
addopts = --ignore=src/esm_runscripts/coupling/
norecursedirs = tests/helpers src/esm_catalog tests/test_esm_catalog
```

setup.cfg
mandresm: Makes sense to me. Again, something that can change once these dependencies become part of the base `install_requirements`.

setup.py
mandresm: I think that is a very well informed decission

mandresm (approved):
Only one minor suggestion. Thanks for the changes and the clarifications!

mandresm:
Hi @siligam, this is ready to be merged, but before merging it needs bumping and tests to finish. We have a bumping workflow that you'll see me trigger below.

I am going to bump the version as a patch. It is not a patch but is not either a full feature so that's why I'll be using patch.

mandresm:
#bump #patch

PR #1500, opened 2026-06-19
Added echam6 tracer file to output

TronjeKemena:


seb-wahl (approved):

seb-wahl:
I will test this manually. To be consistent, the other namelist templates should modified as well. I'll do so once I don't see any issues with standard FOCI (don't know if it is actually still used in any project)

seb-wahl:
manually merged into geomar

PR #1501, opened 2026-06-19
Feat/albedo

mandresm:
A PR for AWIESM-2 in Albedo:
- It brings back esm_tests for Albedo for AWIESM
- Fixes some issues with AWIESM-2.6.2
- Provides support for more modern versions of OpenMPI and the libraries that depend on it can (it uses new built modules). However, the default is still the old modules.
  To override the default to use modules during compile time one needs to define `computer.mpi_implementation: openmpi_2026` in a yaml file and pass it to esm_master (see https://esm-tools--1482.org.readthedocs.build/en/1482/esm_master.html#configuring-esm-master-for-compile-time-overrides), for example:

  openmpi_2026.yaml
  ```yaml
  computer:
      mpi_implementation: openmpi_2026  
  ```
  Then you can run `esm_master install-awiesm-2.6.2 --modify-config openmpi_2026.yaml` and it will use the openmpi_2026 modules.
  
  For esm_runscripts you just need to add the same lines to the runscript itself.
  
  I am working on a feature that will enable defining variables for esm_master in the command line (e.g.: `esm_master install-awiesm-2.6.2 computer.mpi_implementation=openmpi_2026`), but this will come in a different PR after this is merged.
  
@ackerlar and @christian-stepanek, I'll need a review and approval. It might also be good to announce this once it is merge to the whole Paleo group. It is possible that the new modules solve problems with ucx/pmi, so it might be worth a try if people are having problems. However, I am not sure if swapping modules will keep results bit reproducible with respect the old modules. That's why I decided not to enforce the new modules. @christian-stepanek has offered help to test how things change so that Paleo can decide (later) if you want to have the new modules as the default.

It is very important that if you use the new modules in the esm_runscripts you also use binaries compiled with the new modules and not the old ones.


configs/machines/albedo.yaml
```diff
@@ -66,29 +66,70 @@ computer:
         '30min' if ${compute_time_minutes} <= 30 else
             '12h' if ${compute_time_minutes} <= 720 else '48h'
         ))"
-    additional_flags: ["--qos=${qos}"]
+    additional_flags: ["--qos=${qos}", "--comment='no-switch-enforce'"]
```
mandresm, line 69: This is to reduce queueing time in Albedo, although then the allocated nodes can communicate a bit slower since they are more spread across the cluster. But I think this should be okay.

ackerlar (approved):

configs/machines/albedo.yaml
```diff
@@ -66,29 +66,70 @@ computer:
...
 
-    mpi_implementation: openmpi
+    load_python: False
+    choose_computer.load_python:
+        True:
+            load_python_command: "load python/3.9.12"
+        False:
+            load_python_command: []
```
pgierz, line 76: Why once str and once list? Shouldn't the False case be empty string?

.github/workflows/esm_tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
...
   #   with:
   #       computer: blogin1
-  # albedo:
-  #   uses: ./.github/workflows/esm_tests_computer_workflow.yml
-  #   with:
-  #       computer: albedo1
+  albedo:
+    uses: ./.github/workflows/esm_tests_computer_workflow.yml
+    with:
+        computer: albedo1
```
pgierz, line 33: Hooray! 🥳

configs/machines/albedo.yaml
```diff
@@ -144,12 +173,13 @@ computer:
...
         - "load ${netcdfcxx_module}"
         - "load ${eccodes_module}"
+        - "load ${libaec_module}"
         # General Tools:
         - "load cdo/2.0.5"
         - "load nco/5.0.1"
         - "load git/2.35.2"
         - "load perl/5.35.0-gcc12.1.0"
-        - "load python/3.9.12"
+        - ${computer.load_python_command}
```
pgierz, line 182: This is my concern from above. 

Instinctively, this expands to `["foo", "bar", "baz", []]` if you have empty list case...?

Clearly this is not actually the case, or it wouldn't work, so the parser handles it well, but it looks counterintuitive, at least to me. What do you think?

pgierz (commented):
Looks good and can go in as is, but I have some questions/comments first before I push the green "yes can go in".

configs/machines/albedo.yaml
mandresm: `add_` extends if an element is a `list`. Not great.

I say if you have time to improve it (and it is a high priority for you to improve it) then it needs is own PR, as this requires changing the parser and running `esm_tests` on the HPC.

For now I am merging it as it is.

mandresm:
#bump

PR #1504, opened 2026-06-29
feat(esm_catalog): PR-A1b — core STAC infrastructure (flat layout)

mandresm:
Hi @siligam and @pgierz,

I thought I'll give it a try to breaking down even more #1484 into smaller units that I could review. While on the process of doing that using claude-code, I also added my suggestions. This might be more time-efficient than just writing what I think and @siligam implementing it. I have included you both as co-authors. Since the code is rewritten by claude-code and myself this is the only way I found to keep you as contributors. Let me know if you are okay with this.

It would be great if one of you could review it.

This PR's scope is to keep add minimal core functionality to build the dictionary for STAC. These includes the following changes:

- ESMItem(dict) and ESMCollection(dict) — STAC Item and Collection as plain-dict subclasses; JSON-serialisable with no extra dependencies
- context.py — CollectionContext dataclass (breaks scan<->stac cycle)
- uri.py — parse_uri / to_uri helpers for local and remote paths
- registry.py — EXTENSION_URLS dict; custom schemas use esm-tools.github.io placeholder URLs until schemas land in configs/stac-extensions/
- Flat layout: no stac/ or extensions/ subdirectories
- Core item properties only: variable, datetime, format, output_frequency, variables list (fields 1, 3, 5, 8, 10 from the metadata reference)
- 12 tests, all passing

## TODO
- [x] Cleanup
- [x] Docstrings
- [x] Black & isort


src/esm_catalog/item.py
```diff
@@ -0,0 +1,130 @@
...
+        """Build the STAC Item from a file path, scan metadata, and collection context.
+
+        Args:
+            path:     Path to the source file (local Path, UPath, or URI string).
+            metadata: Dict returned by scan_netcdf() or scan_grib().
+            ctx:      CollectionContext with experiment_id, component, collection_id.
+        """
+        self.path = self._resolve_path(path)
+        self.metadata = metadata
+        self.ctx = ctx
```
siligam: saving `path`, `metadata` and `ctx` as instance variables may lead to dual representation of data. For instance consider the value of `variable` is set to `foo` like this `item['properties']['variable'] = 'foo'` this does not change the original value in `self.metadata['variable']` making it unreliable to refer to get to the  value of `variable` set on `ESMItem`.

src/esm_catalog/uri.py
```diff
@@ -0,0 +1,72 @@
...
+            host = path.fs.host
+            path_part = path.path if hasattr(path, "path") else str(path)
+            if host:
+                return f"{protocol}://{host}{path_part}"
+        # Fallback to str() representation
+        uri = str(path)
+        # Fix malformed URIs like ssh:///path (missing host)
+        if uri.startswith(f"{protocol}:///") and not uri.startswith(f"{protocol}://localhost"):
+            # Can't recover host here, just return as-is
+            pass
```
siligam: here silently passing instead of raising error means the malformed path is used in the Assets. Hmm...  by the way `path.as_uri()` handles this correctly. In-fact, UPath already handles both local and remote paths. why not use it directly instead of this module?

src/esm_catalog/collection.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(self, ctx):
+        super().__init__({
+            "type": "Collection",
+            "id": ctx.collection_id,
+            "stac_version": "1.0.0",
+            "stac_extensions": [],
+            "title": ctx.experiment_id,
+            "description": f"All model output for experiment {ctx.experiment_id}",
+            "license": "proprietary",
```
siligam: is this okay to hardcode the type of `license` or should that come from `ctx`

src/esm_catalog/item.py
mandresm: That's a very good point. At this point I see no benefit of making it a class so I will revert it to functions, as you had it originally.

mandresm: Done!

@siligam was suggesting on a meeitng to use https://pystac.readthedocs.io/en/stable/api/item.html to build the stac item directly. I am working on that already

mandresm: @siligam, see this changes: [change make_item to return a pystac Item, instead of just a dictionary](https://github.com/esm-tools/esm_tools/pull/1504/commits/846c0b80b10bbf64695d6cfb8b6a85a5c894d5b1)

For the fun, this time I have avoided using claude to implement this change, in the hope that I can learn something from pystac.

mandresm: `type`, `stac_version` and `links` I have kept them in `extra_fields`. That's maybe wrong. Do they belong to one of the extensions?

src/esm_catalog/uri.py
mandresm: I don't know the answers to those questions. I think this line comes from the original PR:

https://github.com/esm-tools/esm_tools/blob/22519379dc96e17e65d8d302c6d969375f0ec139/src/esm_catalog/uri.py#L91-L93

> In-fact, UPath already handles both local and remote paths. why not use it directly instead of this module?

Isn't UPath an option? That was my understanding from looking at the code. I don't know when `uri.py` was introduced, I cannot see any `uri` on the original monolithic PR #1473. But I agree, I think we could just kill `uri.py` and substitute it by `upath = UPath(path)` and `upath.as_uri()`. Please, let me know if you agree with that and then I can take care of it.

src/esm_catalog/collection.py
mandresm: I think it would be better to get it from `ctx`. I can take care of that.

Do you think it is worth avoid subclassing the dict to create ESMCollection and instead just having a building function that returns a PySTACCollection a you suggested for the items?

src/esm_catalog/item.py
siligam: `type` and `stac_version` are managed by pystac automatically — `.to_dict()` injects them, no need to set them anywhere. `links` should go through `item.add_link(...)` not `extra_fields`.

`extra_fields` is only for custom top-level JSON fields pystac does not natively know about (e.g. if we ever want `"experiment"` at the top level of the item JSON). Putting standard STAC fields there means they would appear twice in the serialised output.

```python
item.add_link(
    pystac.Link(rel="collection", target=f"#{ctx.collection_id}",
                media_type="application/json")
)
# type, stac_version — do not set, pystac handles them
```

siligam: also in `test_item_single_time_sets_datetime`:

```python
assert "start_datetime" not in dir(item)
```

`dir(item)` lists all attributes and methods on the object — `start_datetime` would not appear there regardless of whether it is set or not, so this always passes and proves nothing. Better:

```python
assert item.common_metadata.start_datetime is None
```

or if you want to verify the serialised output:

```python
assert "start_datetime" not in item.to_dict()["properties"]
```

src/esm_catalog/uri.py
siligam: yes, kill `uri.py`. Both `Path` and `UPath` have `.as_uri()` natively, so `_to_href` simplifies to:

```python
def _to_href(path) -> str:
    return Path(path).resolve().as_uri() if not hasattr(path, "protocol") else path.as_uri()
```

or even just `path.as_uri()` directly if the path coming in is always already a `UPath` or absolute `Path`. the `.resolve()` is just a safety net for local paths.

src/esm_catalog/collection.py
siligam: yes, `pystac.Collection` can be used directly. sketched it out here if it useful: https://gist.github.com/siligam/93a82e11252e81e2b988a4a8df9edcf9

src/esm_catalog/uri.py
mandresm: I have to keep some logic for `_to_href` because `as_uri` kills the host for ssh and ftp, but I can definitely do that inside `_to_href`, no need for the `uri.py`. I'll be commiting that soon.

src/esm_catalog/collection.py
mandresm: I'll give it a go!

src/esm_catalog/item.py
mandresm: > type and stac_version are managed by pystac automatically — .to_dict() injects them, no need to set them anywhere. links should go through item.add_link(...) not extra_fields.

I'm not using `.to_dict` but just initializing the object.

About the `start_datetime` as am attribute, good catch! However, in that test (with `pytest tests/test_esm_catalog -k sets_datetime --trace`) I am seeing that:

```python
>>> item.common_metadata.start_datetime
datetime.datetime(2000, 1, 1, 0, 0, tzinfo=tzutc())
```

and the same for `end_datetime`:

```Python
>>> item.common_metadata.start_datetime
datetime.datetime(2000, 1, 1, 0, 0, tzinfo=tzutc())
```

I have the impression that during instancing the class sees that `start/end_datetime` are None and fills them up with whatever it is defined in `datetime`.

mandresm: > type and stac_version are managed by pystac automatically — .to_dict() injects them, no need to set them anywhere. links should go through item.add_link(...) not extra_fields.

I have removed `type`, `stac_version` and `links` from `extra_fields`. The link is now with set with `item.add_link`. `type` and `stac_version` are set automatically.

> dir(item) lists all attributes and methods on the object — start_datetime would not appear there regardless of whether it is set or not, so this always passes and proves nothing.

This is also sorted out. I'm now resolving this thread.

src/esm_catalog/collection.py
```diff
@@ -0,0 +1,82 @@
...
+            "extent": {
+                "spatial": {"bbox": [[-180.0, -90.0, 180.0, 90.0]]},
+                "temporal": {"interval": [[None, None]]},
+            },
+            "links": [
+                {"rel": "parent", "href": f"#{ctx.experiment_id}", "type": "application/json"},
+            ],
+            "experiment": ctx.experiment_id,
+            "components": [ctx.component],
```
mandresm: @siligam, what are these keys used for? They do not seem to belong to any pystac.Collection attribute or to the metadata so I don't know where to place them. Any ideas?

mandresm: If I take this dictionary and I pass it to `pystac.Collection.from_dict()` then `experiment` and `components` end up in `col.extra_fields`.

src/esm_catalog/collection.py
```diff
@@ -0,0 +1,57 @@
...
+
+    Args:
+        ctx: CollectionContext with experiment_id, component, collection_id.
+
+    Returns:
+        A pystac.Collection with default global extent.
+    """
+    col = Collection(
+        id=ctx.collection_id,
+        description=f"All model output for experiment {ctx.experiment_id}",
```
mandresm: This should probably come from the context. The users are going to define this in their `runscript.yaml` for example in:

```yaml
general:
    description: "my cool experiment"
```

`description` should also be a mandatory field if the simulation is labelled as a `production` simulation. This should probably be handled by the Context class. There are going to be other mandatory fields for production simulations.

What do you think @siligam?


```suggestion
        description=ctx.description,
```

src/esm_catalog/context.py
```diff
@@ -0,0 +1,15 @@
...
+
+@dataclass
+class CollectionContext:
+    """Identity + pre-scanned context for building a collection's items."""
+
+    experiment_id: str
+    component: str
+    collection_id: str
+    experiment_path: Path | None = None
+    namelists_by_component: dict = field(default_factory=dict)
```
mandresm, line 24:
```suggestion
    namelists_by_component: dict = field(default_factory=dict)
    description: str
```

Make the description mandatory (not None or empry) if production true.

mandresm (commented):
This is a self review with some changes needed for the code.

src/esm_catalog/collection.py
mandresm: Done!

src/esm_catalog/collection.py
siligam: yup, `extra_fields` is the right place as they are not native STAC Collection fields.

src/esm_catalog/collection.py
siligam: Agree it should come from `ctx.description`.  On making it mandatory field for production, so I guess production field will be created and all relevant fields are added to that dataclass.

mandresm: You are right that the production field would need to be created, but all the relevant fields need to be present in the data class anyway, independently of whether they have a default or an empty value, right?

What I would have is a checker method that checks "if production" then make sure certain attributes don't have an empty/default/None value.

That do you think? Does such a checker method has its place in the `CollectionContext` class?

mandresm: I'm thinking of something like this:

```python
class CollectionContext:

    experiment_id: str
    component: str
    collection_id: str
    production: bool = False
    description: str = ""
    license: str = ""
    experiment_path: Path | None = None
    namelists_by_component: dict = field(default_factory=dict)

    PRODUCTION_REQUIRED: ClassVar[list[str]] = [
        "description",
        "license",
    ]

    def __post_init__(self):
        if self.production:
            self._check_production_fields()

    def _check_production_fields(self):
        missing = [f for f in self.PRODUCTION_REQUIRED if not getattr(self, f)]
        if missing:
            raise ValueError(
                f"Production context requires non-empty fields: {', '.join(missing)}"
            )
```

siligam: Super. one small thing `if not getattr(self, f)`, how is it handled when the value of `f` is literally `False`. In practice value of `f`  may not a falsy value. how about checking `if getattr(self, f)  in (None, "")`. Other than that every thing looks cool.

mandresm: Good point! False could be a correct value for some future attribute, so we need to fence ourselves against that. I'll include this suggestion then.

siligam (approved):

mandresm:
#bump #patch

PR #1505, opened 2026-06-30
enable esm_master cli config options, to overrid config yamls

mandresm:
Allows to run `esm_master <command>-<model>-<version> section.key=value` to override yaml file configurations.

This is particularly useful for switched like `computer.mpi_implementation`.

This is not needed in runscripts as the user has to provide a runscript.yaml anyway.

# TODOs

- [x] isort and black
- [x] docs
- [x] docstrings

### DIFF
--- a/docs/esm_master.rst
+++ b/docs/esm_master.rst
@@ -82,3 +82,37 @@ It is possible that some models have special compile-time settings that need to 
                     SPECIAL_COMPILE_VAR: "compile_value"        
 
 See :ref:`esm_environment:ESM Environment` for more details on environment configuration options.
+
+CLI Configuration Overrides
+----------------------------
+
+``esm_master`` supports Spack-style ``key=value`` overrides directly on the command line, so you can
+change any configuration value without creating a YAML file::
+
+    $ esm_master install-awiesm-2.6.2 computer.mpi_implementation=openmpi_2026
+
+Keys use dot notation to address nested sections (``section.key``). Multiple overrides can be
+passed at once::
+
+    $ esm_master install-awiesm-2.6.2 computer.mpi_implementation=openmpi_2026 computer.fc=ifort
+
+Values are parsed as YAML scalars, so strings, integers, booleans, and ``null`` all work::
+
+    $ esm_master install-fesom-2.0 computer.omp_num_threads=4
+
+.. note::
+
+   Some configuration keys are themselves named with a dot (e.g. ``namelist.echam`` inside
+   ``add_namelist_changes``). In those cases, change the separator with ``--separator`` so the dot
+   in the key name is not interpreted as a level separator::
+
+      $ esm_master install-awiesm-2.6.2 --separator , general,some.dotted.key=value
+
+The overrides are applied after the target/model config is loaded but **before** ``choose_``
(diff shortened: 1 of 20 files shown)


siligam (commented):
Nice feature, and tagging the overrides with `command_line` provenance is the right way to do it. A couple of inline comments from testing, plus one thing I can't attach to a line: multi-cluster jobs re-invoke `esm_master` via subprocess without forwarding the overrides, so they're silently dropped there. Maybe just worth a warning for now.

docs/esm_master.rst
```diff
@@ -82,3 +82,37 @@ It is possible that some models have special compile-time settings that need to
...
+
+    $ esm_master install-fesom-2.0 computer.omp_num_threads=4
+
+.. note::
+
+   Some configuration keys are themselves named with a dot (e.g. ``namelist.echam`` inside
+   ``add_namelist_changes``). In those cases, change the separator with ``--separator`` so the dot
+   in the key name is not interpreted as a level separator::
+
+      $ esm_master install-awiesm-2.6.2 --separator , general,some.dotted.key=value
```
siligam, line 109: This exact command fails on Python <3.13 (which is all we support): argparse on ≤3.12 can't pick up positionals again after an interleaved optional, so it dies with `error: unrecognized arguments: general,some.dotted.key=value`. Putting `--separator ,` before the target works fine — I think just reordering this example (and the one in the `--help` text in cli.py) fixes it.

src/esm_parser/esm_parser.py
```diff
@@ -792,6 +792,84 @@ def dict_merge(dct, merge_dct, resolve_nested_adds=False, **kwargs):
...
+                f"Invalid override ``{override}``, expected the form "
+                "``key=value`` (e.g. "
+                "``computer.mpi_implementation=openmpi_2026``).",
+            )
+        key, value = override.split("=", 1)
+        target = result
+        path = key.split(separator)
+        for part in path[:-1]:
+            target = target.setdefault(part, {})
+        target[path[-1]] = yaml.safe_load(value)
```
siligam: Two inputs crash here with a raw traceback where a `user_error` would be nicer: a malformed yaml value (`computer.flags=[unclosed` → ParserError) and an override nesting under an earlier scalar one (`computer.fc=ifort computer.fc.flags=O2` → TypeError on the setdefault walk).

docs/esm_master.rst
mandresm: Sorted, I have fixed it using `parse_intermixed_args`

src/esm_parser/esm_parser.py
mandresm: I added both of them. For the second one I am also checking that key/values cannot be overwritten by different overrides:
- `computer.FC=ifort computer.FC=O2` is NOT allowed
- `computer.FC=ifort computer.FC.flags=O2` is NOT allowed
- `computer.FC.bin=ifort computer.FC.flags=O2` is allowed

siligam (approved):
Both issues from my earlier review are fixed (parse_intermixed_args + user_error on bad yaml/conflicting overrides). CI green, LGTM.

mandresm:
#bump

PR #1506, opened 2026-07-06
feat(esm_catalog): PR-A1g — contacts STAC extension

mandresm:
Adds feature `PR-A1g — Contacts extension` described in https://github.com/esm-tools/esm_tools/blob/22519379dc96e17e65d8d302c6d969375f0ec139/src/esm_catalog/PR-A1A-STAC-FOUNDATION-PLAN.md.

Adds contact info to the STAC item, and checks whether mandatory info is provided if the experiment is labelled as production.

## TODOs

- [x] isort and black
- [x] docstrings


src/esm_catalog/item.py
```diff
@@ -89,9 +95,22 @@ def _build_properties(metadata: dict, ctx) -> dict:
...
+        properties["contacts"] = [_contact_to_stac(c) for c in ctx.contacts]
+
     return properties
 
 
+def _contact_to_stac(contact) -> dict:
+    """Convert a Contact dataclass to STAC contacts extension format."""
+    entry: dict = {"name": contact.name, "roles": contact.roles}
+    if contact.orcid:
+        entry["identifier"] = {"scheme": "orcid", "identifier": contact.orcid}
```
siligam: The v0.1.1 schema we declare in `stac_extensions` types `identifier` as a plain string, not an object — I checked v1.0.0 too, still a string, so no released version of the extension accepts the `{scheme, identifier}` shape. Any item with an orcid fails validation against its own declared extension URL. Suggest the ORCID URI as the string, e.g. `entry["identifier"] = f"https://orcid.org/{contact.orcid}"`, and adjusting `test_item_contacts_in_properties` to match.

src/esm_catalog/context.py
```diff
@@ -5,6 +5,46 @@
...
+    def from_dict(cls, entry: dict) -> Contact:
+        """Build a Contact from a dict entry.
+
+        Expected keys: name, orcid, institution, roles.
+        """
+        return cls(
+            name=entry["name"],
+            orcid=entry.get("orcid"),
+            institution=entry.get("institution"),
+            roles=entry.get("roles", ["principal_investigator"]),
```
siligam, line 36: Two small robustness things in `from_dict`: `.get("roles", [...])` only falls back when the key is *absent* — a YAML entry with a bare `roles:` gives `None`, which flows through to `"roles": null` in the item (schema wants an array). And `entry["name"]` above raises a bare `KeyError` instead of the `ValueError` style `validate()` uses — opaque for config authors once the loader lands. Building kwargs only from present/non-null keys (and raising `ValueError` for a missing name) fixes both, and also removes this second copy of the default roles list that can drift from the field default above.

src/esm_catalog/item.py
```diff
@@ -89,9 +95,22 @@ def _build_properties(metadata: dict, ctx) -> dict:
...
 
+    if ctx.contacts:
+        properties["contacts"] = [_contact_to_stac(c) for c in ctx.contacts]
+
     return properties
 
 
+def _contact_to_stac(contact) -> dict:
+    """Convert a Contact dataclass to STAC contacts extension format."""
+    entry: dict = {"name": contact.name, "roles": contact.roles}
```
siligam: `contact.roles` goes in by reference — pystac keeps the properties dict as-is, so every item built from the same ctx shares this one list object, and an in-place mutation anywhere propagates to all items and back into the `Contact`. `list(contact.roles)` is enough.

src/esm_catalog/item.py
```diff
@@ -37,6 +39,10 @@ def make_item(
         path,
     )
 
+    stac_extensions = []
+    if ctx.contacts:
```
siligam: Not a bug, but a pattern question before hpc/datacube/namelist/paleo copy this: the `if ctx.contacts:` guard now lives here and again in `_build_properties`, and the two have to stay in sync or an item declares an extension URL without its properties (or the reverse). With four more extensions coming, one seam per extension that pairs guard → URL + properties would make that drift impossible. Fine to keep this PR as is and align the pattern in A1c/A1d — just want us to pick it deliberately.

tests/test_esm_catalog/test_stac_item.py
```diff
@@ -103,3 +103,64 @@ def test_item_time_range_sets_interval(tmp_path):
     assert item.datetime is None
     assert item.properties["start_datetime"] == "2000-01-01T00:00:00Z"
     assert item.properties["end_datetime"] == "2000-12-31T00:00:00Z"
+
+
+# --- contacts ---
+
+def _ctx_with_contacts(*contacts):
```
siligam: Tiny: `Contact` is imported inline in five tests — could be one top-level import like `test_context.py` does. And `_ctx_with_contacts` duplicates `_ctx`'s hard-coded fields; a `contacts=` kwarg on `_ctx` would do the same job.

siligam (commented):
Nice and compact, tests included. One real conformance bug (the `identifier` shape) plus a few small robustness/cleanup things — details inline.

src/esm_catalog/item.py
siligam:
```suggestion
        entry["identifier"] = f"https://orcid.org/{contact.orcid}"
```

`test_item_contacts_in_properties` (test_stac_item.py:135) would need the matching update: `assert contacts[0]["identifier"] == "https://orcid.org/0000-0001-2345-6789"`.

src/esm_catalog/context.py
mandresm: > only falls back when the key is absent — a YAML entry with a bare roles: gives None, which flows through to "roles": null in the item (schema wants an array)

I'm okay with this. I think validation of the type is the business of the schema, not something that needs to be done at instancing time.

> And entry["name"] above raises a bare KeyError instead of the ValueError style validate() uses

I am correcting that by adding a get as well, and renamed the validation to `validate_production_req` so that is clear that this is a validation of production requirements.

src/esm_catalog/item.py
mandresm: You are right, a single function for these two steps is a much better approach. I have taken the contact properties out of the `_build_properties` and added a new function for contacts that both adds the contact properties and also the extension. We can follow this pattern also in the other features.

src/esm_catalog/context.py
siligam: Happy to leave shape-checking to the schema in principle — and it does reject it. The catch is nothing in the pipeline or tests ever calls `.validate()` (grepped `src/esm_catalog` — zero `.validate()`/`validate_all`), and `to_dict`/save don't validate on their own. So today a bare `roles:` → `None` → `"roles": null` just gets written, unchecked.

Runnable repro (pystac 1.14.3): https://gist.github.com/siligam/58cad44b97c831c2f41d22d681c3ddbd

```
to_dict(): OK  -> no validation triggered, null is written
validate(): None is not of type 'array'
```

`to_dict()` writes the null silently; `validate()` catches it, but only when something actually calls it.

Two ways to close it, either's fine by me: the `.get` fallback in `from_dict` (build kwargs only from present/non-null keys so a bare `roles:` falls back to the default), or add a `.validate()` step somewhere in the write path / CI so the schema actually gets consulted. Just flagging that as it stands the null ships silently.

src/esm_catalog/context.py
```diff
@@ -5,6 +5,46 @@
...
+        Expected keys: name, orcid, institution, roles.
+        """
+        return cls(
+            name=entry.get("name"),
+            orcid=entry.get("orcid"),
+            institution=entry.get("institution"),
+            roles=entry.get("roles", ["principal_investigator"]),
+        )
+
+    def validate_production_req(self) -> None:
```
pgierz, line 39: Why isn't this happening during object instantiation?

src/esm_catalog/context.py
```diff
@@ -5,6 +5,46 @@
 from typing import ClassVar
 
 
+@dataclass
```
pgierz: This should probably be frozen:


```suggestion
@dataclass(frozen=True)
```

src/esm_catalog/item.py
```diff
@@ -96,6 +97,29 @@ def _build_properties(metadata: dict, ctx) -> dict:
...
+
+
+def add_contacts(item, ctx) -> None:
+    """Inject contacts extension URL and properties into *item*."""
+    contacts = getattr(ctx, "contacts", [])
+    if not contacts:
+        return
+
+    item.properties["contacts"] = [_contact_to_stac(c) for c in contacts]
+    item.stac_extensions.append(EXTENSION_URLS["contacts"])
```
pgierz: Ahhh! 😱In-Function argument mutation that then magically changes the object in outer scope?

tests/test_esm_catalog/test_context.py
```diff
@@ -61,3 +61,83 @@ def test_production_empty_string_counts_as_missing():
...
+
+
+def test_contact_validate_fails_with_empty_name():
+    with pytest.raises(ValueError, match="name"):
+        Contact(name="", institution="AWI").validate_production_req()
+
+
+def test_contact_validate_fails_with_none_name():
+    with pytest.raises(ValueError, match="name"):
+        Contact(name=None, institution="AWI").validate_production_req()
```
pgierz, line 109: Why do you have a validator method as part of the object? That feels off, somehow...

tests/test_esm_catalog/test_context.py
```diff
@@ -61,3 +61,83 @@ def test_production_empty_string_counts_as_missing():
...
+
+def test_production_context_passes_with_valid_contact():
+    ctx = CollectionContext(
+        experiment_id="exp",
+        component="echam",
+        collection_id="exp",
+        production=True,
+        description="desc",
+        data_license="CC-BY-4.0",
+        contacts=[Contact(name="Jane", institution="AWI")],
```
pgierz: It night be easier to have a fixture for this has a dictionary), pop off the key you don't want and then test object init,

pgierz (commented):
Some suggestions below

src/esm_catalog/item.py
siligam: Fair on the validation timing point, but the mutation critique seems overstated — `add_contacts(item, ctx)` mutates `item` in place, but that's exactly what the name and signature say it does, not hidden magic. It's also the existing convention across this codebase (`_build_properties`, and the paleo extension's `add_paleo_extension`/`add_experiment_type` in `paleo.py` do the same). If we want to move away from mutate-in-place helpers, that's a bigger refactor across `item.py`/`paleo.py`, not just this one function — worth deciding that explicitly rather than as a one-off nit here.

src/esm_catalog/context.py
siligam: Deliberate, not an oversight: Contact doesn't carry its own production/dev flag — only the enclosing CollectionContext does. If Contact validated eagerly in its own __post_init__, institution would become mandatory even for dev-mode contacts, which breaks the class docstring's own contract ("optional in dev, required in production"). CollectionContext already validates conditionally right after construction (see its __post_init__); Contact's validation is deliberately deferred until it's actually needed, since Contact itself has no way to know which context it's being validated for.

src/esm_catalog/context.py
siligam: Fair point — checked, and nothing currently mutates a Contact instance after construction (item.py only reads name/orcid/institution/roles). Leaving it mutable for now to keep this diff focused rather than bundling in a dataclass-semantics change; happy to pick it up separately if you'd rather lock it down now.

tests/test_esm_catalog/test_context.py
siligam: Same answer as the context.py:39 thread — Contact has no production/dev flag of its own, so it can't self-validate at construction time; that's why validate_production_req exists as a separate method called by CollectionContext once it knows the context is production.

tests/test_esm_catalog/test_context.py
siligam: Agreed, test_production_context_requires_description/test_production_missing_both_fields_names_both/test_production_empty_string_counts_as_missing do repeat the full CollectionContext(...) call almost verbatim. Not restructuring the tests in this PR to keep the diff focused, but a dict-fixture + pop-key pattern would clean this up nicely as a follow-up.

src/esm_catalog/context.py
mandresm: I agree with Pavan, I think this is okay as it is. We are saying "Contact dataclass can have missing values" but comes with a method that can check if the data is fit for production. `production` is defined in the `CollectionContext` (equivalent to things that are in the `config["general"]`) and that's IMO the place where the checking of production needs to happen. That's also where other checking for missing user input for runs labelled as production happens.

src/esm_catalog/context.py
mandresm: I have applied it though, we can change it back later. It's a minimal change, so the diffs won't go too crazy here.

tests/test_esm_catalog/test_context.py
mandresm: Changes applied!

src/esm_catalog/context.py
pgierz: OK, after discussion also with Miguel on Webex: this can stay as is.

tests/test_esm_catalog/test_context.py
pgierz: this discussion is deferred to a later, targeted diff where we only look at `Item.validate()` and `Collection.validate()` from the `pystac` side. After some discussion with @mandresm, the idea is to validate once and try to recycle the built-in validator.

src/esm_catalog/item.py
pgierz: OK, keep as is for now, we will look at this cleanly once all the ctx is finalized.

pgierz (approved):

mandresm:
#bump #patch

PR #1507, opened 2026-07-07
feat(esm_catalog): PR-A1d — datacube STAC extension

siligam:
Adds PR-A1d — Datacube extension from the [plan](https://github.com/esm-tools/esm_tools/blob/22519379dc96e17e65d8d302c6d969375f0ec139/src/esm_catalog/PR-A1A-STAC-FOUNDATION-PLAN.md): `cube:dimensions` and `cube:variables` on items when the scan metadata carries them.

Two notes. The extension is applied as one self-contained `add_datacube_extension()` called from `make_item` — the single-seam shape from the #1506 discussion, so guard, URL, and properties can't drift apart. And it turns out the v2.2.0 schema requires `cube:dimensions` whenever the extension is declared, so variables-only metadata is a no-op; there's a test validating a full item against a vendored copy of the schema to keep us honest (the CI workflow gets `jsonschema` added for that).

### DIFF
--- a/src/esm_catalog/item.py
+++ b/src/esm_catalog/item.py
@@ -10,6 +10,8 @@ from typing import Union
 from pystac import Asset, Item, Link
 from upath import UPath
 
+from esm_catalog.datacube import add_datacube_extension
+
 
 def make_item(
     path: Union[Path, UPath, str],
@@ -57,6 +59,8 @@ def make_item(
             media_type="application/json",
         )
     )
+
+    add_datacube_extension(item, metadata)
 
     return item
 
--- a/.github/workflows/esm-catalog-tests.yml
+++ b/.github/workflows/esm-catalog-tests.yml
@@ -34,6 +34,6 @@ jobs:
           # with --no-deps and add only the deps it actually imports (plus a modern
           # pytest; the base pins pytest==7.1.2, which predates 3.12 support).
           pip install -e . --no-deps
-          pip install -U "pytest>=7.4" click loguru pystac shapely universal-pathlib
+          pip install -U "pytest>=7.4" click loguru pystac shapely universal-pathlib jsonschema
       - name: Run esm_catalog tests
         run: pytest tests/test_esm_catalog -v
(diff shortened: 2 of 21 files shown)


src/esm_catalog/item.py
mandresm, line 40:
```suggestion
    )

stac_extensions = []
if metadata.get("dimensions"):
    stack_extensions.append(EXTENSION_URLS["datacube"])
else:
    logger.debug("Some message here")
```

I am making this suggestion probably because I don't understand something. But just checking with you @siligam: We are using `pystac.Item` because the instancing does some validation for us. I would assume that appending extensions to the pystac.Item would also add their own validations during instancing. Is this how it works? If that is the case, then it would make sense to check for which extensions we need before instancing the `pystac.Item` and then instancing it with all the extensions and metadata and let the metadata be evaluated. Does it make any sense or am I getting it completely wrong?

If I am right, I would also need to change the PR of the contacts to follow this pattern.

mandresm (commented):
Great job @siligam!

I have a design question, and maybe a suggestion, depending on whether I understand extensions or not.

src/esm_catalog/item.py
siligam: Checked this - `Item()` doesn't actually validate anything against the extension schemas at construction time, no matter when you set `stac_extensions` or the properties. Confirmed by building an Item with the datacube URL but no `cube:dimensions` set - constructs fine, no error. The only place pystac validates against a schema is `item.validate()`, which hits the network for the schema - that's actually why I added the vendored schema + test here, since nothing checks this for us otherwise.

So instancing-with-everything vs. adding after doesn't matter, no validation difference either way. Happy to keep `add_datacube_extension()` as is, and no need to change the contacts PR for this reason.

mandresm: Thanks! That makes sense. That's great because then we can validate later on if we need to.

mandresm (approved):

mandresm:
#bump #patch

PR #1508, opened 2026-07-07
test(esm_catalog): additional coverage for the #1504 core

siligam:
The extra test coverage for #1504 we agreed on in Webex. Covers make_item edge cases (string/URI path inputs, naive-datetime normalization to UTC, open-ended ranges, the missing-datetime contract, item-id format and path-hash collision guard, media-type selection, variables/output_frequency filtering), collection extent behavior (first-bbox sentinel replacement, min/max merging, monotonic temporal growth), and production-context validation edges. Tests only, no source changes — 17 → 40 core tests.

One observation from probing (not enshrined in a test): an open-ended time range currently emits `start_datetime` without `end_datetime`, which the STAC spec recommends against. Happy to file it separately if you think it's worth tightening.

### DIFF
--- a/tests/test_esm_catalog/test_context.py
+++ b/tests/test_esm_catalog/test_context.py
@@ -39,3 +39,25 @@ def test_production_context_requires_description():
             collection_id="exp-alpha",
             production=True,
         )
+
+
+def test_production_missing_both_fields_names_both():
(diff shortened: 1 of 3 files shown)


mandresm (approved):
Nice, thanks. I'll be merging now.

PR #1511, opened 2026-07-15
feat(esm_catalog): PR-A1c — HPC storage extension

siligam:
Part of the esm_catalog decomposition, per [the PR-A1a foundation plan](https://raw.githubusercontent.com/esm-tools/esm_tools/22519379dc96e17e65d8d302c6d969375f0ec139/src/esm_catalog/PR-A1A-STAC-FOUNDATION-PLAN.md). Follows #1504 (A1b core), branches from `release`.

Adds `src/esm_catalog/hpc.py`: `add_hpc_extension(item, path, machine_config)` sets `hpc:facility`, `hpc:system`, `hpc:storage_tier` (item properties) and `hpc:storage_type` (data asset) from a caller-supplied `machine_config` dict, plus `hpc:last_access` from `path.stat().st_atime`.

No path-pattern matching or filesystem probing — per the plan, the caller already knows which machine it's running on and passes that machine's config section down directly. `hpc:state`/`hpc:recall_time_estimate` (subprocess-based `dmattr`/`lfs` HSM queries) are dropped per plan.

`CollectionContext` gets a new optional `machine_config` field; `make_item` wires `add_hpc_extension` in unconditionally (no-op beyond `hpc:last_access` when `machine_config` is `None`, for graceful degradation).

Also adds `configs/stac-extensions/hpc/v1.0.0/schema.json` (custom extension — no external community schema exists for this one) and `tests/test_esm_catalog/test_hpc.py`.

## Test plan
- [x] `pytest tests/test_esm_catalog` — 50 passed
- [x] New `test_hpc.py` covers: no-op without config/stat, last_access-only degradation, facility/system/tier population, hot/warm/cold tier mapping, missing asset safety, idempotent URL append, `make_item` wiring, and schema validation via `jsonschema`


configs/stac-extensions/storage/v1.0.0/schema.json
mandresm, line 1: Should it be just simply storage station, instead of HPC storage extension, to be more general. I think that might be more accurate, since we might have data in the tape at some point. @pgierz, any opinions here? What about you @siligam ?

Sorry I did not think about this earlier...

tests/test_esm_catalog/schemas/hpc-v1.0.0.json
mandresm, line 1: Isn't this a duplication of `configs/stac-extensions/hpc/v1.0.0/schema.json`? Why not to use that file already for the testing?

src/esm_catalog/storage.py
```diff
@@ -0,0 +1,89 @@
...
+
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import TYPE_CHECKING, Optional, Union
+
+from esm_catalog.registry import EXTENSION_URLS
+
+if TYPE_CHECKING:
+    import pystac
+    from upath import UPath
```
mandresm:
```suggestion
import pystac
from upath import UPath
```

I think pystac and upath are always a requirement, aren't they?

src/esm_catalog/storage.py
```diff
@@ -0,0 +1,89 @@
...
+Fields hpc:state and hpc:recall_time_estimate (subprocess-based HSM state
+queries via dmattr/lfs) were dropped: they require machine-specific tools and
+add operational risk for marginal query value.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import TYPE_CHECKING, Optional, Union
```
mandresm:
```suggestion
from typing import Optional, Union
```

See comment below

mandresm (changes requested):

tests/test_esm_catalog/schemas/hpc-v1.0.0.json
pgierz: +1, same question.

configs/stac-extensions/storage/v1.0.0/schema.json
pgierz: Yes, storage is good.

src/esm_catalog/storage.py
pgierz: Are we ever **not** `TYPE_CHECKING`?

configs/stac-extensions/hpc/v1.0.0/schema.json
```diff
@@ -0,0 +1,72 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://esm-tools.github.io/stac-extensions/hpc/v1.0.0/schema.json",
+  "title": "HPC Storage Extension",
+  "description": "STAC HPC Storage Extension for STAC Items, describing where on an HPC facility a data file lives.",
```
pgierz: Needs correcting

configs/stac-extensions/storage/v1.0.0/schema.json
```diff
@@ -0,0 +1,72 @@
...
+  ],
+  "properties": {
+    "stac_extensions": {
+      "type": "array",
+      "contains": {
+        "const": "https://esm-tools.github.io/stac-extensions/hpc/v1.0.0/schema.json"
+      }
+    },
+    "type": {
+      "const": "Feature"
```
pgierz, line 18: Not sure I understand?

configs/stac-extensions/hpc/v1.0.0/schema.json
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+  },
+  "definitions": {
+    "fields": {
+      "type": "object",
+      "properties": {
+        "hpc:facility": {
+          "title": "Facility",
+          "description": "The HPC facility operating the system, e.g. 'AWI', 'DKRZ'.",
```
pgierz: This should use ROR

configs/stac-extensions/hpc/v1.0.0/schema.json
```diff
@@ -0,0 +1,72 @@
...
+          "enum": [
+            "hot",
+            "warm",
+            "cold"
+          ]
+        },
+        "hpc:storage_type": {
+          "title": "Storage type",
+          "description": "The filesystem or storage technology backing this file, e.g. 'lustre', 'gpfs', 'hpss'.",
+          "type": "string"
```
pgierz: To consider: en enum of allowed values?

src/esm_catalog/storage.py
```diff
@@ -0,0 +1,89 @@
...
+from esm_catalog.registry import EXTENSION_URLS
+
+if TYPE_CHECKING:
+    import pystac
+    from upath import UPath
+
+
+def add_hpc_extension(
+    item: "pystac.Item",
+    path: Union[Path, "UPath"],
```
pgierz: Why not just stick barely with u-path?

src/esm_catalog/storage.py
```diff
@@ -0,0 +1,89 @@
...
+    return datetime.fromtimestamp(stat.st_atime, tz=timezone.utc).isoformat()
+
+
+def _derive_tier(storage_type: str) -> str:
+    """Map a raw filesystem/storage type to a coarse hot/warm/cold tier."""
+    if storage_type in ("hpss", "dmf", "tape"):
+        return "cold"
+    if storage_type == "gpfs":
+        return "warm"
+    return "hot"
```
pgierz: This hard codes alot of assumptions

src/esm_catalog/hpc.py
```diff
@@ -0,0 +1,89 @@
...
+whatever machine the scan is running on — with (any of) these keys:
+    facility, system, storage_type
+
+Detection is not done by probing the filesystem or matching path patterns:
+the caller already knows which machine it is running on, so it resolves and
+passes in that machine's own config section directly.
+
+Fields hpc:state and hpc:recall_time_estimate (subprocess-based HSM state
+queries via dmattr/lfs) were dropped: they require machine-specific tools and
+add operational risk for marginal query value.
```
pgierz: Remove from doc string, this is not relevant.

src/esm_catalog/storage.py
```diff
@@ -0,0 +1,89 @@
...
+    if populated:
+        url = EXTENSION_URLS["hpc"]
+        if url not in item.stac_extensions:
+            item.stac_extensions.append(url)
+
+
+def _get_last_access(path) -> Optional[str]:
+    """Return path's last-access time as an ISO 8601 UTC string, or None."""
+    try:
+        stat = path.stat()
```
pgierz, line 94: Danger → this induces a filesystem operation. Do we want to do that?

src/esm_catalog/storage.py
```diff
@@ -0,0 +1,89 @@
...
+        item.properties["hpc:last_access"] = last_access
+        populated = True
+
+    if populated:
+        url = EXTENSION_URLS["hpc"]
+        if url not in item.stac_extensions:
+            item.stac_extensions.append(url)
+
+
+def _get_last_access(path) -> Optional[str]:
```
pgierz: Why not give back proper datetime type?

src/esm_catalog/hpc.py
```diff
@@ -0,0 +1,89 @@
...
+from typing import TYPE_CHECKING, Optional, Union
+
+from esm_catalog.registry import EXTENSION_URLS
+
+if TYPE_CHECKING:
+    import pystac
+    from upath import UPath
+
+
+def add_hpc_extension(
```
pgierz: This function needs to be reworked after we choose a name.

src/esm_catalog/item.py
```diff
@@ -10,7 +10,7 @@
 from pystac import Asset, Item, Link
 from upath import UPath
 
-from esm_catalog.datacube import add_datacube_extension
+from esm_catalog.datacube import add_datacube_extension, add_hpc_extension
```
pgierz: Is this import correct? Looks like a bad submodule ...

There does not seem to be something called datacube.py

tests/test_esm_catalog/schemas/hpc-v1.0.0.json
```diff
@@ -0,0 +1,72 @@
...
+          "title": "Last access time",
+          "description": "ISO 8601 UTC timestamp of the file's last access, as reported by the filesystem.",
+          "type": "string",
+          "format": "date-time"
+        }
+      },
+      "patternProperties": {
+        "^(?!hpc:)": {}
+      },
+      "additionalProperties": false
```
pgierz, line 69: Why false?

tests/test_esm_catalog/schemas/hpc-v1.0.0.json
```diff
@@ -0,0 +1,72 @@
...
+    }
+  },
+  "definitions": {
+    "fields": {
+      "type": "object",
+      "properties": {
+        "hpc:facility": {
+          "title": "Facility",
+          "description": "The HPC facility operating the system, e.g. 'AWI', 'DKRZ'.",
+          "type": "string"
```
pgierz: I would recommend "institution" and to cross link the ROR here

tests/test_esm_catalog/test_storage.py
```diff
@@ -0,0 +1,149 @@
...
+    item = Item(
+        id="x",
+        geometry=None,
+        bbox=None,
+        datetime=datetime(2000, 1, 1, tzinfo=timezone.utc),
+        properties={},
+    )
+    if with_asset:
+        item.add_asset("data", Asset(href="file:///tmp/x.nc"))
+    return item
```
pgierz: This belongs in a fixture

tests/test_esm_catalog/test_hpc.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+    if with_asset:
+        item.add_asset("data", Asset(href="file:///tmp/x.nc"))
+    return item
+
+
+def test_noop_without_machine_config_or_stat(tmp_path):
+    item = _bare_item()
+    missing = tmp_path / "does-not-exist.nc"
+    add_hpc_extension(item, missing, machine_config=None)
```
pgierz: Does it bother anyone else that this doesn't raise an error?

tests/test_esm_catalog/test_storage.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def test_machine_config_sets_facility_system_and_tier(tmp_path):
+    f = tmp_path / "temp.nc"
+    f.write_bytes(b"x")
+    item = _bare_item()
+    add_hpc_extension(
+        item,
+        f,
+        machine_config={"facility": "AWI", "system": "albedo", "storage_type": "lustre"},
```
pgierz: The hypothetical test could at least be correct. We don't have a lustre.

tests/test_esm_catalog/test_hpc.py
```diff
@@ -0,0 +1,149 @@
...
+import pytest
+from pystac import Asset, Item
+
+from esm_catalog.context import CollectionContext
+from esm_catalog.hpc import add_hpc_extension
+from esm_catalog.item import make_item
+from esm_catalog.registry import EXTENSION_URLS
+
+HPC_URL = EXTENSION_URLS["hpc"]
+SCHEMA_PATH = Path(__file__).parent / "schemas" / "hpc-v1.0.0.json"
```
pgierz: I think this is poorly designed. You have 2 copies of the schema, one just for testing, the other "real"

I a concerned about drift, just have one I would say.

pgierz (changes requested):
I think this still needs some comments and discussion.

src/esm_catalog/item.py
siligam: Already fixed on this branch's tip (52a73a1f2) — that was a merge artifact from stacking on the datacube PR. Current item.py imports `add_hpc_extension` from `esm_catalog.hpc` and no longer calls `add_datacube_extension`. Nothing left to do here.

siligam: Correction to my earlier reply — I was looking at a stale local branch ref. The bug was real: the release merge (which pulled in the merged datacube PR) collapsed both imports into `from esm_catalog.datacube import add_datacube_extension, add_hpc_extension`, and `add_hpc_extension` lives in hpc.py, not datacube.py — so item.py couldn't even be imported. Fixed in 84e80e1, split into two correct imports. Fixing it also exposed a stale assumption in test_stac_datacube.py (bare item expected no stac_extensions at all, which predates add_hpc_extension's intentional last_access-from-stat() behavior) — fixed alongside.

tests/test_esm_catalog/schemas/hpc-v1.0.0.json
siligam: Fixed in d58f61e — removed tests/test_esm_catalog/schemas/hpc-v1.0.0.json and pointed test_hpc.py's SCHEMA_PATH at configs/stac-extensions/hpc/v1.0.0/schema.json directly. Full suite green (60 passed).

configs/stac-extensions/storage/v1.0.0/schema.json
siligam: Done in 0b6af50 — renamed the extension to 'storage' across the board: hpc.py -> storage.py, add_hpc_extension -> add_storage_extension, EXTENSION_URLS key, schema dir configs/stac-extensions/hpc -> storage, and the hpc: field prefix -> storage:. Also dropped the storage_tier/storage_type stutter under the new prefix (-> storage:tier / storage:type) since STAC extension fields don't usually repeat the extension name. Full suite green (60 passed).

src/esm_catalog/storage.py
siligam: Fixed in a870540 — path.stat() is no longer called unconditionally. Added a probe_last_access: bool = False parameter to add_storage_extension (and a matching CollectionContext.probe_last_access field, default False), so storage:last_access is only populated when a caller explicitly opts in. Avoids the implicit filesystem/network op for remote or tape-backed paths by default. Full suite green (62 passed).

src/esm_catalog/storage.py
siligam: Fixed in e724d3e — _get_last_access now returns a real UTC datetime instead of an ISO string. Formatting to a string only happens at the point of insertion into item.properties, using pystac.utils.datetime_to_str for consistency with pystac's own RFC3339 'Z' format (rather than Python's default +00:00 offset). Full suite green (62 passed).

src/esm_catalog/storage.py
siligam: Fixed in 705b72a — the biggest assumption was defaulting anything outside the hpss/dmf/tape/gpfs whitelist to 'hot', which silently mislabeled any unrecognized filesystem as fast. _derive_tier now returns None for unrecognized/missing storage_type instead of guessing, and machine_config can pass storage_tier explicitly to override the heuristic for sites whose storage doesn't fit the built-in mapping. Added tests for the unrecognized-type and explicit-override paths. Full suite green (65 passed).

configs/stac-extensions/hpc/v1.0.0/schema.json
siligam: Fixed in e8126ba — the description no longer describes it as HPC-specific ('...where on an HPC facility a data file lives'), consistent with the extension rename to 'storage'. Now: 'STAC Storage Extension for STAC Items. Describes where a data file lives: institution, system, storage tier, and last-access time.'

configs/stac-extensions/hpc/v1.0.0/schema.json
siligam: Fixed in e8126ba — renamed storage:facility -> storage:institution (matching your other comment below) and its description now recommends a ROR identifier (https://ror.org/), e.g. https://ror.org/032e6b942 for AWI, falling back to a plain name when no ROR record exists.

configs/stac-extensions/storage/v1.0.0/schema.json
siligam: This is the standard STAC extension pattern: the top-level 'type' JSON-schema key at line 6 ("type": "object") describes the schema document itself, while the nested properties.type here constrains the STAC Item's own 'type' field to the literal 'Feature' (every STAC Item is a GeoJSON Feature). Same construct as tests/test_esm_catalog/schemas/datacube-v2.2.0.json's "type": {"const": "Feature"} already in this repo, and matches the official stac-extensions schemas. Happy to add a comment if that'd help, though JSON Schema itself doesn't support inline comments.

configs/stac-extensions/hpc/v1.0.0/schema.json
siligam: Left storage:type open-ended rather than adding a strict enum — filesystem/storage technologies vary a lot across facilities (ceph, s3, nfs, local disk, ...) and an enum would reject any site whose backend we didn't anticipate, which cuts against the same 'don't assume, let the site say what it knows' principle behind the recent _derive_tier fix (unrecognized storage_type now leaves storage:tier unset instead of guessing 'hot', rather than rejecting the value outright). storage:tier itself does have an enum (hot/warm/cold) since that's a small, closed, meaningful set. Open to adding one for storage:type too if you'd rather have a closed vocabulary with an 'other' escape hatch — let me know.

tests/test_esm_catalog/schemas/hpc-v1.0.0.json
siligam: This was on the duplicate test schema, which is now gone (we removed it and pointed tests at the canonical configs/stac-extensions/storage/v1.0.0/schema.json instead) — but the same additionalProperties: false lives there, so answering here: it's paired with patternProperties "^(?!storage:)": {} right above it. That negative-lookahead matches any key NOT starting with storage:, so arbitrary non-extension STAC properties are still allowed; additionalProperties: false only rejects storage:* keys that aren't in the explicit list above (i.e. a typo'd or made-up storage:foo field). Same pattern as the official stac-extensions schemas use to scope validation to just their own namespace.

tests/test_esm_catalog/test_storage.py
siligam: Fixed in 18aa30e — converted to a make_bare_item pytest fixture (a factory, since with_asset needs to vary per test).

tests/test_esm_catalog/test_storage.py
siligam: Fair — I can't actually verify what filesystem albedo runs, so pairing that real system name with a hypothetical storage_type='lustre' was an unverified claim I shouldn't have made. Renamed the example system to a fictional 'example-hpc' in 18aa30e; storage_type='lustre' is now clearly just an example technology, not an assertion about AWI's real infrastructure.

tests/test_esm_catalog/test_hpc.py
siligam: Yes, it did bother me too once I looked closer — fixed in 631f9be. except Exception was swallowing everything, so a missing file and a caller bug (wrong argument type, no .stat()) were indistinguishable, both silently producing no storage:last_access. Narrowed to except OSError, which still covers the legitimate cases (missing file, permission denied, a remote filesystem surfacing a network failure as OSError), but lets a real bug propagate. Added tests for both paths. Full suite green (67 passed).

src/esm_catalog/storage.py
siligam: Fixed in e7ab22c — switched to real imports (import pystac, from upath import UPath) since both are hard runtime dependencies already (item.py imports them unconditionally too), and dropped the TYPE_CHECKING guard along with the string-quoted forward refs.

src/esm_catalog/storage.py
siligam: Resolved in e7ab22c as part of moving pystac/upath to real imports — Union is no longer needed since path is now typed as plain UPath (see the comment on 'stick barely with u-path' below), so the import is just 'from typing import Optional'.

src/esm_catalog/storage.py
siligam: Fixed in e7ab22c — path is now typed as plain UPath (no more Union[Path, UPath]), and make_item normalizes any input (str, Path, or UPath) to a UPath before calling into any extension function, so storage.py never has to handle a bare pathlib.Path.

src/esm_catalog/hpc.py
siligam: Fixed in e7ab22c — removed that paragraph from the module docstring.

src/esm_catalog/hpc.py
siligam: Reworked across this review round: renamed to add_storage_extension (0b6af50), gated the filesystem stat() behind an explicit probe_last_access opt-in (a870540), returned a real datetime instead of a string (e724d3e), stopped defaulting unrecognized storage_type to 'hot' (705b72a), renamed facility -> institution with ROR guidance (e8126ba), narrowed the swallowed exception to OSError (631f9be), and just now switched to real pystac/upath imports with a UPath-only signature (e7ab22c). Should be in good shape now — flag anything that still needs another pass.

tests/test_esm_catalog/schemas/hpc-v1.0.0.json
siligam: Resolved along with the duplicate-schema removal — this test copy is gone (d58f61e), tests now load configs/stac-extensions/storage/v1.0.0/schema.json directly, which has the institution/ROR fix (e8126ba).

tests/test_esm_catalog/test_hpc.py
siligam: Agreed and fixed in d58f61e — removed the duplicate and pointed the tests at configs/stac-extensions/storage/v1.0.0/schema.json (the canonical one) directly.

siligam:
All review comments addressed with fixes (rename to `storage`, dropped duplicate test schema, real pystac/upath imports, `probe_last_access` opt-in instead of unconditional stat(), no more defaulting unrecognized storage_type to "hot", narrowed the swallowed exception, etc — see inline replies for commit refs). Two open questions (enum for `storage:type`, comment on the schema `type` field) got explanations rather than changes — happy to add either if you'd rather have them.

Also rebased onto current `release` to pick up #1506 — was conflicting with it in `context.py`/`item.py`. Tests still green (82 passed).

Ready for re-review.

tests/test_esm_catalog/test_storage.py
pgierz: > I can't actually verify what filesystem albedo runs

I will believe that @claude can't verify that, but @siligam you certainly can! Have you looked into our spaces documentation? https://spaces.awi.de/spaces/HELP/pages/426525152/System+overview+-+albedo+s+hardware#Systemoverviewalbedo'shardware-Filesystem

tests/test_esm_catalog/test_storage.py
```diff
@@ -0,0 +1,216 @@
...
+            geometry=None,
+            bbox=None,
+            datetime=datetime(2000, 1, 1, tzinfo=timezone.utc),
+            properties={},
+        )
+        if with_asset:
+            item.add_asset("data", Asset(href="file:///tmp/x.nc"))
+        return item
+
+    return _make
```
pgierz: This shows up in the other PR as well. This belongs in a common place

Two comments that are actually one: 
1 + 2. No factory. Just make two fixtures, that is clearer: `item` and `item_with_asset`

If you are concerned about code duplication, this should work:

```suggestion
def _base_item():
      return Item(
          id="x",
          geometry=None,
          bbox=None,
          datetime=datetime(2000, 1, 1, tzinfo=timezone.utc),
          properties={},
      )

@pytest.fixture
def item():
    return _base_item()

@pytest.fixture
def item_with_asset():
    item = _base_item()
    item.add_asset("data", Asset(href="file:///tmp/x.nc"))
    return item
```

One could have `item_with_asset(item)` and save one more line, but I am unsure if that mutates the fixture permanently. This is (one of) the reasons I am not so happy with mutation in our codebase (in general)

tests/test_esm_catalog/test_storage.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def test_noop_without_machine_config_or_probe(tmp_path, make_bare_item):
+    item = make_bare_item()
```
pgierz:
```suggestion
def test_noop_without_machine_config_or_probe(tmp_path, item):
```

tests/test_esm_catalog/test_storage.py
```diff
@@ -0,0 +1,216 @@
...
+    missing = tmp_path / "does-not-exist.nc"
+    add_storage_extension(item, missing, machine_config=None)
+    assert item.properties == {}
+    assert item.stac_extensions == []
+
+
+def test_missing_path_with_probe_omits_last_access_without_raising(tmp_path, make_bare_item):
+    item = make_bare_item()
+    missing = tmp_path / "does-not-exist.nc"
```
pgierz: same pattern here

src/esm_catalog/storage.py
```diff
@@ -0,0 +1,122 @@
...
+# not be silently assumed fast. Sites with a type outside this list, or a
+# type where the default doesn't hold, can set machine_config["storage_tier"]
+# explicitly to bypass this heuristic entirely.
+_TIER_BY_STORAGE_TYPE = {
+    "hpss": "cold",
+    "dmf": "cold",
+    "tape": "cold",
+    "gpfs": "warm",
+    "lustre": "hot",
+}
```
pgierz: This sort of information should be coming from the machine yamls, in my opinion? I would not have any kind of assumptions about the machines (and their path names) in the Python side. Let on in lines 70, you do some "derive_tier" stuff. That also looks wrong. The machine config should give you back a dict that can be used to pass in a path fragment and get out a storage tier Enum value. No magic strings, please.

tests/test_esm_catalog/test_storage.py
pgierz, line 1: many of these tests run against made-up machine configurations. I would also add a set of tests to run against our actual machine configurations (we have them all in the repo anyway).

pgierz (changes requested):
I'll take over from here @siligam, you can review my changes.

pgierz:
After consideration, I am closing this PR.

The problem it set out to solve — *where does this experiment's data live, who hosts it, how do I get it* — turns out to be fully covered by existing standards plus optional `fsspec` adaptors. Building a custom extension would duplicate them and collide with a published one.

Some points:

1. **Name/prefix collision.** There is already an official [`stac-extensions/storage`](https://github.com/stac-extensions/storage) extension that owns the `storage:` prefix (cloud object-store access: `storage:schemes` / `storage:refs`, region, requester_pays).
2. **`last_access` is volatile runtime state**, not catalog metadata — it changes on every read. I am also not sure what the purpose is for having this?
3. **`storage:type` (gpfs/lustre) is not actionable.** Nothing using a catalog will behave differently based on the filesystem brand. What *is* interesting and useful for programs using this — which driver/protocol opens the file, and whether it needs staging. This can be encoded is the **URI scheme** of the href, which `fsspec` dispatches on natively (`file://`, `scoutfs://`, `s3://`, if we are lucky, one day `slk://`).
4. **Multiple locations are already modelled** by [`stac-extensions/alternate-assets`](https://github.com/stac-extensions/alternate-assets) (mirrors of the same asset), which composes with the official storage extension for cloud copies.
5. **Institution is STAC core.** The operating institution is a [`providers`](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md) entry with `roles: ["host"]` (producer / processor / licensor / host are the four core roles). No extension needed.

PR #1512, opened 2026-07-15
feat(esm_catalog): PR-A1e — namelist extension

siligam:
Part of the esm_catalog decomposition, per [the PR-A1a foundation plan](https://raw.githubusercontent.com/esm-tools/esm_tools/22519379dc96e17e65d8d302c6d969375f0ec139/src/esm_catalog/PR-A1A-STAC-FOUNDATION-PLAN.md). Follows #1504 (A1b core), branches from `release`.

Adds `src/esm_catalog/namelist.py`:
- `add_namelist_extension(collection, namelists)` — collection-level `nml:files`, `nml:groups`, `nml:parameters` (flattened `"group:key"` → scalar, for CQL2 filtering). Drops `nml:raw` from the pr-a1a version per the plan — storing the full raw namelist as JSON on every collection could be megabytes for complex setups.
- `add_namelist_item_extension(item, ctx)` — item-level `nml:{component}:{group}:{key}` entries across **all** components, read from `ctx.namelists_by_component` (kept as-is per the plan; performs no scanning).

Wired into `make_collection` (using the collection's own component's entry from `ctx.namelists_by_component`) and `make_item`.

Also adds `configs/stac-extensions/namelist/v1.0.0/schema.json` (custom extension — no external community schema exists) and `tests/test_esm_catalog/test_stac_namelist_ext.py`.

## Test plan
- [x] `pytest tests/test_esm_catalog` — 51 passed
- [x] New tests cover: collection no-op/flattening/files+groups/`nml:raw` dropped/nested-dict+long-list skipping/idempotent URL append, item no-op/multi-component coverage, and schema validation (via `jsonschema`) for both Collection and Item shapes


siligam:
@mandresm heads-up: per our Webex split this one was yours — I went ahead and implemented it anyway while I had momentum finishing the A1 plan (A1c/A1f landed as #1511/#1513 the same day, and this one follows the plan's spec directly: `nml:raw` dropped, files/groups/parameters kept, item-level function unchanged). If you'd already started on it or want it done differently, happy to close this in favour of yours — otherwise it's ready for your review.

src/esm_catalog/namelist.py
```diff
@@ -0,0 +1,115 @@
...
+    collection.extra_fields["nml:files"] = files
+    collection.extra_fields["nml:groups"] = sorted(groups)
+    collection.extra_fields["nml:parameters"] = parameters
+
+    url = EXTENSION_URLS["namelist"]
+    if url not in collection.stac_extensions:
+        collection.stac_extensions.append(url)
+
+
+def _flatten_for_search(namelists: dict) -> dict:
```
pgierz: I would prefer some kind of meaningful types here to make it easier to read. Dict of what? I suppose you mean:

```suggestion
def _flatten_for_search(namelists: dict[Upath, f90nml.Namelist) -> dict:
```

What types are you expecting?

src/esm_catalog/namelist.py
```diff
@@ -0,0 +1,115 @@
...
+from esm_catalog.registry import EXTENSION_URLS
+
+if TYPE_CHECKING:
+    import pystac
+
+# Values with more elements than this are skipped when flattening for search.
+_MAX_LIST_LENGTH = 10
+
+
+def add_namelist_extension(collection: "pystac.Collection", namelists: dict) -> None:
```
pgierz: `namelists: dict` gives no meaning as to what it is. 

I also do not think that it is good that this function mutates `collection`, was that the design pattern @mandresm and you had discussed? I know I was gone for a while, I may have missed something...

tests/test_esm_catalog/schemas/namelist-v1.0.0.json
```diff
@@ -0,0 +1,95 @@
...
+  "description": "STAC Namelist Extension for STAC Items and Collections, exposing Fortran namelist parameters for search.",
+  "type": "object",
+  "required": [
+    "stac_extensions"
+  ],
+  "properties": {
+    "stac_extensions": {
+      "type": "array",
+      "contains": {
+        "const": "https://esm-tools.github.io/stac-extensions/namelist/v1.0.0/schema.json"
```
pgierz: Are we going to be publishing schemas? Has that been decided?

tests/test_esm_catalog/test_stac_namelist_ext.py
```diff
@@ -0,0 +1,179 @@
...
+    assert col.stac_extensions == []
+
+
+def test_collection_extension_flattens_parameters():
+    col = make_collection(_ctx())
+    add_namelist_extension(
+        col, {"namelist.echam": {"runctl": {"delta_time": 450, "lcouple": True}}}
+    )
+    assert col.extra_fields["nml:parameters"]["runctl:delta_time"] == 450
+    assert col.extra_fields["nml:parameters"]["runctl:lcouple"] is True
```
pgierz: It would be nice to confirm the behavior also with a real `f90nml` object. I think (I hope!) that is what is passed around in the actual program flow.

pgierz (commented):
I need a second, more comprehensive review for this one, but here are some initial comments.

src/esm_catalog/namelist.py
siligam: Good ask, though not quite `dict[Upath, f90nml.Namelist]` — this function never sees raw `f90nml.Namelist` objects or `Upath` keys. The actual shape is `dict[str, dict[str, dict[str, Any]]]`: filename → group → key → scalar (see the docstring a few lines up). Will add that as the type hint.

tests/test_esm_catalog/test_stac_namelist_ext.py
siligam: Fair — current tests pass plain dicts as stand-ins. Will add a test that builds a real `f90nml.Namelist` and converts it the way the (future) scan layer would, so the flattening logic is exercised against the real type, not just its dict shape.

src/esm_catalog/namelist.py
siligam: Same shape as the add_contacts/paleo discussion elsewhere — add_namelist_extension(collection, namelists) mutates collection in place by design, matching every other add_*_extension function in this codebase (item.py, paleo.py, now datacube.py).

tests/test_esm_catalog/schemas/namelist-v1.0.0.json
siligam: Good catch — no, this was never actually decided. There was a different approach to this in the earlier, larger PR before it got split into this series of smaller ones; some of that context didn't carry over cleanly through the split. I've taken a couple of stabs at a schema-publishing approach here, but I don't have a strong opinion on the right way forward — happy to defer to whatever you and Miguel land on.

configs/stac-extensions/namelist/v1.0.0/schema.json
```diff
@@ -0,0 +1,95 @@
...
+          "type": [
+            "string",
+            "number",
+            "boolean",
+            "null"
+          ]
+        },
+        {
+          "type": "array",
+          "maxItems": 10,
```
pgierz: Why max 10?

src/esm_catalog/namelist.py
pgierz, line 1: I'd like to go over this one again. In particular I am not sure if adding the nml properties to the item level is correct, or if that belongs in the collection. I remember there were reasons, so I want a second look...

src/esm_catalog/namelist.py
```diff
@@ -0,0 +1,120 @@
...
+
+from typing import TYPE_CHECKING, Any
+
+from esm_catalog.registry import EXTENSION_URLS
+
+if TYPE_CHECKING:
+    import pystac
+
+# Values with more elements than this are skipped when flattening for search.
+_MAX_LIST_LENGTH = 10
```
pgierz: Why?

siligam:
@mandresm just a nudge on this one whenever you get a chance — pgierz flagged it needs a fuller review pass, been sitting since the 7/28 merge.

pgierz (changes requested):
wip, review not finished.

src/esm_catalog/namelist.py
```diff
@@ -0,0 +1,129 @@
...
+        groups.update(file_groups)
+    parameters: dict[FlatKey, NamelistValue] = {}
+    for filename, group, key, value in _iter_queryable_params(namelists):
+        parameters[f"{filename}:{group}:{key}"] = value
+    fields = {
+        "nml:files": sorted(namelists),
+        "nml:groups": sorted(groups),
+        "nml:parameters": parameters,
+    }
+    validate({"type": "Collection", "stac_extensions": [_URL], **fields}, "namelist")
```
pgierz: This is wrong. Shouldn't we validate the collection itself, rather than an on-the-fly-one?

src/esm_catalog/namelist.py
```diff
@@ -0,0 +1,129 @@
...
+    item: pystac.Item, namelists_by_component: NamelistsByComponent
+) -> None:
+    """Set item-level nml:{component}:{group}:{key} from *namelists_by_component*."""
+    props: dict[str, NamelistValue] = {}
+    for component, namelists in namelists_by_component.items():
+        for filename, group, key, value in _iter_queryable_params(namelists):
+            props[f"nml:{component}:{filename}:{group}:{key}"] = value
+    if not props:
+        return
+    validate({"type": "Feature", "stac_extensions": [_URL], "properties": props}, "namelist")
```
pgierz: Same here

src/esm_catalog/namelist.py
```diff
@@ -0,0 +1,129 @@
...
+        for group_name, params in group_entries:
+            if counts[group_name] > 1:
+                index = next_index.get(group_name, 0)
+                group = f"{group_name}[{index}]"
+                next_index[group_name] = index + 1
+            else:
+                group = group_name
+            for key, value in params.items():
+                if _is_queryable(value):
+                    yield filename, group, key, value
```
pgierz, line 113: I still dislike that we even need this approach...

pgierz (commented):
A few things we still need to discuss, see below (or above)

src/esm_catalog/namelist.py
```diff
@@ -0,0 +1,129 @@
...
+FlatKey = str
+"""A flattened 'file:group:key' identifier, e.g. 'namelist.echam:runctl:delta_time'."""
+
+Namelist = f90nml.Namelist
+"""A parsed Fortran namelist (group -> parameters; nested groups are Namelists)."""
+
+NamelistValue = Union[str, int, float, bool, None, list, Namelist]
+"""An f90nml value: a scalar, a list, or a nested group (Namelist)."""
+
+ComponentNamelists = dict[NamelistFileName, Namelist]
```
pgierz: `Filename` not `FileName` (I am picky about this):

```suggestion
ComponentNamelists = dict[NamelistFilename, Namelist]
```

In fact, I would even prefer it to be:


```suggestion
NamelistFile = UPath  # assume appropriate import... 
ComponentNamelists = dict[NamelistFile, Namelist]
```

pgierz: ~Maybe dict isn't even the right type here, we might need a named tuple...~ never mind, that was a silly idea

pgierz (commented):
Some notes above.

pgierz (approved):

src/esm_catalog/collection.py
```diff
@@ -28,6 +32,10 @@ def make_collection(ctx) -> Collection:
         extra_fields={"experiment": ctx.experiment_id, "components": [ctx.component]},
     )
     col.add_link(Link("parent", f"#{ctx.experiment_id}", media_type="application/json"))
+    # [TODO] PG: Ugly call signature for this level and inward is acceptable for now
+    add_namelist_collection_extension(
+        col, ctx.namelists_by_component.get(ctx.component, {})
```
mandresm, line 37: is this a single component or does `ctx.component` content multiple components?

tests/test_esm_catalog/test_stac_namelist_ext.py
```diff
@@ -0,0 +1,282 @@
...
+    namelists.
+    """
+    path = request.param
+    try:
+        return ShippedNamelist(path.name, f90nml.read(str(path)))
+    except Exception as exc:  # jinja templates, non-namelist content, parse errors
+        pytest.skip(f"f90nml cannot parse {path.relative_to(_NAMELIST_ROOT)}: {exc}")
+
+
+def test_every_shipped_namelist_flattens_and_validates(
```
mandresm, line 259: good idea!

mandresm (approved):
Thanks @pgierz, for your last rework. I only have one question, but I think it can be merged anyway

src/esm_catalog/collection.py
pgierz: multiple, it is the shape `{'fesom': ['namelist.io': {...}, 'namelist.ice': {...},]}`

pgierz:
#bump

PR #1513, opened 2026-07-15
feat(esm_catalog): PR-A1f — paleo extension + experiment classification

siligam:
Part of the esm_catalog decomposition, per [the PR-A1a foundation plan](https://raw.githubusercontent.com/esm-tools/esm_tools/22519379dc96e17e65d8d302c6d969375f0ec139/src/esm_catalog/PR-A1A-STAC-FOUNDATION-PLAN.md). Follows #1504 (A1b core), branches from `release`. Fields 21–27.

Adds `src/esm_catalog/paleo.py`:
- `add_paleo_extension(item, paleo_config, paleo_year, reference_year)` — sets `paleo:year`, `paleo:display`, `paleo:reference_year`, `paleo:epoch`, `paleo:period`. Year priority: explicit param → `paleo_config["reference_year"]` → extreme year (< 0 or > 9999) parsed from the item's own datetime strings. No-op for non-paleo items.
- `add_experiment_type(item)` — classifies every item as `paleo`/`control`/`historical` from the `paleo:year` property or the item's own datetimes.

Plan-mandated fixes vs the old pr-a1a version:
- `_add_experiment_type` no longer reads `nml:echam:runctl:dt_start` (fragile assumption that echam is always a component, plus an ordering dependency on the namelist extension) — there's a regression test asserting the namelist property is ignored.
- `paleo:years_bp` is only added when `experiment_type == "paleo"`, not for every dated item.
- Dead `if TYPE_CHECKING: pass` removed; uncalled `detect_paleo_simulation` removed.

`CollectionContext` gets an optional `paleo_config` field (the `general.paleo` config section as a plain dict); `make_item` wires both functions in.

Also adds `configs/stac-extensions/paleo/v1.0.0/schema.json` (custom extension) and `tests/test_esm_catalog/test_stac_paleo.py`.

Note: like #1511/#1512, this touches the same two spots in `item.py` — whichever of the three merges last will need trivial conflict resolution (keep all import/call lines).

## Test plan
- [x] `pytest tests/test_esm_catalog` — 65 passed
- [x] New tests cover: no-op behavior, explicit/config/deep-time-string year resolution, epoch/period, Ma/ka/BCE/CE formatting, ISO year parsing edge cases, all three experiment_type classes, years_bp only for paleo, namelist-property independence, make_item wiring, and jsonschema validation (schema also probed to reject bad enum values and non-integer years)


src/esm_catalog/paleo.py
```diff
@@ -0,0 +1,182 @@
...
+
+    item.properties["paleo:year"] = geo_year
+    item.properties["paleo:display"] = _format_geological(geo_year, reference_year)
+    item.properties["paleo:reference_year"] = reference_year
+
+    if paleo_config:
+        if "epoch" in paleo_config:
+            item.properties["paleo:epoch"] = paleo_config["epoch"]
+        if "period" in paleo_config:
+            item.properties["paleo:period"] = paleo_config["period"]
```
siligam: Same silent-null-write class as the `roles` thing on #1506, just spelled with `if key in dict:` + direct index instead of `.get(key, default)` — so a `.get(..., [` grep won't catch it. A bare `epoch:` (or `period:`) under `general.paleo` in YAML is present-but-`None`, so `"epoch" in paleo_config` is `True` and `"paleo:epoch": null` gets written into the item. If the paleo schema types these as strings it's schema-invalid, and nothing in `src/esm_catalog` calls `.validate()`, so it ships unchecked (see #1506 discussion + repro: https://gist.github.com/siligam/58cad44b97c831c2f41d22d681c3ddbd).

Cheap fix: gate on truthiness instead of presence — `if paleo_config.get("epoch"):` / `if paleo_config.get("period"):`.

siligam: Fixed in 19a5fafd6 — gated on `paleo_config.get("epoch")`/`.get("period")` truthiness, plus a `test_bare_epoch_period_do_not_write_null` regression test. Full paleo suite green (26 passed).

src/esm_catalog/paleo.py
pgierz, line 1: General comment. LGM is 21k, not 20k

src/esm_catalog/paleo.py
```diff
@@ -0,0 +1,184 @@
...
+geological time periods (e.g., Last Glacial Maximum at 20,000 years ago,
+Eocene at 50 million years ago), plus a coarse experiment_type
+classification for every item.
+
+Properties added by add_paleo_extension (paleo simulations only):
+    paleo:year           - Geological year (negative = past, e.g., -20000)
+    paleo:display        - Human-readable format (e.g., "22.0 ka", "66.0 Ma")
+    paleo:reference_year - Reference year for "years ago" calculation
+    paleo:epoch          - Optional geological epoch (e.g., "Pleistocene")
+    paleo:period         - Optional geological period (e.g., "Quaternary")
```
pgierz: One should also consider a list of strings for abbreviation as a core attribute here. What do you think @christian-stepanek?

configs/stac-extensions/paleo/v1.0.0/schema.json
```diff
@@ -0,0 +1,63 @@
...
+          "type": "integer"
+        },
+        "experiment_type": {
+          "title": "Experiment classification",
+          "type": "string",
+          "enum": [
+            "paleo",
+            "control",
+            "historical"
+          ]
```
pgierz: This feels incomplete and needs discussion. What are you trying to do with this attribute? Please explain.

What about future scenarios? That leads the question: is "paleo" the right name...

src/esm_catalog/paleo.py
```diff
@@ -0,0 +1,184 @@
...
+    else:
+        exp_type = "historical"
+
+    item.properties["experiment_type"] = exp_type
+    if exp_type == "paleo":
+        item.properties["paleo:years_bp"] = 1950 - start_year
+        _register(item)
+
+
+def _register(item: "pystac.Item") -> None:
```
pgierz: Dostring? Magical side effect function?

src/esm_catalog/paleo.py
```diff
@@ -0,0 +1,184 @@
...
+) -> Optional[int]:
+    """Determine the geological year for an item, or None if not paleo."""
+    if explicit_year is not None:
+        return explicit_year
+
+    if paleo_config and "reference_year" in paleo_config:
+        return paleo_config["reference_year"]
+
+    year = _resolve_start_year(item)
+    if year is not None and (year < 0 or year > 9999):
```
pgierz: Why is this check bounded by O and 9999?

pgierz (changes requested):
Before I finish my review: why is the paleo datetime library not used here? Passing around integers or floats instead of date objects seems like it could be improved.

siligam:
@pgierz re: the datetime question in your review — good catch, real gap. `_parse_year_from_iso` was hand-rolling year extraction with string slicing specifically because stdlib `datetime` can't hold years outside `[1, 9999]`, but `esm_calendar.Date` in this same repo was already built to handle exactly that ("compatible with paleo (negative dates)"). Swapped it in — `_parse_year_from_iso` now just does `Date(dt_str).year`, tested it against negative/deep-time/extreme years and it parses correctly (e.g. `-020000-01-01` → `-20000`). Full paleo + catalog suite still green (66 passed).

Also fixed the LGM year while I was in there — should be ~21 ka, not 20 ka, in both the module docstring and the schema description.

Left the abbreviation-attribute and schema-naming/scope comments open since those looked aimed partly at @christian-stepanek.

src/esm_catalog/paleo.py
siligam: Docstring's already there — it was in the first commit of this PR, before this review, so nothing to add there. The mutation itself matches the established convention in this codebase (add_contacts, _check_production_fields, etc.), and the docstring already says explicitly that it mutates stac_extensions — not hidden, just the pattern used everywhere else here.

src/esm_catalog/paleo.py
```diff
@@ -0,0 +1,184 @@
...
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Optional
+
+from esm_calendar import Date
+
+from esm_catalog.registry import EXTENSION_URLS
+
+if TYPE_CHECKING:
+    import pystac
```
mandresm:
```suggestion
```

src/esm_catalog/paleo.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def add_paleo_extension(
+    item: "pystac.Item",
```
mandresm:
```suggestion
    item: pystac.Item,
```

src/esm_catalog/paleo.py
mandresm: We always use pystac and it's imported above.

src/esm_catalog/item.py
```diff
@@ -62,6 +63,8 @@ def make_item(
 
     add_contacts(item, ctx)
     add_datacube_extension(item, metadata)
+    add_paleo_extension(item, ctx.paleo_config)
```
mandresm: This does not only add the extension but also some metadata, so I would call it something like `add_paleo_data` or similar

src/esm_catalog/paleo.py
```diff
@@ -0,0 +1,184 @@
...
+        # present-but-None and would otherwise write a schema-invalid null.
+        if paleo_config.get("epoch"):
+            item.properties["paleo:epoch"] = paleo_config["epoch"]
+        if paleo_config.get("period"):
+            item.properties["paleo:period"] = paleo_config["period"]
+
+    _register(item)
+
+
+def add_experiment_type(item: "pystac.Item") -> None:
```
mandresm:
```suggestion
def add_experiment_type(item: pystac.Item) -> None:
```

mandresm: Does this function belong here or somewhere more general?

mandresm: I think this is mixing our internal definition of paleo (for the paleo dates functionality) with the `general.scenario` terminology. This can be taken directly from `general.scenario`. What do you think @pgierz? Can you give me an example of the name of a paleo scenario?

src/esm_catalog/paleo.py
```diff
@@ -0,0 +1,184 @@
...
+
+def _register(item: "pystac.Item") -> None:
+    """Add the paleo extension URL to *item*.stac_extensions, once."""
+    url = EXTENSION_URLS["paleo"]
+    if url not in item.stac_extensions:
+        item.stac_extensions.append(url)
+
+
+def _resolve_paleo_year(
+    item: "pystac.Item",
```
mandresm:
```suggestion
    item: pystac.Item,
```

src/esm_catalog/paleo.py
```diff
@@ -0,0 +1,184 @@
...
+    if paleo_config and "reference_year" in paleo_config:
+        return paleo_config["reference_year"]
+
+    year = _resolve_start_year(item)
+    if year is not None and (year < 0 or year > 9999):
+        return year
+    return None
+
+
+def _resolve_start_year(item: "pystac.Item") -> Optional[int]:
```
mandresm:
```suggestion
def _resolve_start_year(item: pystac.Item) -> Optional[int]:
```

src/esm_catalog/paleo.py
mandresm: From reading the code and the comments, the 9999 seems to come from `esm_calendar.Date` but I have not found it in the source code, only in some namelists.

tests/test_esm_catalog/schemas/paleo-v1.0.0.json
mandresm, line 1: this is a repeated schema I believe, for the single purpose of testing. It can be removed, we only need one.

mandresm (changes requested):
Some suggestions and questions

src/esm_catalog/paleo.py
pgierz: This always gets mixed up, so in my view scenario paleo should be things like time period abbreviations, so LIG or LGM or HOL 6 K

src/esm_catalog/paleo.py
pgierz: Maybe the better question is: what are we trying to check here exactly?

configs/stac-extensions/paleo/v1.0.0/schema.json
```diff
@@ -0,0 +1,63 @@
...
+      "const": "Feature"
+    },
+    "properties": {
+      "type": "object",
+      "properties": {
+        "paleo:year": {
+          "title": "Geological year",
+          "description": "Geological year the simulation represents (negative = past, e.g. -21000 for LGM).",
+          "type": "integer"
+        },
```
pgierz: This might also need some concept of range?

tests/test_esm_catalog/test_stac_paleo.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def _bare_item(dt=datetime(2000, 1, 1, tzinfo=timezone.utc)):
+    return Item(
+        id="i",
+        geometry=None,
+        bbox=None,
+        datetime=dt,
+        properties={},
+    )
```
pgierz: Should be a proper fixture

pgierz (commented):
I'd like to check for any geology/proxy standards we could recycle.

pgierz:
OK, I have discussed with @claude extensively. The overall idea is that, since we are under-the-hood doing JSON-LD (just like the WRCP is doing for CMIP) it would be good to recycle as many names as we can. Here is a synthesis of the discussion with the AI. I'll add some more notes from my perspective later on:

---

**Recycle the `paleo:` vocabulary instead of minting it — and check the altitude**

Did a vocabulary review on this (motivated by my earlier "is 'paleo' the right name / any standards to recycle" comments). Most of what the extension mints already exists as published, dereferenceable vocabularies. Proposing we reference those instead of custom terms — same "complement, don't reinvent" line `esm-prod-cv` takes with WCRP-universe.

**1. `experiment_type` enum → drop it; use the WCRP-CMIP CV.**
`paleo`/`control`/`historical` and the scenario names are already canonical CMIP6/PMIP4 terms:
- `experiment_id`: `lgm`, `midHolocene`, `lig127k`, `past1000` (verbatim in [`CMIP6_experiment_id.json`](https://github.com/WCRP-CMIP/CMIP6_CVs/blob/main/CMIP6_experiment_id.json))
- `activity_id`: `PMIP` — there is no standalone "paleo" activity in WCRP; paleoclimate is `PMIP`. So `experiment_type` reduces to reusing `activity_id`.

This is also what @mandresm was pushing on (take it from the CV / `general.scenario`, don't invent an internal enum). Validate the strings with `esgvoc`.

**2. `paleo:epoch` / `paleo:period` (free strings) → IRIs into the ICS chart.**
The International Chronostratigraphic Chart is published as SKOS/RDF Linked Data, CC-BY, with stable IRIs:
- `http://resource.geosciml.org/classifier/ics/ischart/Pleistocene` (resolves: rank Epoch, bounds −2.58 → −0.0117 Ma)
- `.../ischart/Quaternary`

Attach via a standard predicate — `dcterms:temporal` accepts a bare named-period IRI ([DCAT-2 pattern](https://www.w3.org/TR/vocab-dcat-2/)) — so both predicate and value are standard, and the free string becomes a followable link. (ICS units nest, so pointing at the finest unit + `skos:broader` avoids storing both.)

**3. `paleo:years_bp` / `paleo:display` → GeoSchemas `gstime:` + OWL-Time.**
The EarthCube-P419 / GeoSchemas temporal extension (`http://schema.geoschemas.org/contexts/temporal#`) covers exactly this: `geologicTimeUnitAbbreviation` (`"BP"`, `"ka"`, `"Ma"`, `"Ga"`) on an OWL-Time `time:TimePosition`. "BP" defines the 1950 datum, so `paleo:reference_year` becomes redundant. Bonus fields we don't have but will want for proxy data: `uncertainty`, `uncertaintySigma`, `radiocarbonCalibrationCurve`. (Caveat: `gstime:` is a community extension, not a W3C/OGC Rec — vendor/pin a version; don't rely on live IRI resolution. Same for caching the ICS vocab — its host flaked during my checks.)

**Net:** the extension collapses from 7 minted fields to mostly references into ICS + WCRP-CMIP + GeoSchemas, with ~no minted predicates left. Strong answer to "why a custom extension at all," and it makes the catalog joinable with CMIP7 / esgvoc federation.

**4. Placement question (separate from vocabulary) — should this be on the Item or the Collection?**
These functions run in `make_item`, so paleo identity lands per-Item (per file). But by our own `Dataset` == STAC `Collection` == one experiment rule, the experiment classification (`activity_id`/`experiment_id`/epoch) is constant across all files of a run — it reads as Collection-level, with only the per-file time slice (`datetime`, maybe chunk `years_bp`) belonging on the Item. Is the Item-level placement here a consequence of how the STAC-view SPA consumes metadata? If the SPA needs it on Items, fine — but let's make that explicit rather than incidental, and consider duplicating the experiment identity up to the Collection.

Happy to pair on a revised `schema.json` that references these. Also flagging PaST thesaurus (NOAA/PANGAEA paleo variable names) for whenever we catalog proxy datasets — same recycle pattern, different axis.

christian-stepanek:
Sticking to standards as much as possible is good. Yet, we have to acknowledge that paleo is so diverse that there will always be an additional period name that has not been considered yet. In the confluence table of our old simulations I had tried to solve this by giving general names for periods (e.g. Holocene, Last Glacial Maximum, Last Interglacial, Pliocene, Miocene, Oligocene, Eocene, Cretaceous, etc.) and also providing the option to modify these names where appropriate by allowing adding attributes (e.g. "late" or "early" for Pliocene and Miocene) plus an added scalar range (first date, average date, last date) that lets the user decide which rough period that simulation would fall into (this later is also relevant for transient runs to provide the exact start and end dates of the run, e.g. 21kaBP to 1850 - so yes, the scalars have to again be linked to a time unit, these could be ka BP, Ma BP, or CE).

In fairness, this is quite complex, but this is owing to the diversity of simulations being produced. And, I doubt that the problem can be solved ad hoc. We will have to be able to define new periods and year ranges on demand - to avoid that this will end up in chaos would probably necessitate a devising a strategy that is either clearly documented or implemented with rules in any edit fields where one can enter a unit for a simulation (the latter may not even be possible).

src/esm_catalog/item.py
siligam: Fixed in e6a6323 — renamed to add_paleo_data.

src/esm_catalog/paleo.py
siligam: Fixed in e6a6323 — switched to a real `import pystac` and dropped the TYPE_CHECKING guard, same fix already applied to storage.py in #1511.

src/esm_catalog/paleo.py
siligam: Fixed in e6a6323 as part of moving pystac to a real import — all the string-quoted "pystac.Item" forward refs are now plain pystac.Item.

src/esm_catalog/paleo.py
siligam: Fixed in e6a6323 as part of the real pystac import.

src/esm_catalog/paleo.py
siligam: Fixed in e6a6323 as part of the real pystac import.

src/esm_catalog/paleo.py
siligam: Fixed in e6a6323 as part of the real pystac import.

src/esm_catalog/paleo.py
siligam: Confirmed and fixed in e6a6323 — it does come from stdlib datetime: Python's datetime.year is restricted to [datetime.MINYEAR, datetime.MAXYEAR] = [1, 9999], so a year outside that range can only have arrived via esm_calendar.Date's paleo-aware parsing of a pre-formatted ISO string, never a real stdlib datetime object. Replaced the hardcoded 0/9999 with datetime.MINYEAR/MAXYEAR directly, which also fixed an off-by-one: year 0 is outside what a stdlib datetime can represent (MINYEAR=1) but the old `year < 0` check didn't catch it. Added a regression test (test_year_zero_triggers_paleo).

siligam: See my reply above to Miguel's comment — the check is: 'is this year outside what a stdlib datetime object can represent (MINYEAR..MAXYEAR = 1..9999)?' If so, it can only have come from esm_calendar.Date parsing a deep-time ISO string, so treat it as paleo. Now spelled with datetime.MINYEAR/MAXYEAR instead of hardcoded 0/9999, and fixed the year=0 edge case that fell through the old check.

tests/test_esm_catalog/schemas/paleo-v1.0.0.json
siligam: Fixed in e6a6323 — removed the duplicate and pointed test_stac_paleo.py at configs/stac-extensions/paleo/v1.0.0/schema.json directly, same fix already applied for hpc/storage in #1511.

src/esm_catalog/paleo.py
siligam: Already fixed in c2c465096 (esm_calendar.Date-based year parsing) before this review round — LGM is 21,000 years ago throughout the docstring and reference_year examples now.

tests/test_esm_catalog/test_stac_paleo.py
siligam: Fixed in e6a6323 — converted to a make_bare_item pytest fixture (a factory, since the datetime argument varies per test), same pattern used in test_storage.py for #1511.

siligam:
Ran a quick check against real production data to firm this up. Pulled the actual `mis31` PlioMIP3 catalog (Christian's production scan, 47k items) — no paleo extension in it yet, but the item datetimes are real: `5501-01-02` to `7100-01-01` (model-internal calendar for a genuine MIS31 ~1.07 Ma Pleistocene run, not real CE dates).

Ran the current `add_experiment_type()` from this branch against those exact values with no `paleo_config` supplied (matches how the real scan was run):

```
5501-01-02T00:30:00Z -> experiment_type: "historical"
7100-01-01T00:30:00Z -> experiment_type: "historical"
```

So a real deep-time paleo run gets silently tagged "historical" whenever `paleo_config.reference_year` isn't set — the `<1800/1800-1950/>1950` thresholds assume the item's raw year approximates a real CE year, which doesn't hold for model-internal calendars like this one. Concretely backs @pgierz's point: this should come from explicit config/CV terms (`activity_id`/`experiment_id`), not be inferred from the datetime, and should probably fail loudly rather than silently default when that config is missing for a paleo run.

tests/test_esm_catalog/test_stac_paleo.py
```diff
@@ -15,46 +15,52 @@
...
+    def _make(dt=datetime(2000, 1, 1, tzinfo=timezone.utc)):
+        return Item(
+            id="i",
+            geometry=None,
+            bbox=None,
+            datetime=dt,
+            properties={},
+        )
+
+    return _make
```
pgierz: what, why? Why not just return an item?

```suggestion
@pytest.fixture
def item():
    return Item(
        id="i",
        geometry=None,
        bbox=None,
        datetime=datetime(2000, 1, 1, tzinfo=timezone.utc),
        properties={},
    )
```

tests/test_esm_catalog/test_stac_paleo.py
```diff
@@ -15,46 +15,52 @@
...
 
 
-def test_noop_without_config_or_year():
-    item = _bare_item()
-    add_paleo_extension(item)
+def test_noop_without_config_or_year(make_bare_item):
+    item = make_bare_item()
+    add_paleo_data(item)
     assert "paleo:year" not in item.properties
     assert item.stac_extensions == []
```
pgierz:
```suggestion
def test_noop_without_config_or_year(item):
    add_paleo_data(item)
    assert "paleo:year" not in item.properties
    assert item.stac_extensions == []
```

tests/test_esm_catalog/test_stac_paleo.py
```diff
@@ -15,46 +15,52 @@
...
 
 
-def test_explicit_paleo_year():
-    item = _bare_item()
-    add_paleo_extension(item, paleo_year=-20000)
+def test_explicit_paleo_year(make_bare_item):
+    item = make_bare_item()
+    add_paleo_data(item, paleo_year=-20000)
```
pgierz:
```suggestion
def test_explicit_paleo_year(item):
    add_paleo_data(item, paleo_year=-20000)

**Review by pgierz — CHANGES_REQUESTED** [review]
The test code can be simplified considerably. I do not understand why there is a wrapper-ish callable sent into tests as a fixture, that seems over-done...just pass the regular `Item`? I've made a few suggestions, but this would need to be done everywhere.

File: tests/test_esm_catalog/test_stac_paleo.py (hunk 13 above)
Comment (siligam) [reply tests/test_esm_catalog/test_stac_paleo.py]: Done in 6b49f06a9 — replaced the `make_bare_item` factory with a plain `item` fixture, per your suggestion below.

File: tests/test_esm_catalog/test_stac_paleo.py (hunk 14 above)
Comment (siligam) [reply tests/test_esm_catalog/test_stac_paleo.py]: Done in 6b49f06a9 — swapped in exactly this. Two tests needed a non-default datetime (1850, 1000), so those construct their `Item` inline instead of going through the fixture, rather than reintroducing a factory.

File: tests/test_esm_catalog/test_stac_paleo.py (hunk 15 above)
Comment (siligam) [reply tests/test_esm_catalog/test_stac_paleo.py]: Applied in 6b49f06a9.

File: tests/test_esm_catalog/test_stac_paleo.py (hunk 16 above)
Comment (siligam) [reply tests/test_esm_catalog/test_stac_paleo.py]: Applied in 6b49f06a9 (and the rest of the `make_bare_item`-using tests converted the same way).

**Comment by pgierz** [comment]
Alright, most things here have been reduced to "can be added properly somewhere else". The genuinely new thing we need is paleo_datetime, paleo_start_datetime, and paleo_end_datetime to associate the runs with a paleo period. The numbers are what is important, what the geological name of this thing is can be added by the scientists themselves. 

I will do a self-rereview tomorrow.

File: src/esm_catalog/paleo.py [hunk 17]
```diff
@@ -0,0 +1,84 @@
...
+
+from esm_catalog.registry import EXTENSION_URLS
+
+_PALEO_URL = EXTENSION_URLS["paleo"]
+
+# The config keys map 1:1 onto the paleo: item properties.
+_KEYS = ("datetime", "start_datetime", "end_datetime")
+
+
+@lru_cache(maxsize=None)
```
mandresm, line 45: cool!

src/esm_catalog/item.py
```diff
@@ -62,6 +63,7 @@ def make_item(
 
     add_contacts(item, ctx)
     add_datacube_extension(item, metadata)
+    add_paleo_data(item, ctx.paleo_config)
```
mandresm, line 66: Does it make sense to have this at the level of the item, or would it make more sense to have it at the level of the collection?

mandresm (commented):
Only one comment. Thanks!

mandresm (approved):

src/esm_catalog/item.py
mandresm: After discussion with @pgierz, we agreed that it makes sense to have it at both levels.

mandresm:
IMO this is ready to merge, but one would need to fix the tests.

Responding to @christian-stepanek 

> Yet, we have to acknowledge that paleo is so diverse that there will always be an additional period name that has not been considered yet. In the confluence table of our old simulations I had tried to solve this by giving general names for periods (e.g. Holocene, Last Glacial Maximum, Last Interglacial, Pliocene, Miocene, Oligocene, Eocene, Cretaceous, etc.) and also providing the option to modify these names where appropriate by allowing adding attributes (e.g. "late" or "early" for Pliocene and Miocene) plus an added scalar range (first date, average date, last date) that lets the user decide which rough period that simulation would fall into

@christian-stepanek, thanks a lot for your valuable insights. With this implementation the user will be able to label the period as they consider best, but then the user will also need to specify a scalar range within the `general.paleo`, for instance `general.paleo.date_time/start_datetime/end_datetime`.

@pgierz, is it maybe worth adding to the schema a key:value for a `paleo.period` (type-checking: str) so that users don't choose different keys for the labeling of the period? In that case, I think it would also make sense for the schema to exclude any invented keys. What do you think?

tests/test_esm_catalog/test_stac_paleo.py
```diff
@@ -0,0 +1,238 @@
...
+from esm_catalog.collection import make_collection
+from esm_catalog.context import CollectionContext
+from esm_catalog.item import make_item
+from esm_catalog.paleo import add_paleo_data, add_paleo_summary
+from esm_catalog.registry import EXTENSION_URLS
+
+PALEO_URL = EXTENSION_URLS["paleo"]
+SCHEMA_PATH = (
+    Path(__file__).parents[2] / "configs" / "stac-extensions" / "paleo" / "v1.0.0" / "schema.json"
+)
```
pgierz: This should be retrieved with data packaging

pgierz (changes requested):
Almost ready

.github/workflows/esm-catalog-tests.yml
```diff
@@ -34,6 +36,12 @@ jobs:
           # with --no-deps and add only the deps it actually imports (plus a modern
           # pytest; the base pins pytest==7.1.2, which predates 3.12 support).
           pip install -e . --no-deps
-          pip install -U "pytest>=7.4" click loguru pystac shapely universal-pathlib jsonschema
+          # esm_catalog now imports esm_tools (for get_config_filepath); its
+          # __init__ pulls pkg_resources, yaml and colorama at import. pkg_resources
+          # ships with setuptools but was removed in setuptools 81, so pin below it.
+          # paleodatetime is GitHub-only (not on PyPI); needed by the paleo tests.
```
pgierz: AI bookkeeping comment, not needed.

pgierz (changes requested):

pgierz (approved):

pgierz:
#bump

PR #1527, opened 2026-07-30
Draft: T1 Rename

### LINKED ISSUE #1520: T1: Rename current workflow units to the agree with the workflow vocabulary #1519
Rename the current workflow parameters into the new agreed workflow vocabulary #1519

If we find that the current workflow is incompatible with the new vocabulary, skip this step.

pgierz:
Before refactoring the workflow, I wanted to align the internal variable names to P-Plan's ontology. Aside from the copy/paste sed-style renames, I also added test cases to ensure the structure of the keys/values in the workflow dictionary do not change. This will be replaced with proper objects in the next round.

For #1520.

Below this line will go an LLM-Generated PR summary:
- - - - -


tests/test_esm_runscripts/test_workflow_dag.py
```diff
@@ -0,0 +1,170 @@
...
+# package would pull in dask/distributed (arch-broken in this env) and may
+# resolve to a different checkout; loading the module by path with this
+# branch's src/ on sys.path keeps the test pinned to the code under review.
+_REPO = pathlib.Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(_REPO / "src"))
+_spec = importlib.util.spec_from_file_location(
+    "wf_under_test", _REPO / "src/esm_runscripts/workflow.py"
+)
+workflow = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(workflow)
```
pgierz: This is wrong, and should be packaged correctly.

tests/test_esm_runscripts/test_workflow_dag.py
```diff
@@ -0,0 +1,170 @@
...
+    "last": "tidy",
+}
+
+
+# Same chain, expressed purely in the new single ordering property.
+DEFAULT_CHAIN_NEW = {
+    "prepcompute": {"parent_plan": "prepcompute"},  # entry: no preceded_by
+    "compute": {"parent_plan": "compute", "preceded_by": "prepcompute"},
+    "tidy": {"parent_plan": "tidy", "preceded_by": "compute"},
+}
```
pgierz: Belongs in a fixture

tests/test_esm_runscripts/test_workflow_dag.py
```diff
@@ -0,0 +1,170 @@
...
+    "couple_out": {"parent_plan": "couple_out", "run_after": "tidy"},
+}
+
+COUPLED_NEW = {
+    "couple_in": {"parent_plan": "couple_in"},
+    "prepcompute": {"parent_plan": "prepcompute", "preceded_by": "couple_in"},
+    "compute": {"parent_plan": "compute", "preceded_by": "prepcompute"},
+    "tidy": {"parent_plan": "tidy", "preceded_by": "compute"},
+    "couple_out": {"parent_plan": "couple_out", "preceded_by": "tidy"},
+}
```
pgierz: Also belongs in fixture? The primary question here is how to package and handle "Golden" answers?

src/esm_runscripts/batch_system.py
```diff
@@ -596,52 +596,52 @@ def get_submit_command(config, batch_or_shell, runfilename):
...
     @staticmethod
-    def write_simple_runscript(config, cluster, batch_or_shell="batch"):
+    def write_simple_runscript(config, plan, batch_or_shell="batch"):
 
-        # if no cluster is specified, work on the one we are in
-        # if not cluster:
-        #    cluster = config["general"]["jobtype"]
+        # if no plan is specified, work on the one we are in
+        # if not plan:
+        #    plan = config["general"]["jobtype"]
```
pgierz: Dead code, can be removed entirely

src/esm_runscripts/batch_system.py
```diff
@@ -650,43 +650,43 @@ def write_simple_runscript(config, cluster, batch_or_shell="batch"):
...
                     for line in extra:
                         runfile.write(line + "\n")
 
                     # Add actual commands
                     commands = batch_system.get_run_commands(
-                        config, subjob, batch_or_shell
+                        config, sub_plan, batch_or_shell
                     )
-                    # commands = clusterconf.get("data_task_list", [])
+                    # commands = plan_conf.get("data_task_list", [])
```
pgierz: Dead code, can be removed.

tests/test_esm_runscripts/test_workflow_dag.py
pgierz, line 1: this file needs another refinement.

src/esm_runscripts/workflow.py
```diff
@@ -1,38 +1,166 @@
...
+
+_LEGACY_WORKFLOW_KEYS = {
+    "subjob_clusters": "plans",
+    "subjobs": "sub_plans",
+    "subjob_cluster": "parent_plan",
+    "run_after": "preceded_by",
+    "first_task_in_queue": "entry_point",
+    "last_task_in_queue": "exit_point",
+    "next_run_triggered_by": "next_iteration_trigger",
+}
```
pgierz: Does this belong here, or in a data file?

src/esm_runscripts/workflow.py
```diff
@@ -1,38 +1,166 @@
...
+
+
+def _rename_legacy_key(container, old):
```
pgierz: type hints

src/esm_runscripts/workflow.py
```diff
@@ -363,68 +454,55 @@ def collect_all_workflow_information(config):
...
-                                ] == subjob + "_" + model
-
-                    # if not in another cluster, each subjob gets its own
+                    if sub_plan in gw_config["sub_plans"]:
+                        del gw_config["sub_plans"][sub_plan]
+                    # make sure preceded_by refers to that model-specific sub-plan
+                    for other_sub_plan in gw_config["sub_plans"]:
+                        other_conf = gw_config["sub_plans"][other_sub_plan]
+                        if other_conf.get("preceded_by") == sub_plan:
+                            other_conf["preceded_by"] = sub_plan + "_" + model
```
pgierz: Patterns like this with `sub_plan + "_" + model` can be turned to f-strings.

pgierz (commented):
This needs another round, some comments below.

tests/test_esm_runscripts/test_workflow_dag.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@pytest.fixture
+def coupled_legacy(shared_datadir):
+    return _load_workflow(shared_datadir, "workflow_coupled_legacy")
+
+
+@pytest.fixture
+def coupled_new(shared_datadir):
+    return _load_workflow(shared_datadir, "workflow_coupled_new")
```
pgierz: This looks like repeated code, can it be condensed?

tests/test_esm_runscripts/test_workflow_dag.py
```diff
@@ -0,0 +1,130 @@
...
+    ``workflow_block`` is a ``general.workflow`` dict (sub_plans + entry/exit
+    keys) in either vocabulary. The returned shape (``next_submit`` edges,
+    ``called_from`` back-pointers, and the entry/exit plan *names*) is
+    vocabulary-independent, so both vocabularies must produce the same value.
+    """
+    gw = {"plans": {}, **copy.deepcopy(workflow_block)}
+    config = {"general": {"workflow": gw}}
+    workflow.translate_legacy_workflow_keys(config)
+    workflow.complete_plans(config)
+    workflow.order_plans(config)
```
pgierz, line 40: This part mimics the workflow manager in the main flow

tests/test_esm_runscripts/test_workflow_dag.py
```diff
@@ -0,0 +1,130 @@
...
+    wf = config["general"]["workflow"]
+    plans = wf["plans"]
+    return {
+        "next_submit": {p: plans[p].get("next_submit") for p in sorted(plans)},
+        "called_from": {p: plans[p].get("called_from") for p in sorted(plans)},
+        "first": wf.get("entry_point", wf.get("first_task_in_queue")),
+        "last": wf.get("exit_point", wf.get("last_task_in_queue")),
+    }
```
pgierz, line 48: Here, I am not sure. At this point, plans are just dict of strs? So if we "sort" them, doesn't that change the shape of the DAG?

src/esm_runscripts/workflow.py
```diff
@@ -1,38 +1,162 @@
...
+#   configs/esm_software/esm_runscripts/workflow_legacy_keys.yaml
+# ``run_before`` is not listed there -- it is the inverse of ``preceded_by`` and
+# is flipped onto the target sub-plan by :func:`_flip_legacy_run_before`.
+_LEGACY_KEYS_SPEC = esm_parser.yaml_file_to_dict(
+    f"{esm_tools.get_config_filepath()}"
+    "/esm_software/esm_runscripts/workflow_legacy_keys.yaml"
+)
+_LEGACY_WORKFLOW_KEYS = _LEGACY_KEYS_SPEC["renames"]
+_KEYS_DEPRECATED_SINCE = _LEGACY_KEYS_SPEC["deprecated_since"]
+_KEYS_REMOVED_IN = _LEGACY_KEYS_SPEC["removed_in"]
```
pgierz: I would prefer constant-level docstrings on the objects here, rather than a large block comment above.

src/esm_runscripts/workflow.py
```diff
@@ -66,223 +191,186 @@ def display_nicely(config):
...
+    # A plan with no predecessor is only valid once it is the entry point (the
+    # entry may have been carried onto it by the loop above).
+    for plan in gw_config["plans"]:
+        if (
+            "preceded_by" not in gw_config["plans"][plan]
+            and plan != gw_config.get("entry_point")
+        ):
+            logger.error(f"Don't know when to execute plan {plan}.")
+            logger.error(gw_config)
+            sys.exit(-1)
```
pgierz, line 294: I'm surprised that this block condenses so much

pgierz (commented):
Another round is needed.

tests/test_esm_runscripts/test_workflow_dag.py
pgierz: This turns out to just present the keys in as sorted form, nothing else. The inner lists (the actual order of the dag) is not changed.

src/esm_runscripts/workflow.py
pgierz: Yes, this makes sense. We used to have "run before" and "run after" (or comes before/comes after). P-Plan has only *one* word for this, and that is the correct choice (after lots of thinking). Backrefs to where a step's parent is in the chain above it is still possible...

I would still like a second pair of eyes on this block though.

PR #1538, opened 2026-08-05
esm-catalog: Cleanup after foundation

### LINKED ISSUE #1498: Catalog cleanup
Things that need to be cleaned up after all the features of the catalogs are implemented:

- [ ] Consider unifying the `.github/workflows/esm-catalog-tests.yml` with `github/workflows/esm_tools_actions.yml`. See https://github.com/esm-tools/esm_tools/pull/1483#discussion_r3429556124
- [ ] Cleanup the `nonrecurse` for `[tools:pytest]` once catalog become a no opt. See https://github.com/esm-tools/esm_tools/pull/1483#discussion_r3427490717
- [ ] Consider adding all the raises into user_error

pgierz:
Closes #1498


src/esm_catalog/datacube.py
```diff
@@ -19,49 +19,89 @@
...
-    dims = metadata.get("dimensions", {})
-    if not dims:
+
+    dimensions: list[str]
+    unit: str
+    description: str
+
+
+CubeVariables = dict[VariableName, CubeVariable]
+"""Emitted ``cube:variables``: variable name -> CubeVariable."""
```
pgierz: Don't these things belong in `types.py`?

src/esm_catalog/datacube.py
```diff
@@ -19,49 +19,89 @@
...
 
-    cube_vars = _to_cube_variables(metadata.get("variables", []))
-    if cube_vars:
-        item.properties["cube:variables"] = cube_vars
+    # remote schema (upstream stac-extensions) — nothing local to validate against
+    apply_extension(item, Extension.datacube, validate=False)
 
-    register_extension(item, EXTENSION_URLS["datacube"])
 
+def _to_cube_variables(variables: list[dict]) -> CubeVariables:
```
pgierz: `list[dict]` is unclear

src/esm_catalog/collection.py
```diff
@@ -2,59 +2,85 @@
...
 
-def make_collection(ctx: CollectionContext) -> Collection:
-    """Construct a pystac Collection for an experiment component.
+def make_collection(exp_metadata: ExperimentMetadata) -> Collection:
+    """Construct a pystac Collection for an experiment.
 
-    Args:
-        ctx: CollectionContext with experiment_id, component, collection_id.
+    A Collection *is* the experiment — all of its components — so it carries
+    every component's namelists, not one.
```
pgierz: This comment is still AI bookkeeping and a leftover from Collection = component.

src/esm_catalog/collection.py
```diff
@@ -2,59 +2,85 @@
...
-        license=ctx.data_license or "proprietary",
-        extra_fields={"experiment": ctx.experiment_id, "components": [ctx.component]},
-    )
-    col.add_link(Link("parent", f"#{ctx.experiment_id}", media_type="application/json"))
-    # [TODO] PG: Ugly call signature for this level and inward is acceptable for now
-    add_namelist_collection_extension(
-        col, ctx.namelists_by_component.get(ctx.component, {})
+        title=exp_metadata.experiment_id,
+        license=exp_metadata.data_license or "proprietary",
+        extra_fields={"components": sorted(exp_metadata.namelists_by_component)},
```
pgierz: This looks wrong. namelist by component?

src/esm_catalog/collection.py
```diff
@@ -2,59 +2,85 @@
...
         extent=Extent(
-            spatial=SpatialExtent(bboxes=[[-180.0, -90.0, 180.0, 90.0]]),
+            spatial=SpatialExtent(bboxes=[DEFAULT_BBOX]),
             temporal=TemporalExtent(intervals=[[None, None]]),
         ),
-        title=ctx.experiment_id,
-        license=ctx.data_license or "proprietary",
-        extra_fields={"experiment": ctx.experiment_id, "components": [ctx.component]},
-    )
-    col.add_link(Link("parent", f"#{ctx.experiment_id}", media_type="application/json"))
```
pgierz, line 34: What happened to parent? There should be no Parent, that is correct -- the experiment is the root level.

tests/test_esm_catalog/test_stac_collection.py
```diff
@@ -29,92 +24,134 @@ def _item(dt, bbox=None, dt_end=None):
...
     dt = datetime(2000, 1, 1, tzinfo=timezone.utc)
-    update_extent(col, _item(dt, bbox=[0, 0, 10, 10]))
-    assert col.extent.spatial.bboxes == [[0, 0, 10, 10]]
+    update_extent(collection, _item(dt, bbox=[0, 0, 10, 10]))
+    assert collection.extent.spatial.bboxes == [[0, 0, 10, 10]]
 
 
 def test_spatial_extent_merges_second_bbox():
-    col = make_collection(_ctx())
+    collection = make_collection(make_exp_metadata())
```
pgierz: this line is repeated a lot -- should it be a fixture?

tests/test_esm_catalog/test_stac_datacube.py
```diff
@@ -3,29 +3,22 @@
...
-        bbox=None,
-        datetime=datetime(2000, 1, 1, tzinfo=timezone.utc),
-        properties={},
-    )
+DATACUBE_URL = EXTENSION_URLS["datacube"]
+# Derive the vendored schema filename from the registry URL's version segment
+# (…/datacube/<version>/schema.json) so a registry bump points at a missing file
+# and fails loudly, instead of silently validating against a stale local copy.
+_DATACUBE_VERSION = DATACUBE_URL.split("/datacube/")[1].split("/")[0]
+SCHEMA_PATH = Path(__file__).parent / "schemas" / f"datacube-{_DATACUBE_VERSION}.json"
```
pgierz: Wrong way to load a schema?

pgierz (commented):
Some first notes, needs another pass...

src/esm_catalog/item.py
```diff
@@ -1,173 +1,233 @@
...
     )
 
-    add_contacts(item, ctx)
-    add_datacube_extension(item, metadata)
-    add_namelist_item_extension(item, ctx.namelists_by_component)
-    add_paleo_data(item, ctx.paleo_config)
+    add_contacts_item_extension(item, exp_metadata)
+    add_datacube_item_extension(item, file_metadata)
+    add_namelist_item_extension(item, exp_metadata.namelists_by_component)
+    add_paleo_item_extension(item, exp_metadata.paleo_config)
```
pgierz, line 78: These two still look wrong, why are they different? Just pass in the entire exp_metadata?

tests/test_esm_catalog/conftest.py
```diff
@@ -41,12 +39,8 @@ def bare_item():
...
-def make_ctx():
-    """Factory for a CollectionContext — the end-to-end builder tests need one."""
-
-    def _make(**kwargs):
-        return CollectionContext(
-            experiment_id="exp", component="echam", collection_id="exp", **kwargs
-        )
-
-    return _make
+def temp_nc(tmp_path):
```
pgierz: type hint???

tests/test_esm_catalog/helpers.py
```diff
@@ -0,0 +1,68 @@
...
+
+from esm_catalog.models import ExperimentMetadata
+
+
+def make_exp_metadata(**kwargs) -> ExperimentMetadata:
+    """An ExperimentMetadata with sensible defaults; override any field via kwargs."""
+    return ExperimentMetadata(**{"experiment_id": "exp-alpha", **kwargs})
+
+
+def make_file_metadata(**kwargs) -> dict:
```
pgierz: Shouldn't this be FileMetadata as return?

pgierz (commented):
getting there, still some small improvements to be considered.

src/esm_catalog/item.py
```diff
@@ -1,173 +1,233 @@
...
     )
 
-    add_contacts(item, ctx)
-    add_datacube_extension(item, metadata)
-    add_namelist_item_extension(item, ctx.namelists_by_component)
-    add_paleo_data(item, ctx.paleo_config)
+    add_contacts_item_extension(item, exp_metadata.contacts)
+    add_datacube_item_extension(item, file_metadata)
+    add_namelist_item_extension(item, exp_metadata.namelists_by_component)
+    add_paleo_item_extension(item, exp_metadata.paleo_config)
```
pgierz: The file_metadata might also only need specific sub-parts?

pgierz (commented):
One more comment

pgierz:
@mandresm: I am happy with this as it is, please review.

Main points: this sets consistent names, accessors, and function interfaces for the various parts that we merged in separately.

src/esm_catalog/collection.py
```diff
@@ -2,59 +2,80 @@
 
 from __future__ import annotations
 
-from pystac import (Collection, Extent, Item, Link, SpatialExtent,
-                    TemporalExtent)
+from pystac import Collection, Extent, Item, SpatialExtent, TemporalExtent
 
-from esm_catalog.context import CollectionContext
```
mandresm, line 8: so CollectionContext is killed, okay

src/esm_catalog/collection.py
```diff
@@ -2,59 +2,80 @@
...
+    -------
+    pystac.Collection
+        A Collection with a default global extent and no parent — the
+        experiment is the catalog root.
     """
-    col = Collection(
-        id=ctx.collection_id,
-        description=ctx.description,
+    collection = Collection(
+        id=exp_metadata.experiment_id,
```
mandresm: where does experiment_id come from? Be aware of possible duplicates (e.g. different people naming different experiments the same, the same person creating different experiments with the same name in different directories...)

src/esm_catalog/collection.py
mandresm: yes, that sounds stange. I expect a list of components, I don't now why this has anything to do with the namelists

src/esm_catalog/contacts.py
```diff
@@ -0,0 +1,34 @@
...
+from __future__ import annotations
+
+import pystac
+
+from esm_catalog.models import Contact
+from esm_catalog.registry import Extension
+from esm_catalog.stac_ext import apply_extension
+
+
+def add_contacts_item_extension(item: pystac.Item, contacts: list[Contact]) -> None:
```
mandresm: The production checkers are gone if I understand correctly. We should have an schema for our production requirements from the `context.py` deleted file (`["name", "institution"]`) to don't loose functionality here

src/esm_catalog/collection.py
```diff
@@ -2,59 +2,80 @@
...
-    add_namelist_collection_extension(
-        col, ctx.namelists_by_component.get(ctx.component, {})
+        title=exp_metadata.experiment_id,
+        license=exp_metadata.data_license or "proprietary",
+        extra_fields={"components": sorted(exp_metadata.namelists_by_component)},
     )
-    add_paleo_summary(col, ctx.paleo_config)
-    return col
+    add_namelist_collection_extension(collection, exp_metadata.namelists_by_component)
+    add_paleo_collection_extension(collection, exp_metadata.paleo_config)
```
mandresm, line 51: don't the contacts also belong to the collection level?

src/esm_catalog/collection.py
```diff
@@ -2,59 +2,80 @@
...
 
-from esm_catalog.paleo import add_paleo_summary
+# STAC-valid "coverage unknown" default; update_extent swaps it out on the first
+# real item bbox (tracked by a flag, since a global item equals this by value).
+DEFAULT_BBOX: BBox = [-180.0, -90.0, 180.0, 90.0]
 
 
-def make_collection(ctx: CollectionContext) -> Collection:
-    """Construct a pystac Collection for an experiment component.
+def make_collection(exp_metadata: ExperimentMetadata) -> Collection:
```
mandresm, line 18: With the deletion of the context all of our production checkers are also gone. I think it would make sense to move them to an schema as well so that we can still do our production validation.

src/esm_catalog/models.py
```diff
@@ -0,0 +1,84 @@
...
+class StacContact(TypedDict, total=False):
+    """A Contact serialized to the STAC contacts extension entry shape."""
+
+    name: str
+    roles: list[str]
+    identifier: str
+    organization: str
+
+
+class Contact(BaseModel):
```
mandresm, line 31: why is this here and not in contwcts.py? I think I am missing something

src/esm_catalog/models.py
```diff
@@ -0,0 +1,84 @@
+"""Pydantic models for experiment identity and PI/author contacts."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Optional, TypedDict, cast
+
+from pydantic import BaseModel, ConfigDict, Field, SkipValidation, field_serializer
```
mandresm, line 9: I cannot find where BaseModel is defined. I clearly don't understand how pydantic works. Can you briefly explain how this works?

tests/test_esm_catalog/test_stac_collection.py
```diff
@@ -28,93 +31,124 @@ def _item(dt, bbox=None, dt_end=None):
...
+        namelists_by_component={
+            "echam": {"namelist.echam": {"runctl": {"delta_time": 450}}},
+            "fesom": {"namelist.fesom": {"clockinit": {"yearnew": 1850}}},
+        }
+    )
+    collection = make_collection(exp_metadata)
+    assert collection.extra_fields["components"] == ["echam", "fesom"]
+    params = collection.extra_fields["nml:parameters"]
+    assert params["echam:namelist.echam:runctl:delta_time"] == 450
+    assert params["fesom:namelist.fesom:clockinit:yearnew"] == 1850
```
mandresm, line 62: looks good

tests/test_esm_catalog/test_stac_namelist_ext.py
```diff
@@ -238,6 +258,13 @@ def _shipped_namelist_files() -> list[Path]:
     ]
 
 
+def test_shipped_namelist_corpus_is_present():
```
mandresm: what is this corpus_is_present? is this a naming that we want to follow and it's followed in other projects? if it's not, i suggest a change to a more specific term

mandresm (changes requested):
Thanks @pgierz for this very necessary cleanup. Some minor suggestions and two general requests:
1. can you also provide a brief developer documentation where it is briefly explained how the architecture of this is thought and the intended use of the major components (collections and items, how to extend them, how to validate, our strategy for validation). In this way we can always come back to the code to add a feature or fix a bug with a fast read on the documentation.
2. can you provide a schema for the production-keys that were represented in the production checkers in the previous version?

once addressed, feel free to merge

tests/test_esm_catalog/test_stac_namelist_ext.py
pgierz: ...this is LLM junk, I think(?) It wants to check if the tests themselves even works, by ensuring the namelists we give (via git clone or checkout) are even there. I will confirm the purpose of this one.

src/esm_catalog/models.py
pgierz: `BaseModel` is internal to `pydantic` and is "This is a thing that has validate-able properties" 

This page is perhaps a good starting point: https://pydantic.dev/docs/validation/dev/concepts/models/

src/esm_catalog/models.py
pgierz: I think this was to avoid cross-imports. I'll double check.

src/esm_catalog/datacube.py
pgierz: From @claude: Discussed with @pgierz, he agrees. Keeping these local, consistent with `paleo.py` / `namelist.py`: `types.py` is the shared *scanner contract* (`FileMetadata`, `ScannedVariable`, `VariableName` — the input every reader must meet), while extension-**emitted** shapes live with their extension. `CubeVariable`/`CubeVariables` are the datacube extension's output shape, so they stay in `datacube.py`.

src/esm_catalog/datacube.py
pgierz: From @claude: Discussed with @pgierz, he agrees. Resolved — `_to_cube_variables` now takes `list[ScannedVariable]`; no bare `list[dict]` remains.

tests/test_esm_catalog/test_stac_namelist_ext.py
pgierz: Discussed with the Intelligence. This is indeed junk and can go away -- it is guarding against "someone might not have checked out namelists". 

I will remove this in the next commits...

src/esm_catalog/models.py
pgierz: From @claude:  `Contact` stays here: models.py is the pydantic data-vocabulary home (its docstring covers contacts), while contacts.py is the STAC-extension applier that imports it. No cross-import issue — contacts.py → models is one-way, and moving it wouldn't break a cycle either; it's a data-vs-behavior split.

From me: sure, I agree with this. models.py for anything that is pydantic-like is fine, gives you only one place to look. The downside is that it breaks code-locality, but for declaring types, that might not be so important...

src/esm_catalog/collection.py
pgierz: This is better now. it is just "components" with the hint from the type-alias showing you "name of component, e.g. fesom" (or something like this). Comment closed...

src/esm_catalog/collection.py
pgierz: This gets a proper hash now instead of just an expid from the user. It should be noted, however, that the layer that connects this into the database will need to do a check again as well, I don't want to accidentally have a db entry overridden.

src/esm_catalog/contacts.py
pgierz: There is now a Schema for this.

src/esm_catalog/collection.py
pgierz: This is ready now, but only minimally ensures you have Contact and that all of the "default stac" things are set.

When we rectify the reqs against @christian-stepanek, that schema shall be expanded and completed.

pgierz:
I added a docs section to the handbook and we have a preliminary schema. I will bump and merge once tests pass.

pgierz:
#bump

PR #1543, opened 2026-08-11
scan: vocabulary

pgierz:
Add the scan layer's shared value types and aliases.

### DIFF
--- a/src/esm_catalog/scan/types.py
+++ b/src/esm_catalog/scan/types.py
@@ -1,0 +1,70 @@
+"""Legible vocabulary for the scan layer.
+
+Internal immutable value objects (frozen dataclasses) plus a couple of string
+aliases that recur across the walk, sourcing, parallel, and storage modules.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Literal, Optional
+
+from upath import UPath
+
+from esm_catalog.types import ComponentName
+
+ScanPhase = Literal["sourcing", "reading", "writing"]
+"""Which stage of a scan a :class:`ProgressEvent` reports."""
+
+RunStamp = str
+"""A run-segment stamp, e.g. '20000101-20001231' — one finished_config window."""
+
+Md5 = str
+"""A lowercase hex MD5 checksum."""
+
+
+@dataclass(frozen=True)
+class OutputFile:
+    """One output file to scan, with its component and (if known) checksum.
+
+    Produced by the sourcing layer from the experiment config, so the walk does
+    not have to infer the component from the path.
+    """
+
+    path: UPath
+    component: ComponentName
+    md5: Optional[Md5] = None
+
+
+@dataclass(frozen=True)
+class ScanFailure:
+    """A file the scan could not read; collected for the failures sidecar."""
+
+    path: str
+    error: str
+
+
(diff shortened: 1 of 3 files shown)


src/esm_catalog/scan/types.py
```diff
@@ -0,0 +1,71 @@
+"""Legible vocabulary for the scan layer.
+
+Internal immutable value objects (frozen dataclasses — no validation or
+serialization is needed, so pydantic would be overkill) plus a couple of string
+aliases that recur across the walk, sourcing, parallel, and storage modules.
```
pgierz: Reword, this is AI language leaking out -- just needs to say internal immutable value objects, no mention of pydnantic required.

src/esm_catalog/types.py
```diff
@@ -50,20 +71,20 @@ class ScannedVariable(TypedDict, total=False):
...
     component: ComponentName
     format: str
-    dimensions: dict
+    dimensions: CubeDimensions
     datetime_start: datetime
     datetime_end: datetime
     datetime_str: str
-    output_frequency: str
-    geometry: dict
+    frequency: str
```
pgierz, line 88: why is this changed?

src/esm_catalog/scan/types.py
```diff
@@ -0,0 +1,71 @@
...
+class OutputFile:
+    """One output file to scan, with its component and (if known) checksum.
+
+    Produced by the sourcing layer from the experiment config, so the walk does
+    not have to infer the component from the path.
+    """
+
+    path: UPath
+    component: ComponentName
+    md5: Optional[Md5] = None
```
pgierz, line 36: This might bite with @mandresm's file-dict work, I will need to keep it in mind.

src/esm_catalog/scan/types.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@dataclass(frozen=True)
+class ScanFailure:
+    """A file the scan could not read; collected for the failures sidecar."""
+
+    path: str
```
pgierz, line 43: Not UPath?

pgierz (commented):
Some first notes for the LLM.

src/esm_catalog/types.py
pgierz: This is to align the left-hand-side with CMIP vocabulary, which always uses. "frequency".

src/esm_catalog/scan/types.py
pgierz, line 1: For the question: "Don't we already have a `types.py`?"

Yes, we do. I do not entirely know why the LLM decided to put this into its own file (it claims something about locality and cleanliness, which...hmm...ok...)

My recommendation is to take this as-is for the moment, and, at the end of the project, we go through and move all the types to a common place (e.g. `src/esm_tools/types.py` and ensure that they are used across the entire project correctly. That is a big enough PR to be it's own thing.

pgierz:
While I have some notes on this one, I think it is good as-is and could be marked as ready so the next PR in this stack can continue. We will (unfortunately) need another round of cleanup after the feature is fully complete, but that will only be easy to see once it is put together. 

I would at this stage rather not nitpick about things that can be arguably placed under the vague category of "taste" for what is easy-to-ready code.

PR #1544, opened 2026-08-11
scan: reader framework

pgierz:
Add the file-reader framework: the Reader protocol, registry, and format detection.

### DIFF
--- a/src/esm_catalog/scan/format.py
+++ b/src/esm_catalog/scan/format.py
@@ -1,0 +1,85 @@
+"""File-format detection for the scan layer.
+
+A file's format is resolved by extension first, then a magic-byte sniff — so
+extension-less ESM model output (ECHAM/FESOM write files like ``expid_200001.01``)
+is still recognised. The returned ``FileFormat`` selects the reader (see
+``reader.py``); detection never opens the file with a heavy library.
+"""
+
+from __future__ import annotations
+
+from enum import auto
+
+from upath import UPath
+
+from esm_catalog._compat import StrEnum
+
+
+class FileFormat(StrEnum):
+    """A scannable file format, as resolved by :func:`detect`."""
+
+    netcdf = auto()
+    grib = auto()
+
+
+class UnknownFormatError(ValueError):
+    """Raised when a path matches no known format by extension or magic bytes."""
+
+
+_NETCDF_SUFFIXES = frozenset({".nc", ".nc4", ".cdf", ".netcdf"})
+"""Filename suffixes that unambiguously mean NetCDF."""
+
+_GRIB_SUFFIXES = frozenset({".grb", ".grb2", ".grib", ".grib2"})
+"""Filename suffixes that unambiguously mean GRIB."""
+
+_GRIB_MAGIC = b"GRIB"
+"""GRIB1/GRIB2 signature at offset 0."""
+
+_HDF5_MAGIC = b"\x89HDF"
+"""NetCDF-4 / HDF5 signature at offset 0."""
+
+_CDF_MAGIC = b"CDF"
+"""Classic NetCDF signature (CDF\\x01 / \\x02 / \\x05) at offset 0."""
+
+
+def detect(path: UPath) -> FileFormat:
+    """Resolve the :class:`FileFormat` of *path*.
+
+    Extension is tried first (cheap, and authoritative when present); otherwise
+    the first four bytes are sniffed, which catches the extension-less files ESM
+    models emit.
+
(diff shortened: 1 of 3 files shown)


src/esm_catalog/scan/format.py
```diff
@@ -0,0 +1,85 @@
...
+
+class UnknownFormatError(ValueError):
+    """Raised when a path matches no known format by extension or magic bytes."""
+
+
+_NETCDF_SUFFIXES = frozenset({".nc", ".nc4", ".cdf", ".netcdf"})
+"""Filename suffixes that unambiguously mean NetCDF."""
+
+_GRIB_SUFFIXES = frozenset({".grb", ".grb2", ".grib", ".grib2"})
+"""Filename suffixes that unambiguously mean GRIB."""
```
pgierz, line 33: I don't like this part necessarily.

src/esm_catalog/scan/format.py
```diff
@@ -0,0 +1,85 @@
...
+_CDF_MAGIC = b"CDF"
+"""Classic NetCDF signature (CDF\\x01 / \\x02 / \\x05) at offset 0."""
+
+
+def detect(path: UPath) -> FileFormat:
+    """Resolve the :class:`FileFormat` of *path*.
+
+    Extension is tried first (cheap, and authoritative when present); otherwise
+    the first four bytes are sniffed, which catches the extension-less files ESM
+    models emit.
```
pgierz, line 50: Reword needed.

pgierz (commented):
Only small comments from my side here. The only thing I don't like it hard-coding extensions in a list in the code...

src/esm_catalog/scan/format.py
```diff
@@ -0,0 +1,85 @@
...
+from upath import UPath
+
+from esm_catalog._compat import StrEnum
+
+
+class FileFormat(StrEnum):
+    """A scannable file format, as resolved by :func:`detect`."""
+
+    netcdf = auto()
+    grib = auto()
```
pgierz, line 22: I'd like to confirm here with @JanStreffing, in particular for OIFS/LPJG -- the LLM only injected the "most common" formats we typically use. However, these two are far from exhaustive. Does anything we produce still use other formats (old FORTRAN service file, raw ASCII text, direct binary dump, xml, ...)

The error message shows you both the detected suffix as well as the magic bytes, so at least it is obvious what it was trying to open.

src/esm_catalog/scan/reader.py
```diff
@@ -0,0 +1,69 @@
...
+from esm_catalog.types import FileMetadata
+
+
+class UnsupportedContentError(Exception):
+    """A file whose format is recognised but whose content a reader cannot handle
+    in the current context -- e.g. NetCDF-3 classic reached over a remote stream,
+    which the HDF5-only backend cannot open (and which is too large to fetch
+    whole). Treated as a clean skip (unsupported), never a read failure; a local
+    scan of the same file, read through the default engine, catalogues it.
+    """
```
pgierz, line 26: Why is this here? I would have enjoyed something more like our `types.py` or `models.py`, and instead have a central place for custom exceptions and error messages...

src/esm_catalog/scan/reader.py
```diff
@@ -0,0 +1,69 @@
...
+    Attributes
+    ----------
+    supports_remote : bool
+        Whether the reader can read a non-local path (a ``UPath`` whose protocol
+        is not ``file``). The walk routes remote paths only to readers that
+        declare ``True``; a local-only reader (e.g. one shelling out to a native
+        library) declares ``False``.
+    """
+
+    supports_remote: bool
```
pgierz, line 42: "Taste": doc-string directly on the attribute, instead of on the class level doc? This is nitpicking though, and does not really matter

src/esm_catalog/scan/reader.py
```diff
@@ -0,0 +1,69 @@
...
+        ...
+
+
+READERS: dict[FileFormat, Reader] = {}
+"""The format -> reader registry; populated by :func:`register` at import time."""
+
+
+def register(file_format: FileFormat, reader: Reader) -> None:
+    """Register *reader* as the handler for *file_format* (replacing any prior)."""
+    READERS[file_format] = reader
```
pgierz, line 66: I would prefer a replacement warning here. It will probably never get triggered, but if something overrides the reader, it should be noisy enough to be obvious.

pgierz (commented):
Only some small things, mostly just style disagreements between myself and the LLM...

PR #1545, opened 2026-08-11
scan: NetCDF reader

pgierz:
Add the NetCDF reader that extracts STAC metadata from a NetCDF file.

### DIFF
--- a/src/esm_catalog/scan/readers/netcdf/__init__.py
+++ b/src/esm_catalog/scan/readers/netcdf/__init__.py
@@ -1,0 +1,99 @@
+"""The NetCDF reader for the scan layer.
+
+Opens a NetCDF file with xarray and extracts the
+:class:`~esm_catalog.types.FileMetadata` a STAC Item is built from -- variables,
+cube dimensions, spatial extent, temporal range, and an inferred frequency.
+Everything is read from the file: coordinates are found by their CF attributes
+(via :mod:`cf_xarray`), never by hard-coded names, and nothing is guessed from
+the filename. The component name, checksum, and the run span an ``fx`` file is
+stamped with are the sourcing layer's job.
+
+The absence of a time coordinate is meaningful: a time-invariant file yields no
+temporal range and no frequency, and that omission is the ``fx`` signal the
+downstream layer keys on -- it also means the item's datetime stamp cannot come
+from the file and must be supplied from the experiment run span.
+
+The implementation is split by concern into submodules (``open``, ``coords``,
+``timeaxis``, ``dimensions``, ``variables``, ``frequency``, ``encoding``). This
+module is the public face: the :class:`NetCDFReader` and its registration.
+``_drop_surrogates`` is re-exported for the encoding test.
+"""
+
+from __future__ import annotations
+
+import cf_xarray  # noqa: F401 - registers the .cf accessor on Dataset/DataArray
+from upath import UPath
+
+from esm_catalog.scan.format import FileFormat
+from esm_catalog.scan.reader import register
+from esm_catalog.types import FileMetadata
+
+from .coords import _extract_bbox
+from .dimensions import _extract_dimensions
+from .encoding import _drop_surrogates
+from .frequency import _infer_frequency
+from .open import _open_dataset
+from .timeaxis import _extract_time_range
+from .variables import _extract_variables
+
+__all__ = ["NetCDFReader"]
+
+
+class NetCDFReader:
+    """Reads :class:`~esm_catalog.types.FileMetadata` from a NetCDF file.
+
+    Attributes
+    ----------
+    supports_remote : bool
+        ``True`` -- remote paths are read through ``h5netcdf`` over fsspec.
+    """
+
+    supports_remote = True
+
+    def read(self, path: UPath) -> FileMetadata:
+        """Open *path* with xarray and extract its scan metadata.
+
+        Parameters
+        ----------
+        path : UPath
+            The NetCDF file to read, local or remote.
+
+        Returns
+        -------
+        FileMetadata
+            Every key the file supports. ``datetime_start``/``datetime_end``,
+            ``frequency`` and ``datetime_str`` are omitted for a time-invariant
+            file -- their absence marks it as ``fx``.
+        """
+        with _open_dataset(path) as opened:
+            # Stamp CF axis/coordinate attributes onto coordinates the model left
+            # bare (regex over common names) so the extractors below find lat, lon,
+            # level and time whether or not the file carried CF attributes -- no
+            # fixed name list, which never stays complete.
+            dataset = opened.cf.guess_coord_axis()
+            variables = _extract_variables(dataset)
+            dimensions = _extract_dimensions(dataset)
+            bbox, geometry = _extract_bbox(dataset)
+            start, end = _extract_time_range(dataset)
+            frequency = _infer_frequency(dataset)
+            primary = next(iter(dataset.data_vars), "unknown")
+
+        metadata: FileMetadata = {
+            "variable": primary,
+            "variables": variables,
+            "dimensions": dimensions,
+            "bbox": bbox,
+            "geometry": geometry,
+            "format": "netcdf",
+        }
+        if start is not None:
+            metadata["datetime_start"] = start
+            metadata["datetime_str"] = start.strftime("%Y%m")
+        if end is not None:
+            metadata["datetime_end"] = end
+        if frequency is not None:
+            metadata["frequency"] = frequency
+        return _drop_surrogates(metadata)
+
+
+register(FileFormat.netcdf, NetCDFReader())
--- a/src/esm_catalog/scan/readers/netcdf/coords.py
+++ b/src/esm_catalog/scan/readers/netcdf/coords.py
@@ -1,0 +1,82 @@
+"""Spatial coordinates: latitude/longitude detection, bounding box, geometry.
+
+Latitude and longitude are located by their CF attributes (``standard_name``,
+``units``, ``axis``) through :mod:`cf_xarray`, not by a fixed list of names -- no
+such list is ever complete. A file whose lat/lon cannot be found or validated
+falls back to the whole-Earth extent.
+"""
+
+from __future__ import annotations
+
+from typing import Optional
+
+import cf_xarray  # noqa: F401 - registers the .cf accessor on Dataset/DataArray
+import numpy as np
+import xarray as xr
+
+from esm_catalog.types import BBox, Geometry
+
+SpatialExtent = tuple[BBox, Geometry]
+"""A bounding box paired with its GeoJSON polygon."""
+
+_GLOBAL_BBOX: BBox = [-180.0, -90.0, 180.0, 90.0]
+"""The whole-Earth fallback extent, used when no valid geographic coords are found."""
+
+
+def _extract_bbox(dataset: xr.Dataset) -> SpatialExtent:
+    """Return the ``(bbox, geometry)`` from latitude/longitude coordinates.
+
+    Falls back to the whole-Earth extent when no geographic coordinates are present
+    or their ranges fail validation.
+    """
+    latitude = _cf_coord(dataset, "latitude")
+    longitude = _cf_coord(dataset, "longitude")
+    if latitude is None or longitude is None:
+        return _global_extent()
+
+    try:
+        lat_min, lat_max = float(np.nanmin(latitude)), float(np.nanmax(latitude))
+        lon_min, lon_max = float(np.nanmin(longitude)), float(np.nanmax(longitude))
+    except (ValueError, TypeError):
+        return _global_extent()
+
+    in_range = -90 <= lat_min <= lat_max <= 90 and -180 <= lon_min <= lon_max <= 180
+    if not in_range:
+        return _global_extent()
+
+    bbox: BBox = [lon_min, lat_min, lon_max, lat_max]
+    return bbox, _bbox_to_polygon(bbox)
+
+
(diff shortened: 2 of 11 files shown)


src/esm_catalog/scan/readers/netcdf.py
```diff
@@ -0,0 +1,478 @@
...
+
+SpatialExtent = tuple[BBox, Geometry]
+"""A bounding box paired with its GeoJSON polygon."""
+
+
+_LATITUDE_NAMES = ("lat", "latitude", "nav_lat", "y", "XLAT")
+"""Coordinate names treated as latitude, most specific first."""
+
+_LONGITUDE_NAMES = ("lon", "longitude", "nav_lon", "x", "XLONG")
+"""Coordinate names treated as longitude, most specific first."""
```
pgierz: Here again, I am unsure if I like that it is hard-coding some names. This list will never be finished

src/esm_catalog/scan/readers/netcdf.py
```diff
@@ -0,0 +1,478 @@
...
+
+    Local paths open straight from the ``UPath`` with xarray's default engine.
+    Remote paths (``memory://``, ``ssh://``, ``scoutfs://``, ``s3://``, ...) are
+    opened through the ``UPath``'s own fsspec filesystem -- which already carries
+    its storage options / credentials -- and handed to ``h5netcdf`` as a file
+    object, because ``h5netcdf`` cannot consume a remote path directly. A dataset
+    with a non-standard calendar or missing time units raises on decode; we then
+    reopen with ``decode_times=False`` and decode the time coordinate by hand
+    downstream. Each attempt opens a fresh handle, since a consumed remote stream
+    cannot be re-read.
```
pgierz: I was under the impression that UPath was able to be used natively, also in xarray? I am unconvinced that `h5netcdf` cannot be remote, that might have been network hiccup while developing on my side.

src/esm_catalog/scan/readers/netcdf.py
```diff
@@ -0,0 +1,478 @@
...
+        store = path.fs.open(path.path, "rb")
+        try:
+            return xr.open_dataset(store, engine="h5netcdf", **open_kwargs)
+        except OSError as error:
+            if _is_signature_error(error):
+                raise UnsupportedContentError(
+                    f"{path.name}: NetCDF-3 classic is not readable over a remote "
+                    "stream; scan it on the cluster (local read) instead"
+                ) from error
+            raise
```
pgierz: not `raise error`?

src/esm_catalog/scan/readers/netcdf.py
pgierz, line 1: There are a few things I need a second pair of eyes on here. A lot of this is manipulating the filename and getting other patterns that might not always be there, so I think we need to consider how much of this part is robust.

I would also consider splitting this into helpers and main functionality, it seems as if this file is very mixed up about it's own separation of concerns...

pgierz (commented):
This needs another pass still

src/esm_catalog/scan/readers/netcdf/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+Opens a NetCDF file with xarray and extracts the
+:class:`~esm_catalog.types.FileMetadata` a STAC Item is built from -- variables,
+cube dimensions, spatial extent, temporal range, and an inferred frequency.
+Everything is read from the file; the component name and checksum are the
+sourcing layer's job.
+
+Local paths use xarray's default engine; remote ``UPath``\\s are read through
+``h5netcdf`` over fsspec (ssh://, scoutfs://, s3://, ...). The absence of a time
+coordinate is meaningful: a time-invariant file yields no temporal range and no
+frequency, and that omission is the ``fx`` signal the downstream layer keys on.
```
pgierz: "The absence of a time coordinate is meaningful"

... yes, but we must be aware, people who write models are not the most tidy of creatures. It might not be called "time". I will need to see how this is determined...

pgierz: Not sure I like the remote part here. I understand why it is interesting to have, however, it adds a level of complexity and guarding we might not need. I also really really dislike that it hard-codes a specific engine for xarray based upon "file is remote or not"

src/esm_catalog/scan/readers/netcdf/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+from esm_catalog.scan.reader import register
+from esm_catalog.scan.readers.netcdf.coords import _extract_bbox
+from esm_catalog.scan.readers.netcdf.dimensions import _extract_dimensions
+from esm_catalog.scan.readers.netcdf.encoding import _drop_surrogates
+from esm_catalog.scan.readers.netcdf.frequency import _infer_frequency
+from esm_catalog.scan.readers.netcdf.naming import _datetime_str
+from esm_catalog.scan.readers.netcdf.open import _open_dataset
+from esm_catalog.scan.readers.netcdf.timeaxis import _extract_time_range
+from esm_catalog.scan.readers.netcdf.variables import _extract_variables
+from esm_catalog.types import FileMetadata
```
pgierz, line 29: Shouldn't all of these be relative imports?

src/esm_catalog/scan/readers/netcdf/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+
+__all__ = ["NetCdfReader"]
+
+
+class NetCdfReader:
```
pgierz: In both: NetCDF not NetCdf. We should at least follow the formal name of the thing... my inner Mr. Monk is protesting.

src/esm_catalog/scan/readers/netcdf/encoding.py
```diff
@@ -0,0 +1,21 @@
...
+    """Recursively strip lone surrogate code points from every string in *value*.
+
+    NetCDF attributes written by Fortran models can carry non-UTF-8 bytes, which
+    xarray surfaces as lone surrogates (``\\udc80``…). Those cannot be encoded to
+    the UTF-8 of a stac-geoparquet shard -- pyarrow raises ``UnicodeEncodeError:
+    surrogates not allowed`` when the item is serialised. Round-trip through
+    UTF-8 with replacement so the metadata is always encodable.
+    """
+    if isinstance(value, str):
+        return value.encode("utf-8", "replace").decode("utf-8")
```
pgierz, line 16: ?????

src/esm_catalog/scan/readers/netcdf/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Attributes
+    ----------
+    supports_remote : bool
+        ``True`` -- remote paths are read through ``h5netcdf`` over fsspec.
+    """
+
+    supports_remote = True
+
+    def read(self, path: UPath) -> FileMetadata:
```
pgierz, line 53: should be static method?

src/esm_catalog/scan/readers/netcdf/coords.py
```diff
@@ -0,0 +1,81 @@
...
+from esm_catalog.types import BBox, Geometry
+
+SpatialExtent = tuple[BBox, Geometry]
+"""A bounding box paired with its GeoJSON polygon."""
+
+_LATITUDE_NAMES = ("lat", "latitude", "nav_lat", "y", "XLAT")
+"""Coordinate names treated as latitude, most specific first."""
+
+_LONGITUDE_NAMES = ("lon", "longitude", "nav_lon", "x", "XLONG")
+"""Coordinate names treated as longitude, most specific first."""
```
pgierz: Not sure I like this -- this list will never be complete.

src/esm_catalog/scan/readers/netcdf/coords.py
```diff
@@ -0,0 +1,81 @@
...
+
+def _global_extent() -> SpatialExtent:
+    """Return the whole-Earth bounding box and its polygon."""
+    return _GLOBAL_BBOX, _bbox_to_polygon(_GLOBAL_BBOX)
+
+
+def _bbox_to_polygon(bbox: BBox) -> Geometry:
+    """Return the closed GeoJSON polygon of a ``[west, south, east, north]`` box."""
+    lon_min, lat_min, lon_max, lat_max = bbox
+    return {
```
pgierz, line 71: It would be perhaps more sensible to actually use:


```suggestion
    return Geometry({
```

src/esm_catalog/scan/readers/netcdf/dimensions.py
```diff
@@ -0,0 +1,50 @@
...
+        if "time" in str(name).lower():
+            entry["type"] = "temporal"
+            entry["extent"] = _time_extent_iso(coord)
+        elif name in _LATITUDE_NAMES:
+            entry["type"] = "spatial"
+            entry["axis"] = "y"
+        elif name in _LONGITUDE_NAMES:
+            entry["type"] = "spatial"
+            entry["axis"] = "x"
+        elif name in ("lev", "level", "depth", "z"):
```
pgierz: this is a mixture -- above we have hard coded constants (also lists/tuples), and then here, we have an inline hard-coded list?

Again, I am not a fan. Models are not tidy enough to hard code expected names...

src/esm_catalog/scan/readers/netcdf/frequency.py
```diff
@@ -0,0 +1,91 @@
...
+        return "subhr"
+    if hours < 2.0:
+        return "1hr"
+    if hours < 4.5:
+        return "3hr"
+    if hours < 12.0:
+        return "6hr"
+    if step_days < 2.0:
+        return "day"
+    return "mon"
```
pgierz: Is there a CF library we can recycle here? Again, I do not think it is so good to do these kinds of cascading checks...

src/esm_catalog/scan/readers/netcdf/frequency.py
```diff
@@ -0,0 +1,91 @@
...
+
+from esm_catalog.scan.readers.netcdf.naming import _frequency_from_filename
+
+
+def _infer_frequency(dataset: xr.Dataset, path: UPath) -> Optional[str]:
+    """Infer a CF-style frequency code (CMIP ``frequency``) for *dataset*.
+
+    The signal is taken, in priority order, from: the global ``frequency``
+    attribute; the step between the first two time values; or the digit length of
+    the filename's numeric stamp.
```
pgierz: Please no. Do not try to take apart the filename to figure out the frequency, that is a horrible idea.

src/esm_catalog/scan/readers/netcdf/frequency.py
```diff
@@ -0,0 +1,91 @@
...
+        step_days = _time_step_days(times, coord)
+        if step_days is not None:
+            return _step_days_to_frequency(step_days)
+
+    return _frequency_from_filename(path)
+
+
+def _time_step_days(times: np.ndarray, coord: xr.DataArray) -> Optional[float]:
+    """Return the step, in days, between the first two values of the time axis."""
+    first, second = times[0], times[1]
```
pgierz: ...and for non-uniform time axes? This is a solution based upon the assumption that this is always uniform...

src/esm_catalog/scan/readers/netcdf/naming.py
pgierz, line 1: The entire idea behind this file is (in my view) not a good approach...

src/esm_catalog/scan/readers/netcdf/open.py
```diff
@@ -0,0 +1,65 @@
...
+        store = path.fs.open(path.path, "rb")
+        try:
+            return xr.open_dataset(store, engine="h5netcdf", **open_kwargs)
+        except OSError as error:
+            if _is_signature_error(error):
+                raise UnsupportedContentError(
+                    f"{path.name}: NetCDF-3 classic is not readable over a remote "
+                    "stream; scan it on the cluster (local read) instead"
+                ) from error
+            raise
```
pgierz, line 46: I'd like an independent check on this, it seems like a block guarding against something we will likely never see in practice.

src/esm_catalog/scan/readers/netcdf/open.py
```diff
@@ -0,0 +1,65 @@
...
+        if _is_time_decode_error(error):
+            logger.debug("Time decode failed for {}; retrying undecoded", path)
+            return _open(decode_times=False)
+        raise
+
+
+def _is_time_decode_error(error: Exception) -> bool:
+    """Return whether *error* is xarray failing to decode the time coordinate."""
+    message = str(error)
+    return "unable to decode time" in message or "Failed to decode" in message
```
pgierz, line 60: I don't like the second part of this, after the `or`. "Failed to decode" can also be happening for other reasons...

src/esm_catalog/scan/readers/netcdf/timeaxis.py
```diff
@@ -0,0 +1,100 @@
...
+
+def _extract_time_range(dataset: xr.Dataset) -> DateTimeRange:
+    """Return the ``(start, end)`` datetimes of the ``time`` coordinate.
+
+    Handles decoded ``datetime64`` values, ``cftime`` objects, and integer-encoded
+    time that xarray could not decode (via the coordinate's CF units and calendar).
+    Returns ``(None, None)`` when there is no time coordinate — the time-invariant
+    case.
+    """
+    coord = dataset.coords.get("time")
```
pgierz: Boooo! `"time"` assumes clean model names. Not always the case (Old FESOM used `T` for example)

pgierz (commented):
This one still needs some work, I think...

pgierz (commented):
approve

PR #1547, opened 2026-08-11
chore: consolidate StrEnum shim, Optional annotations, style doc

pgierz:
Move the StrEnum shim into _compat, switch remaining annotations to Optional[...], and add the esm_catalog style doc.

### DIFF
--- a/src/esm_catalog/paleo.py
+++ b/src/esm_catalog/paleo.py
@@ -27,6 +27,8 @@ The values come from the ``general.paleo`` config section (see
 
 from __future__ import annotations
 
+from typing import Optional
+
 import pystac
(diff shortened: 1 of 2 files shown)


src/esm_catalog/STYLE.md
pgierz, line 1: wtf is this? No, bad LLM. Do not put markdown notes in our repo!

pgierz (commented):
Tiny PR, but, adds things we do not need. Requires another iteration...

pgierz (commented):
I cannot self-approve, but I understand this code based upon the final version that I am reading through. For stacking purposes, this can go in.

PR #1548, opened 2026-08-11
scan: STAC item fx + run span

pgierz:
Model a time-invariant (fx) file across the experiment run span, marked frequency=fx.


pgierz:
I have read and understand this portion, it can go in to the stack.

PR #1549, opened 2026-08-11
scan: geoparquet shards

pgierz:
Write scanned STAC items into stac-geoparquet shards.

### DIFF
--- a/src/esm_catalog/storage/geoparquet.py
+++ b/src/esm_catalog/storage/geoparquet.py
@@ -1,0 +1,145 @@
+"""stac-geoparquet shard storage for the catalog.
+
+A scanned experiment is persisted as a set of `stac-geoparquet
+<https://github.com/stac-utils/stac-geoparquet>`_ shards rather than a database.
+Two kinds of shard live side by side:
+
+- ``<expid>_stac_fx.parquet`` holds the time-invariant Items and is *rewritten
+  in full* on every scan, pinned to the current run span. There is exactly one
+  per catalog.
+- ``<expid>_stac_<run_stamp>.parquet`` shards hold the per-run-segment
+  time-series Items and are *append-only*: each finished run segment writes its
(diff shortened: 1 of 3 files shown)


src/esm_catalog/storage/geoparquet.py
```diff
@@ -0,0 +1,122 @@
...
+import pyarrow as pa
+import pyarrow.parquet as pq
+import pystac
+from stac_geoparquet.arrow import parse_stac_items_to_arrow, to_parquet
+from upath import UPath
+
+from esm_catalog.scan.types import RunStamp
+
+FX_SHARD: str = "fx.parquet"
+"""Filename of the single time-invariant shard, rewritten in full each scan."""
```
pgierz: Is this used anywhere?

src/esm_catalog/storage/geoparquet.py
```diff
@@ -0,0 +1,122 @@
...
+    ----------
+    run_stamp : RunStamp
+        The run-segment stamp, e.g. ``'20000101-20001231'``.
+
+    Returns
+    -------
+    str
+        The shard filename, e.g. ``'ts-20000101-20001231.parquet'``.
+    """
+    return f"ts-{run_stamp}.parquet"
```
pgierz: I am unsure if I like this file being called `ts`. For me it would make more sense to have this somehow as `{expid}_stac_{run_stamp}.parquet` or similar. I am not sure if we should use the "stac" part as "component" in the middle of the file, that can be discussed still...

src/esm_catalog/storage/geoparquet.py
pgierz, line 1: This file assumes a lot of internal knowledge of how pyarrow works, which I do not have.

pgierz (commented):
OK, I understand this one, can go in.

PR #1550, opened 2026-08-11
scan: experiment metadata + output-file sourcing

pgierz:
Source experiment metadata and output-file paths from the run finished_config, with a production-requirements schema.

### DIFF
--- a/src/esm_catalog/scan/sourcing.py
+++ b/src/esm_catalog/scan/sourcing.py
@@ -1,0 +1,572 @@
+"""Source experiment metadata and output-file paths from an ESM-Tools run.
+
+The catalog is built from what an ESM-Tools run *declared*, not from what a
+filesystem walk happens to find. Every completed run writes a ``finished_config``
+(one per run segment) under ``<exp_root>/config/`` recording the experiment's
+identity, its scientific metadata, and the absolute target path of every output
+file. This module reads those configs and turns them into the scan layer's fixed
+contracts -- :class:`~esm_catalog.models.ExperimentMetadata` and
+:class:`~esm_catalog.scan.types.OutputFile` -- so the config is the single,
+authoritative discovery path.
+
+Assumptions about the (loosely specified) finished_config schema, all handled
+tolerantly with ``.get`` and defaults:
+
+* ``general.expid`` holds the experiment id; the filename prefix is the fallback.
+* A run segment's window comes from ``general`` run dates (``start_date`` /
+  ``end_date``, or ``run_datestamp``), falling back to the ``YYYYMMDD-YYYYMMDD``
+  suffix ESM-Tools appends to the filename. ``run_start`` / ``run_end`` on the
+  returned metadata are the min / max across all segments (their union).
+* Experiment metadata lives in ``general.metadata``: ``Description``, ``Authors``
+  (-> :class:`Contact` list) and ``License``. A component's own ``metadata``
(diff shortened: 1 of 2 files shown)


pgierz:
This one needs work before we can review it -- it seems to have put a lot of the legible types directly in the code, instead of separately.

src/esm_catalog/scan/sourcing.py
```diff
@@ -0,0 +1,572 @@
...
+  ``end_date``, or ``run_datestamp``), falling back to the ``YYYYMMDD-YYYYMMDD``
+  suffix ESM-Tools appends to the filename. ``run_start`` / ``run_end`` on the
+  returned metadata are the min / max across all segments (their union).
+* Experiment metadata lives in ``general.metadata``: ``Description``, ``Authors``
+  (-> :class:`Contact` list) and ``License``. A component's own ``metadata``
+  describes the model, not the experiment, and is not read.
+* The concrete output files (and their MD5s) come from the tidy-phase
+  file-operations logs, the authoritative record of what the run produced. Each
+  component's ``outdata_targets`` mapping holds glob *patterns*, used only as a
+  fallback when tidy logs are absent.
```
pgierz, line 26: This part shall be replaced by @mandresm's work with `file-dicts`

src/esm_catalog/scan/sourcing.py
```diff
@@ -0,0 +1,572 @@
...
+    model_config = ConfigDict(extra="allow")
+
+    expid: Any = None
+    start_date: Any = None
+    initial_date: Any = None
+    current_date: Any = None
+    end_date: Any = None
+    final_date: Any = None
+    run_datestamp: Any = None
+    paleo: Any = None
```
pgierz, line 78: I'm not sure I like this being here, it belongs as a plugin part.

src/esm_catalog/scan/sourcing.py
pgierz, line 1: A lot about this file I don't really like, but it will have to do for now. 

@mandresm, Just a thought: Long long down the road, when we have proper objects everywhere, I would ideally want to have something like:

```python
>>> path_to_a_yaml = ...
>>> sim = esm_tools.load_sim_from_config(UPath(path_to_a_yaml), thin=True))  # Thin true/false for bootstrapping a config with fake dates or whatever it is that esm-master really does
>>> sim.start_date, sim.end_date
<esm_tools.SimulationDate 1850-01-01 12:00:00>, <esm_tools.SimulationDate 1850-01-31 23:59:30>
>>> sim.files
[ esm_tools.SimulationFile, ... ]
>>> sim.files[0]
{ ... } # file dict representation
```

src/esm_catalog/scan/sourcing.py
```diff
@@ -0,0 +1,572 @@
...
+    return None
+
+
+def _paleo_config(docs: Iterable[FinishedConfigDoc]) -> Optional[PaleoConfig]:
+    """The ``general.paleo`` section, from the first segment that declares one."""
+    for doc in docs:
+        paleo = _general(doc).paleo
+        if isinstance(paleo, dict) and paleo:
+            return paleo
+    return None
```
pgierz, line 409: This should be pluggable, not here...but ok for now.

pgierz (commented):
OK, I understand this code. It is not really a solution that I want long-term (for that we need file-dicts), but now it will do the job.

PR #1551, opened 2026-08-11
scan: incremental workspace

pgierz:
Track catalog workspace state so a rescan only processes new files.

### DIFF
--- a/src/esm_catalog/scan/workspace.py
+++ b/src/esm_catalog/scan/workspace.py
@@ -1,0 +1,62 @@
+"""The CLI workspace state at ``<exp_root>/catalog/esm-catalog.json``.
+
+``scanned`` maps a time-varying file's path to the md5 it was catalogued at, so
+a re-scan skips it when unchanged. Time-invariant (fx) files are deliberately
+never recorded here — they are re-read every scan (cheap, there are few) because
+their datetime is the experiment run span, which grows when the run is extended,
+so ``fx.parquet`` is rewritten each scan.
+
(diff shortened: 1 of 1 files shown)


src/esm_catalog/scan/workspace.py
```diff
@@ -0,0 +1,59 @@
...
+from upath import UPath
+
+from esm_catalog.scan.types import Md5
+from esm_catalog.types import ExperimentId
+
+CATALOG_DIRNAME = "catalog"
+"""The catalog subdirectory created under an experiment root."""
+
+STATE_FILENAME = "esm-catalog.json"
+"""The workspace-state file inside the catalog directory."""
```
pgierz, line 27: This should probably have something like the experiment name stamped into it?

src/esm_catalog/scan/workspace.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class WorkspaceState(BaseModel):
+    """The persisted ``esm-catalog.json`` — init config plus scan bookkeeping."""
+
+    experiment_id: ExperimentId
+    server_url: Optional[str] = None
+    scanned: dict[str, Md5] = {}
```
pgierz: Not sure I like this type; what is src? I suppose one particular output file, right?

pgierz (commented):
This one is fairly straightforward, just some minor comments.

pgierz (commented):
Alright, I understand this. It's just a state tracking to speed up repeat scans. We could argue about the internal format of the thing, but for how: good enough.

PR #1552, opened 2026-08-11
scan: parallel executor

pgierz:
Run the per-file scan across a process pool or a Dask cluster.

### DIFF
--- a/src/esm_catalog/scan/parallel.py
+++ b/src/esm_catalog/scan/parallel.py
@@ -1,0 +1,205 @@
+"""Order-preserving parallel map with a process-pool or Dask backend.
+
+The scan pipeline is embarrassingly parallel: each input is handled
+independently, and the results are collected back in input order. This module
+provides one function, :func:`parallel_map`, that fans a *worker* out over a
+list of *inputs* and returns the results in the **same order** as the inputs
+(deterministic — never as-completed).
+
+Two backends live behind that single function:
+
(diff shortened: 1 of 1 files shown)


src/esm_catalog/scan/parallel.py
```diff
@@ -0,0 +1,205 @@
+"""Order-preserving parallel map with a process-pool or Dask backend.
+
+The scan pipeline is embarrassingly parallel: each input is handled
+independently, and the results are collected back in input order. This module
+provides one function, :func:`parallel_map`, that fans a *worker* out over a
+list of *inputs* and returns the results in the **same order** as the inputs
+(deterministic — never as-completed).
```
pgierz: why is order important?

src/esm_catalog/scan/parallel.py
```diff
@@ -0,0 +1,205 @@
...
+Errors are **not** swallowed. If the worker raises for any input, that
+exception propagates out of :func:`parallel_map` (the process-pool backend
+re-raises it when the result is collected; the Dask backend re-raises it on
+gather). A partial run therefore fails loudly rather than returning a truncated
+or sentinel-filled list.
+"""
+
+from __future__ import annotations
+
+import logging
```
pgierz: No, use loguru.

src/esm_catalog/scan/parallel.py
pgierz: OK, looks good. One can connect an externally running dask cluster as well, that is nice to have.

PR #1553, opened 2026-08-11
scan: ingest orchestration

pgierz:
Orchestrate sourcing, parallel reading, and shard writing into one scan run.


src/esm_catalog/scan/ingest.py
pgierz, line 1: OK, this can go in.

PR #1554, opened 2026-08-11
cli: scan command

pgierz:
Wire the scan command with a progress display over local or remote roots.


src/esm_catalog/cli.py
pgierz, line 1: OK, this is the high level command wrapper, this looks clean. I could deconstruct and extend this if needed without too much effort. OK to go in.

PR #1568, opened 2026-09-05
Esm catalog/client auth

pgierz:


src/esm_catalog/client.py
```diff
@@ -0,0 +1,160 @@
...
+                f"{action} — server sent a redirect to {location}; check that "
+                "server_url uses https (not http) and has no trailing slash",
+                resp.status_code,
+                resp.text,
+            )
+        raise StacClientError(action, resp.status_code, resp.text)
+
+    def _upsert(self, kind: str, body: StacObject, post_path: str, put_path: str) -> None:
+        resp = self._client.post(self._base + post_path, json=body)
+        if resp.status_code == 409:
```
pgierz, line 135: which magical number is this?!

pgierz (commented):
OK, I understand this code as-is too. There are some minor improvement suggestions that can be batched together at the end, otherwise, this can go in.

PR #1569, opened 2026-09-05
esm-catalog/grib-reader

pgierz:
GRIB reader (basic + pluggable ECHAM enricher), 0-360 grid bbox + scalar time fix, per-server token scoping, push DEBUG/--jobs/error-text fixes.


pgierz (commented):
ok, looks good.

PR #1572, opened 2026-09-05
esm-catalog/fix-1564-datacube-elem-index

### LINKED ISSUE #1564: [AI-Self-Help] FESOM unstructured grid represented as a spatial elem index in cube:dimensions
**AI-Self-Help** — Claude, end-to-end catalog testing. Tracking note.

## Observation
For FESOM output, `cube:dimensions` emits `elem` as a **spatial** dimension with an integer index extent `[0, 244658]` (mesh element count), while the item bbox is the global default. FESOM is unstructured — no clean x/y axis for a datacube.

## Decision needed (not a clear bug)
How should an unstructured grid be represented in the datacube extension?
- Keep `elem` but not typed `spatial` (it is an index, not a geographic axis)?
- Omit spatial datacube dims for unstructured output, rely on geometry/bbox?
- Reference the mesh separately?

Filed for discussion.

pgierz:
Fixes #1564: stop labelling unidentified cube dimensions 'spatial'.

### DIFF
--- a/src/esm_catalog/scan/readers/netcdf/dimensions.py
+++ b/src/esm_catalog/scan/readers/netcdf/dimensions.py
@@ -24,7 +24,10 @@ def _extract_dimensions(dataset: xr.Dataset) -> CubeDimensions:
     dimensions: CubeDimensions = {}
     for name in dataset.sizes:
         coord = dataset.coords.get(name)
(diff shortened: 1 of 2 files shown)


pgierz (commented):
approve

Issue #1531, opened 2026-08-04
ESM Catalog Scan --> Shard


pgierz:

pgierz:
Potentially of interest, from the LLM:
```
❯ xstac? explain.

⏺ xstac = generate STAC metadata from xarray datasets. Bridges NetCDF/Zarr world → STAC.

  What it does:
  - Take xarray Dataset (or Zarr store)
  - Introspect dims/coords/attrs
  - Emit STAC Item with datacube extension filled in — cube:dimensions (x, y, time, level...) and cube:variables (each data var, units, type)

  Why care: hand-writing datacube JSON for multidim climate output = misery. xstac reads the structure you already have and writes the metadata. That's the exact painful part of "STAC for
  models."

  Rough shape:
  import xarray as ds_open
  import xstac

  ds = ds_open.open_zarr("s3://.../run.zarr")

  item_template = {
      "type": "Feature",
      "stac_version": "1.0.0",
      "id": "my-run",
      "properties": {"datetime": None},
      "geometry": {...},   # bbox footprint
      "bbox": [...],
      "assets": {"data": {"href": "s3://.../run.zarr", "roles": ["data"]}},
      "links": [],
  }

  item = xstac.xarray_to_stac(
      ds,
      item_template,
      temporal_dimension="time",
      x_dimension="lon",
      y_dimension="lat",
      reference_system=4326,
  )

  Origin: Microsoft Planetary Computer / stac-utils ecosystem. Used to catalog big Zarr/NetCDF collections (CMIP6, TerraClimate live in Planetary Computer this way).

  Caveats:
  - API surface small, arg names (temporal_dimension etc.) may have shifted — check current docs before wiring
  - Handles the metadata generation, not the serving. Still need pystac/stac-fastapi to store + expose
  - Non-standard grids (rotated pole, unstructured/ICON, curvilinear) — datacube model assumes regular-ish axes. Your FESOM/unstructured stuff may not map clean

  For AWI: regular-grid atmosphere output maps fine. Unstructured ocean (FESOM) = friction, verify before committing.

  Want minimal end-to-end: Zarr → xstac → pystac Collection → served?
```

PR #1575, opened 2026-09-05
refactor(esm_catalog): one type system — pydantic for the record voca…

pgierz:
…bulary

Draft (parked for later integration). Unifies the domain vocabulary on pydantic instead of the TypedDict/pydantic split: the records a scanner produces or a config supplies are now pydantic BaseModels, validated at the boundary where they enter the program.

- types.py: FileMetadata, ScannedVariable -> BaseModel (extra="allow", all fields optional — mirrors the old total=False; borrows the stac-pydantic ItemProperties(extra="allow") idiom).
- paleo.py: PaleoConfig -> BaseModel.
- item.py / datacube.py: consumers switch from fm.get("x") to attribute access.
- datacube/paleo entry points coerce their input via model_validate, so a dict or a model both work (validate-at-boundary); item.py passes real models.
- ExperimentMetadata was already pydantic.
- Benchmark that motivated the old split was bogus: 393k FileMetadata objects cost pydantic ~+1.1s / +315MB total vs plain dicts — noise against an I/O-bound scan; the 393k figure was one large example run, not typical.

Deliberately NOT converted (open call for review): CubeVariable and StacContact are emitted STAC-JSON shapes (serialization output written into item.properties / to_stac()), not ingested records, so they stay TypedDict. Decide later whether to model these too.

Foundation only. The scan readers on esm-catalog/cli-scan build FileMetadata as plain dicts and return them; when this lands there, read() must construct FileMetadata(...) (model_validate already accepts their dict, so the change is small). 423 catalog tests green.

Refs: #1531
Specified-by: Paul Gierz <paul.gierz@awi.de>
Generated-by: Claude Opus 4.8 <noreply@anthropic.com>

- - - - -

I have read and understand this code, it can be merged. 

- - - - -

### DIFF
--- a/src/esm_catalog/datacube.py
+++ b/src/esm_catalog/datacube.py
@@ -58,13 +58,14 @@ def add_datacube_item_extension(item: pystac.Item, file_metadata: FileMetadata) 
     file_metadata : FileMetadata
         The file's scanned metadata. No-op when it carries no dimensions.
     """
-    dimensions = file_metadata.get("dimensions", {})
+    file_metadata = FileMetadata.model_validate(file_metadata)
+    dimensions = file_metadata.dimensions
     if not dimensions:
         return
 
     item.properties["cube:dimensions"] = dimensions
 
-    cube_variables = _to_cube_variables(file_metadata.get("variables", []))
+    cube_variables = _to_cube_variables(file_metadata.variables)
     if cube_variables:
         item.properties["cube:variables"] = cube_variables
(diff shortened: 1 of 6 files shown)


pgierz:
Duplicate of #1574 — same head branch (esm-catalog/types-pydantic) opened against the wrong base (client-auth, which this branch doesn't descend from). #1574 targets release correctly; closing this one.

PR #1579, opened 2026-09-07
Add climberx component (CLIMBER-X / AWI climber-y variant)

pgierz:
## Summary
- Adds a standalone (`climber-clim` flavor) esm-tools component for CLIMBER-X, the PIK EMIC (github.com/cxesmc/climber-x), plus an AWI-internal variant `climber-y` (gitlab.awi.de/paleodyn/models/climber-y) as an alternate `version:`.
- esm_master clones climberx itself via `git-repository`/`branch` per version (same convention as every other component). `comp_command` then drives climber-x's own `configme` tool to clone+configure+build the dependency stack (fesm-utils, yelmo) against that checkout, and `make <flavor>`.
- Namelist and restart file names (`control.nml`, `atm_par.nml`, ..., `atm_restart.nc`, ...) were taken from a real local `configme install climber-x` + repo inspection, not guessed from docs.
- Machine/compiler specifics are gated behind `choose_computer.name: albedo:` per review feedback.

## Known gaps / hazards (flagged inline, confirmed against real Albedo runs)
- **Blocker**: `lnd_alloc()` in `src/lnd/lnd_model.f90` leaves ~328 allocatable arrays uninitialized, crashing the netCDF writer around model year ~1000. Cost two ~2.5h runs on Albedo before being identified. Needs a source patch (compiler flags don't reach derived-type allocatable components). The patch generator that fixed this lived at a personal scratch path on Albedo and is gone before it could be vendored here — exact repro location and a `comp_command` hook for the eventual patch are documented inline.
- **Hazard**: SCRIP map generation (`runme --gen-maps`) is not concurrency-safe — a race on a fresh checkout corrupted a map file and killed a run. Documented as a hard warning against parallel/ensemble installs before maps exist.
- `climber-y` currently has no build system committed upstream (src/*.f90 only) — its `comp_command` will fail until that's fixed AWI-side.
- `input_dir`/`maps` forcing-data wiring depends on the separate `climber-x-input` data repo and a real `pool_dir` path on Albedo — not filled in yet (data exists on Albedo already, ~12GB, path TBD).
- Only `climber-clim` (climate-only, no bgc/ice-sheet) is wired; the other 3 flavors need the private `bgc`/`vilma` components.
- Confirmed the `awi_albedo` configme machine profile does not work with `-c gfortran` (only tested working: `-m linux -c gfortran` with `gcc/12.1.0` + `netcdf-fortran/4.5.4-gcc12.1.0`, and even then `ncdump` isn't on PATH). Not a blocker since we use `-c ifx`, documented to save the next person the trial-and-error.
- No `configs/couplings/` entry yet — standalone only, per current scope.

## Test plan
- [x] YAML syntax validated with `yaml.safe_load`
- [x] Namelist/restart/build info verified against a real `configme install climber-x -m macbook -c gfortran` run (climate-x cloned + fesm-utils/yelmo dep stack cloned; fesm-utils build itself hit an unrelated macOS/arm64 toolchain issue, irrelevant to the Albedo target)
- [x] Real-world build/run findings from Albedo (lnd_alloc bug, map-gen race, gfortran profile mismatch) folded in
- [ ] Not yet tested with `esm_master install-climberx` end-to-end on Albedo
- [ ] lnd_alloc() patch still needs to be regenerated and vendored before this component can run past ~model year 1000

### DIFF
--- a/configs/components/climberx/climberx.yaml
+++ b/configs/components/climberx/climberx.yaml
@@ -212,5 +212,6 @@ computer:
                 - load intel-oneapi-compilers/2024.0.0
                 - load netcdf-c/4.8.1-openmpi4.1.3-oneapi2022.1.0
                 - load netcdf-fortran/4.5.4-oneapi2022.1.0
-                - load udunits/2.2.28
                 - load python/3.10.4
+                # udunits/2.2.28 already loaded by Albedo's own machine
+                # defaults (configs/machines/albedo.yaml), no need to repeat it
--- a/configs/components/climberx/climberx.yaml
+++ b/configs/components/climberx/climberx.yaml
@@ -1,0 +1,196 @@
+climberx:
+    model: climberx
+
+    metadata:
+        Institute: Potsdam Institute for Climate Impact Research (PIK) / AWI Paleodynamics
+        Description: >-
+            Earth system Model of Intermediate Complexity (EMIC): semi-empirical
+            statistical-dynamical atmosphere (SESAM), 3-D frictional-geostrophic
+            ocean (GOLDSTEIN), dynamic-thermodynamic sea ice (SISIM), land model
+            with dynamic vegetation (PALADYN). Optional ocean biogeochemistry
+            (HAMOCC, private), ice sheets (Yelmo/SICOPOLIS) and solid-Earth
+            (VILMA, private) components are not covered by this standalone,
+            climate-only ("climber-clim") integration.
+        Authors: Matteo Willeit, Andrey Ganopolski, Neil R. Edwards, Alexander Robinson, Reinhard Calov, Ralf Greve
+        Publications:
+            - "Willeit et al. 2022, GMD 15, 5905-5948, https://doi.org/10.5194/gmd-15-5905-2022"
+            - "Willeit et al. 2023, GMD 16, 3501-3534, https://doi.org/10.5194/gmd-16-3501-2023"
+        License: GPLv3 (climber-x); climber-y is an AWI-internal variant, license per AWI Paleodynamics group
+
+    # `configme` (https://github.com/fesmc/configme) owns cloning/configuring the
+    # full dependency stack (fesm-utils: LIS+FFTW+utils, yelmo) for climber-x, so
+    # esm_master's own git-repository/branch clone is intentionally NOT used here.
+    # Instead comp_command drives configme itself against model_dir.
+    comp_command: |
+        pip install --quiet --upgrade "git+https://github.com/fesmc/configme"
+        configme install climber-x -m ${climberx_configme_machine} -c ${climberx_compiler} --dir ${model_dir} --overwrite
+        cd ${model_dir}
+        make clean
+        make ${climberx_flavor}
+    clean_command: |
+        cd ${model_dir}
+        make clean
+    install_bins: "."
+    executable: climber.x
+    nproc: 1
+
+    version: x
+    available_versions:
+        - x
+        - "1.3.0"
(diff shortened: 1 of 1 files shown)


configs/components/climberx/climberx.yaml
```diff
@@ -0,0 +1,196 @@
...
+    # - needs a real pool_dir location on Albedo before outdata_files/
+    # input_files can be filled in; not addressed by this standalone-build pass.
+
+computer:
+    add_module_actions:
+        - load intel-oneapi-compilers/2024.0.0
+        - load netcdf-c/4.8.1-openmpi4.1.3-oneapi2022.1.0
+        - load netcdf-fortran/4.5.4-oneapi2022.1.0
+        - load udunits/2.2.28
+        - load python/3.10.4
```
pgierz: This part belongs behind a choose albedo block.

configs/components/climberx/climberx.yaml
```diff
@@ -0,0 +1,196 @@
...
+    # Matches the default standalone scope agreed for this integration; the
+    # other three flavors (climber-clim-bgc / -ice / -bgc-ice) require the
+    # private bgc/vilma components and are out of scope for now.
+    climberx_flavor: climber-clim
+    # Albedo is the only esm-tools machine target for now. configme's machine
+    # naming doesn't match esm-tools' computer.name (esm-tools: "albedo" vs
+    # configme: "awi_albedo"), hence the separate variable instead of
+    # interpolating ${computer.name} directly.
+    climberx_configme_machine: awi_albedo
+    climberx_compiler: ifx
```
pgierz: Here, a choose computer block needs to wrap this...

configs/components/climberx/climberx.yaml
```diff
@@ -0,0 +1,196 @@
...
+    # `configme` (https://github.com/fesmc/configme) owns cloning/configuring the
+    # full dependency stack (fesm-utils: LIS+FFTW+utils, yelmo) for climber-x, so
+    # esm_master's own git-repository/branch clone is intentionally NOT used here.
+    # Instead comp_command drives configme itself against model_dir.
+    comp_command: |
+        pip install --quiet --upgrade "git+https://github.com/fesmc/configme"
+        configme install climber-x -m ${climberx_configme_machine} -c ${climberx_compiler} --dir ${model_dir} --overwrite
+        cd ${model_dir}
+        make clean
+        make ${climberx_flavor}
```
pgierz: For the sake of "cleanliness", I am only sort-of ok with this. It would be good if we could separate out the clone and comp like we do for every other model...

pgierz (commented):
Some comments on the AI's first attempt.

configs/components/climberx/climberx.yaml
pgierz: Fixed — module loads now wrapped in `choose_computer.name: albedo:`.

configs/components/climberx/climberx.yaml
pgierz: Fixed — `climberx_configme_machine`/`climberx_compiler` now wrapped in `choose_computer.name: albedo:`.

configs/components/climberx/climberx.yaml
pgierz: Split it: `git-repository`/`branch` per `choose_version` now does the actual clone via esm_master, same as every other component. `comp_command`'s `configme install` call is left only to configure+build the dependency stack (fesm-utils/yelmo) against that already-present checkout — it no longer re-clones climberx.

PR #1580, opened 2026-09-07
esm-catalog/public-oidc-client

pgierz:
Drops \`client_secret\` from the OIDC client — \`esm-catalog-dev\` is registered as a **public** client on Helmholtz ID (confirmed with HIFIS support, 2026-09-07). PKCE alone authenticates the authorization-code exchange; a secret was never needed, and distributing one to every CLI user's machine defeated the point of using PKCE at all.

Verified empirically before making the change: POSTed to the token endpoint with \`client_id\` only (no secret) and a deliberately-invalid auth code — got \`invalid_grant\` ("wrong code"), not \`invalid_client\`/\`unauthorized_client\`. Confirms the IdP accepts client-auth-free requests for this \`client_id\`.

- \`auth.py\`: \`_request_token\` sends \`client_id\` in the body, drops HTTP Basic auth entirely
- \`config.py\`: \`Settings.client_secret\` removed
- \`cli.py\`: docstring/epilog mentions updated
- \`test_auth.py\`: removed the now-meaningless secret-not-leaked test

Full suite run against this + against unmodified \`prep-release\` for baseline comparison: same 5 pre-existing failures both sides (macOS \`platformdirs\`/XDG env-var quirk, unrelated), zero new failures from this change.

### DIFF
--- a/src/esm_catalog/auth.py
+++ b/src/esm_catalog/auth.py
@@ -199,11 +199,16 @@ def build_login_url(
 def _request_token(
     meta: OIDCMetadata, settings: Settings, data: dict[str, str]
 ) -> TokenSet:
-    """POST to the token endpoint with client-secret-basic auth; return a TokenSet."""
+    """POST to the token endpoint as a public client; return a TokenSet.
+
+    ``esm-catalog-dev`` is registered with Helmholtz AAI as a public client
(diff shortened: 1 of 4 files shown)


pgierz (commented):
ok, this can go in, one secret less to accidentally forget about.

PR #1581, opened 2026-09-07
esm-catalog/suppress-cftime-warning

pgierz:
Suppresses xarray's \`SerializationWarning\` — "Unable to decode time axis into full numpy.datetime64 objects, continuing using cftime.datetime objects instead, reason: dates out of range" — which fires on every paleoclimate file whose dates fall outside \`datetime64[ns]\` range.

It's a warning, not an exception: \`xr.open_dataset(..., decode_times=True)\` succeeds and falls back to cftime objects on its own. Downstream (\`timeaxis.py\`) already handles cftime objects the same as \`datetime64\` (\`_to_utc_datetime\` checks \`hasattr(value, "year")\`), so this was pure noise, not a functional gap — flagged by a human running \`esm-catalog scan\` on a real experiment tree and seeing it repeatedly.

Verified with a synthetic reproduction (a -0500, noleap-calendar dataset) before and after: 0 \`SerializationWarning\`s after the fix, cftime decoding still correct (\`cftime.DatetimeNoLeap\` objects, as expected).

Separate concern from #1580 (client_secret removal), split into its own branch/PR per house convention.

### DIFF
--- a/src/esm_catalog/scan/readers/netcdf/open.py
+++ b/src/esm_catalog/scan/readers/netcdf/open.py
@@ -8,6 +8,8 @@ decoding is reopened with ``decode_times=False`` for hand-decoding downstream.
 
 from __future__ import annotations
 
+import warnings
+
 import xarray as xr
 from loguru import logger
 from upath import UPath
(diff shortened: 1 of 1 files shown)


pgierz (commented):
OK, this can go in.

PR #1561, opened 2026-08-28
FESOM Standalone 2.7.x

pgierz:
Supersedes various PRs 
* #1399
* #1464

PR #1583, opened 2026-09-14
Matrix-parallelize esm_tests CI per (computer, model, runscript)

pgierz:
Splits the esm_tests CI suite so each (computer, model, runscript) triple runs as its own job instead of looping every runscript sequentially inside one job per computer (levante alone: 31 in a row). A discover job enumerates the matrix from each model's config.yaml computers: list; the per-computer reusable workflow now takes model/runscript inputs and scopes test_config.yaml to exactly one runscript before invoking esm_tests. Same flags (-cngeu -s False), same runners, same update_esm_tests_info logic.

Verified: built and pushed as a diagnostic branch off a clean release tip (zero code changes) to confirm the esm_tests CI redness seen on #1561 is not reproducible on release itself -- all 96 checks (68 matrix jobs + existing unit/docs/catalog checks) pass. That also pinpointed #1561's redness: its branch predates recent release commits that introduced config release now expects (e.g. cf_glob_0), so it needs a rebase, not a fix.


pgierz:
Not closing after confirmation, this will serve to restore a green CI base before we go further with the FESOM merges.

pgierz:
As agreed with @mandresm, during his holiday I am merging things I understand and find helpful. This splits and parallelizes esm-tests to be one test per computer/model/runscript combo, and should make the tests online easier to see and go through a bit faster.

PR #1586, opened 2026-09-16
refactor(esm_catalog): make Item/Collection extensions pluggable via pluggy

pgierz:
## Summary
- Add `esm_catalog/plugins.py`: pluggy hookspecs `apply_to_item(item, file_metadata, exp_metadata, hints)` and `apply_to_collection(collection, exp_metadata, hints)`, plus a process-wide cached `PluginManager`.
- `datacube`/`namelist`/`paleo`/`contacts` each register a thin `@hookimpl` wrapper matching whichever `add_*_extension` function(s) they already have (datacube: item-only, contacts: collection-only, namelist/paleo: both).
- `item.py`/`collection.py` no longer hardcode the extension list — they build the base object and fire the hook; `hints` carries the per-scan-hoisted `namelist_props`/`paleo_props`/`validate` values through the hook boundary so the existing hoist-once/validate-once perf optimization survives unchanged.

Prep work for a new `esm-catalog-mcp` server (separate repo) that will register these same extension modules as pluggy plugins to discover MCP tool data — this makes the write-side (Item/Collection building) and future read-side (MCP discovery) share one registration mechanism instead of inventing a second one.

## Test plan
- [x] `pytest tests/test_esm_catalog` — 543 passed, 7 skipped both before and after this change (10 failed / 8 errored pre-existing on `prep-release` HEAD too — stale key-format assertion + env-dependent auth/fixture issues, unrelated to this diff, confirmed via `git stash`)
- [x] Manually verified `make_collection`/`make_item` produce identical output (namelist/paleo/datacube/contacts fields) before and after the refactor

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/esm_catalog/collection.py
```diff
@@ -4,10 +4,8 @@
 
 from pystac import Collection, Extent, Item, SpatialExtent, TemporalExtent
 
-from esm_catalog.contacts import add_contacts_collection_extension
 from esm_catalog.models import ExperimentMetadata
-from esm_catalog.namelist import add_namelist_collection_extension
-from esm_catalog.paleo import add_paleo_collection_extension
```
pgierz, line 10: This is the right way to do this, yes. use a plugin manager, don't hand-import it various times.

pgierz (commented):
OK, this is one of the "human-steered" changes I wanted before declaring this finished. This is arguably the better long term architecture, this can go in from my side.

PR #1589, opened 2026-09-17
FESOM standalone 2.8.0

JanStreffing:
Adds FESOM 2.8.0 as a standalone component, following the layout of the 2.7.x standalone support in #1561.

## What is here

- `configs/components/fesom/fesom-2.8.yaml`, a new file. Based on the 2.7 config, with the version block reduced to `2.8`, `2.8.0`, `2.8-main` and `2.8.0-xios`. recom is a git submodule of fesom2 (`src/recom`), so every version clones with `--recurse-submodules`.
- `namelists/fesom2/2.8.0/`, a verbatim copy of `config/` at the 2.8.0 tag, as `namelists/fesom2/2.7.0/` is of the 2.7.0 tag.
- `runscripts/fesom2/fesom2.8-levante-{1d,reference}.yaml`, the 2.7 runscripts with the version and `model_dir` moved to 2.8.
- Optional XIOS output, off by default.

Nothing outside `fesom-2.8.yaml`, `namelists/fesom2/2.8.0/` and the two new runscripts is touched, so this is purely additive.

## Build

The `comp_command` drops `-DASYNCHRONOUS_IO_THREADS=ON`. In 2.8.0 that option only exists under `dwarf/`; it is not an option of the main build, so the 2.7 config has been passing a no-op.

New in 2.8.0 and left at the upstream defaults: the `io_parallel` group, `ForcingDataPath` / `RestartInPath` / `RestartOutPath` in `paths`, and the hosing switches in `run_config`. `ForcingDataPath` is inert here because the forcing file names the config writes are absolute, which 2.8.0 passes through unchanged.

## XIOS

`fesom-2.8.0-xios` builds the same source with `-DFESOM_WITH_XIOS=ON`. Plain `2.8` and `2.8.0` need no XIOS at all, and `with_xios` defaults to false, so the XML staging never fires unless it is asked for.

XIOS runs in attached mode: `using_server=false`, so the server lives inside the FESOM ranks. The job stays a single executable with no `xios_server.exe` ranks to reserve, and each rank writes its own file, named `VARIABLE.fesom_STARTYEAR-ENDYEAR_RANK.nc`.

Four of the six XMLs come from the model's own `docs/xios_xml/`, so they track the source that is built. Two are shipped here instead:

- `iodef.xml`, because the bundled one is the AWI-ESM3 coupled variant: it pulls in the OpenIFS context and sets `using_server` and `using_oasis` true.
- `file_def_fesom.xml`, because the bundled one asks for six fields on `grid_2d_elem` (`tx_sur`, `ty_sur`, `strength_ice`, `sgm11`, `sgm12`, `sgm22`) that abort the run in `cxios_write_data`:

  ```
  grid.hpp, CGrid::maskField, line 440 ->
  [ Awaiting data of size = 1958, Received data size = 990 ]
  The data array does not have the right size! Grid = grid_2d_elem
  ```

  The element domain is not the problem. `u` and `v` on `grid_3d_elem` are in the upstream `.github/workflows/fesom2_xios.yml` test and write correctly. It is the 2D element send path that hands XIOS the wrong extent. Worth an issue against FESOM.

XIOS has to be built without OASIS, so one of the `_ogcm` or `_oifs` variants or trunk, which set an empty `use_oasis`. An OASIS-enabled `libxios.a` carries unresolved `mod_oasis_*` references, so `libfesom.so` links and `bin/fesom.x` then fails, because a standalone FESOM links no OASIS:

```
esm_master install-xios-2.5_r1910_ogcm
esm_master install-fesom-2.8.0-xios
```

This deliberately does not use `requires: [xios-...]`, which is the mechanism nemo uses. That is currently broken for every xios consumer on `release`: requiring a pinned xios version collides with the default `version` in `configs/components/xios/xios.yaml`. `install-nemo-GYRE_XIOS` and `install-nemo-GYRE_PISCES` both fail the same way. Worth fixing separately, and then this can move to `requires`.

## Tested on levante

Both paths, installed with `esm_master` and run for one day on CORE2, 128 ranks:

- Without XIOS. 48 timesteps in 8.6 s, restarts written for 1960-01-02, SSH CG solver 48 solves with 0 non-convergences, 0 breakdowns, 0 non-SPD iterations. No netCDF output, because the default `namelist.io` is monthly and a 1-day run closes no month, same as the 2.7 smoke test.
- With XIOS. 12 variables x 128 ranks written to `outdata/fesom`, `sst` -1.90 to 30.00 degC, `a_ice` within [0, 1], `unod` +/- 0.8 m/s, `MLD3` 7 to 1015 m.

`esm_master install-fesom-2.8.0 -c` and `install-fesom-2.8 -c` resolve to a plain `cmake -DENABLE_OPENMP=ON` with no XIOS anywhere.

## Not covered

No esm_tests entry, as #1561 added none either. Adding one means a bump of the `esm_tests_info` submodule pointer, which is better done on its own.


pgierz (commented):
I'll take care of updating the submodule pointer and add new 2.8 tests.

pgierz:
#bump minor

pgierz:
I will still need to discuss with @mandresm about test baselines, but this can go in.

pgierz (approved):

PR #1592, opened 2026-09-21
esm_catalog tests: loosen ruamel.yaml.clib pin, fix 3.12 CI

pgierz:
## Summary
- `setup.py` pinned `ruamel.yaml.clib==0.2.7` unconditionally; that version has no Python 3.12 wheel and pip fails building it from source. `ruamel.yaml` itself only needs `clib` on `<3.12`. Loosened to `>=0.2.7`.
- With that fixed, `esm-catalog-tests.yml` no longer needs its `--no-deps` + hand-picked dependency list (which had drifted from `setup.py`'s `catalog` extra and was missing `xarray`/`numpy`, breaking test collection: `ModuleNotFoundError: No module named 'xarray'` — `fake_experiment/config_builder.py` writes NetCDF fixtures via `xarray.Dataset.to_netcdf()`). Now installs `-e ".[catalog]"` directly, plus `xarray`/`scipy` for the test fixtures.

Found while triaging PR #1591 (release → prep-release backmerge), where `esm_catalog tests` (3.9, 3.12) was red on both legs.

## Test plan
- [x] Local dry-run: `pip install -e ".[catalog]"` resolves `ruamel.yaml.clib==0.2.15` (has a 3.12 wheel) cleanly
- [x] CI green on this PR (esm_catalog tests, both 3.9 and 3.12)


.github/workflows/esm-catalog-tests.yml
```diff
@@ -30,15 +30,14 @@ jobs:
...
           # setuptools<81 still bundles pkg_resources (removed in 81), which esm_tools imports.
-          pip install -U "pytest>=7.4" "rich-click>=1.7" loguru pystac shapely universal-pathlib jsonschema \
-            "setuptools<81" pyyaml colorama f90nml "pydantic>=2" typing_extensions \
+          # xarray/scipy: fake_experiment/config_builder.py writes NetCDF fixtures via
+          # xarray's to_netcdf(); scipy gives it a NETCDF3 backend with no HDF5 needed.
+          # h5py: the memory:// (fsspec) scan path opens those fixtures through the
+          # h5netcdf engine, which needs h5py -- scipy's engine only handles real paths.
+          # pytest>=7.4: the base pins pytest==7.1.2, which predates 3.12 support.
+          pip install -U "pytest>=7.4" "setuptools<81" numpy xarray scipy h5py \
             "paleodatetime @ git+https://github.com/pgierz/paleodatetime.git"
```
pgierz: I still think this should go into the setup.py properly, as a test requirement if needed, but not here as part of the install job...

pgierz (commented):
One comment for the LLM regarding how to structure the requirements. I would prefer to only have on source of truth for this.

PR #1595, opened 2026-09-21
feat(esm_catalog): 'distributed render-scripts' CLI command

pgierz:
## Summary
- New `esm-catalog distributed render-scripts` command, sitting next to `scan`/`push`/`auth` as its own group (mirrors `auth login`/`logout`).
- Renders the four SLURM scripts (`sched`/`worker`/`driver`/`cleanup`) that bootstrap the Dask scheduler+worker pipeline `esm-catalog scan --distributed --scheduler ...` attaches to -- this pipeline was validated live at 3000 workers / 32,200 files on Albedo earlier this session, but only existed as hand-copied `.sbatch` files with no CLI surface.
- `esm_catalog/distributed/`: four `.j2` templates (bundled as package data via `setup.py`'s `package_data`) + `render_scripts(context, out_dir)`, which reports every missing required variable together, not one at a time.
- CLI accepts an optional positional `VARS_FILE` (YAML) plus one `--flag` per variable; flags override the file, matching the CLI's existing config resolution order (`config.py`: flag > env > file). `log_dir` defaults to `state_dir()/"logs"` (same `XDG_STATE_HOME` convention already used for the auth token cache).
- Every `apptainer exec` in the templates now runs `--cleanenv --env PYTHONNOUSERSITE=1` -- prevents a host conda env or `~/.local` user-site package from shadowing the container's own `esm_catalog` install (the same bug class a stray `~/.local/lib` `.pth` file caused on a dev laptop earlier this session).
- One-line mention added to `esm_catalog_services.rst`'s Components table.

## Test plan
- [x] `pytest tests/test_esm_catalog/test_cli.py` -- 12 new tests (registration, missing-required-vars error, successful render + submit-order output, worker array size, push-after-scan flag toggling the conditional block, CLI flag overriding a vars file, log_dir default) all pass; 1 pre-existing unrelated failure confirmed identical on unmodified `prep-release`
- [x] Manually exercised the command end-to-end (flags-only, vars-file+override, missing-vars error) and inspected the rendered scripts


src/esm_catalog/distributed/templates/driver.sbatch.j2
```diff
@@ -0,0 +1,40 @@
...
+#SBATCH --mem-per-cpu=2G
+#SBATCH --qos={{ qos | default('12h') }}
+#SBATCH --time={{ walltime | default('04:00:00') }}
+#SBATCH --partition={{ partition | default('smp') }}
+#SBATCH -o {{ log_dir }}/driver.%j.log
+set -euo pipefail
+source /etc/profile.d/modules.sh
+module load {{ apptainer_module | default('apptainer') }}
+export APPTAINER_CACHEDIR={{ scratch_dir }}/apptainer-cache
+IMG=oras://ghcr.io/esm-tools/esm_tools:sif-{{ image_tag }}
```
pgierz: this needs to default to singularity for the time being, and later on, needs to be figured out from the computer dynamically via the config yaml.

